PR #91, opened 2026-02-19
fix: handle read-only .git objects on Windows in shutil.rmtree calls

### LINKED ISSUE #90: [BUG]  apm install fails with [WinError 5] Access is denied on Windows when processing .git objects
Description:

  After a package passes validation, the install step fails on Windows with an access denied error when APM tries to
  manipulate files inside the cloned .git/objects directory.

  Steps to reproduce:
  1. Successfully validate and install an APM package on Windows
  2. Installation fails with:
  ❌ Failed to install org/project/repo: [WinError 5] Access is denied:
  'C:\...\apm_modules\org\project\repo\.git\objects\02\...'

  Expected behavior:
  APM should handle read-only .git object files on Windows (git marks these as read-only by design).

  Environment:
  - APM CLI version: 0.7.3 (pip installation)
  - OS: Windows 11

ryanfk:
## Summary

- On Windows, Git marks pack files and loose objects as **read-only**. When APM tries to remove a previously-cloned `apm_modules` directory or strip the `.git` folder after cloning, `shutil.rmtree` raises `[WinError 5] Access is denied`.
- Adds a `_remove_readonly` error handler that clears the read-only flag before retrying the delete.
- Wires the handler into all five `shutil.rmtree` calls in `github_downloader.py`, replacing both bare calls and `ignore_errors=True` (which silently left files behind on Windows).

## Reproduction

```
apm install dev.azure.com/myorg/myproject/myrepo
# First install fails for any reason (e.g. wrong branch)
# Second install attempt hits:
# ❌ Failed to install myorg/myproject/myrepo: [WinError 5] Access is denied:
#    'C:\...\apm_modules\myorg\myproject\myrepo\.git\objects\02\...'
```

## Fix

```python
def _remove_readonly(func, path, _excinfo):
    """Clear read-only flag on Windows before retrying delete."""
    os.chmod(path, stat.S_IWRITE)
    func(path)

# Before:
shutil.rmtree(target_path)
# After:
shutil.rmtree(target_path, onerror=_remove_readonly)
```

## Test plan

- [x] Verified on Windows 11 — `apm install` now succeeds after a previously failed install leaves read-only `.git` objects in `apm_modules/`
- [x] No-op on Linux/macOS — the `onerror` handler only triggers when the default delete fails, which doesn't happen on systems without read-only git objects

Fixes #90


JustinGOSSES:
@microsoft-github-policy-service rerun

PR #119, opened 2026-02-27
chore: recompile agentic workflow lock files with gh-aw v0.50.6

danielmeppiel:
## Summary

Recompiles all agentic workflow lock files using `gh aw` v0.50.6 (previously compiled with v0.42.2).

## Problem

The lock files generated by v0.42.2 had a runtime-import path resolution bug:
- **Compiler emitted**: `{{#runtime-import workflows/daily-test-improver.md}}`
- **Runtime resolver** saw a path without `.github/` prefix, so it prepended `workflows/` again
- **Result**: looked for `.github/workflows/workflows/daily-test-improver.md` → file not found

Failing run: https://github.com/microsoft/apm/actions/runs/22483217187

## Fix

v0.50.6 compiler now emits: `{{#runtime-import .github/workflows/daily-test-improver.md}}`

The runtime correctly strips the `.github/` prefix and resolves to `.github/workflows/daily-test-improver.md`.

## Changes

- Recompiled 3 workflow lock files (`daily-test-improver`, `daily-doc-updater`, `cli-consistency-checker`)
- Added `.github/aw/actions-lock.json` (new action pin lock file introduced in newer gh-aw versions)


SebastienDegodez:
Is it impossible to inherit a workflow from a source ?

https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow

danielmeppiel:
> Is it impossible to inherit a workflow from a source ?
> 
> https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow

Yes but with [GitHub Agentic Workflows](https://github.github.com/gh-aw/) we rather edit the `.md` file and then the workflow yaml gets generated (`.lock.yml`). So the reuse has to be defined in the markdown file. Do you have a reuse scenario in mind?

PR #178, opened 2026-03-06
fix: handle commit SHA refs in subdirectory package clones

### LINKED ISSUE #177: `apm install` fails on second run when lockfile contains commit SHAs for subdirectory packages
## Bug

Running `apm install` twice in succession fails on all subdirectory (monorepo) packages during the second run.

### Steps to reproduce

1. Start with a clean project (no `apm_modules/`, no `apm.lock`)
2. Run `apm install --trust-transitive-mcp` — succeeds, generates `apm.lock`
3. Run `apm install --trust-transitive-mcp` again — **all subdirectory packages fail** with:
   ```
   fatal: Remote branch <commit-sha> not found in upstream origin
   ```

### Root cause

When the lockfile exists, the install logic appends the locked commit SHA as a git reference (`#<sha>`) to the download ref string. For subdirectory packages, `download_subdirectory_package` passes this SHA as `branch=<sha>` with `depth=1` to `git clone`. Git shallow clones (`--depth=1 --branch=X`) only accept branch/tag names, not arbitrary commit SHAs — so every package fails.

### Expected behavior

Repeated `apm install` should be idempotent — the second run should produce the same result as the first.

### Fix

Detect when the ref is a commit SHA in `download_subdirectory_package` and use a full clone + checkout instead of `--depth=1 --branch=<sha>`.

sergio-sisternes-epam:
Fixes #177

## Problem

`download_subdirectory_package` passed lockfile commit SHAs as `--branch` to shallow clones, which git does not support. This caused every subdirectory package to fail on the second `apm install` when a lockfile was present.

### Reproduction

1. Clean install (`rm -rf apm_modules apm.lock`) → succeeds, writes `apm.lock` with `resolved_commit: <sha>`
2. Run `apm install` again → all subdirectory packages fail with `fatal: Remote branch <sha> not found in upstream origin`

## Solution

- Detect when the ref is a commit SHA (7-40 hex chars) in `download_subdirectory_package`
- For commit SHAs: do a full clone, then `git checkout <sha>`
- For branches/tags: keep existing shallow clone behavior (`--depth=1 --branch=<ref>`)

## Testing

- All 100 existing tests pass (33 skipped — environment-gated)
- Manual verification: `apm install` is now fully idempotent across repeated runs

### DIFF
--- a/scripts/build-binary.sh
+++ b/scripts/build-binary.sh
@@ -62,6 +62,16 @@ else
     echo -e "${YELLOW}UPX not found - binary will not be compressed (install with: brew install upx)${NC}"
 fi
 
+# Inject build SHA into version.py
+VERSION_FILE="src/apm_cli/version.py"
+BUILD_SHA=$(git rev-parse --short HEAD 2>/dev/null || echo "")
+if [ -n "$BUILD_SHA" ]; then
+    echo -e "${YELLOW}Injecting build SHA: $BUILD_SHA${NC}"
+    sed -i.bak "s/^__BUILD_SHA__ = None$/__BUILD_SHA__ = \"$BUILD_SHA\"/" "$VERSION_FILE"
+    # Guarantee restore on any exit (success, failure, or signal)
+    trap 'if [ -f "${VERSION_FILE}.bak" ]; then mv "${VERSION_FILE}.bak" "$VERSION_FILE"; fi' EXIT
+fi
+
 # Build binary
 echo -e "${YELLOW}Building binary with PyInstaller...${NC}"
 uv run pyinstaller build/apm.spec
--- a/src/apm_cli/cli.py
+++ b/src/apm_cli/cli.py
@@ -36,7 +36,7 @@ from apm_cli.utils.console import (
 from apm_cli.utils.github_host import is_valid_fqdn, default_host
 
 # APM imports - use absolute imports everywhere for consistency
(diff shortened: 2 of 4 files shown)


danielmeppiel (approved):
LGTM — correct fix for a real bug (lockfile commit SHAs breaking `--branch`).

**What I verified:**
- Root cause accurately identified: `git clone --branch <sha>` is unsupported for arbitrary commit SHAs ✓
- SHA detection regex `r'^[a-f0-9]{7,40}$'` is the standard pattern (7-char min avoids false positives on short branch names) ✓
- `no_checkout=True` → GitPython translates to `--no-checkout` for `Repo.clone_from` — correct ✓
- `_try_sparse_checkout` also fails for commit SHAs (`git fetch origin <sha> --depth=1`), but gracefully returns `False` and falls through to the now-fixed full clone path — flow is correct ✓
- Exception chaining (`from e`) added — nice improvement ✓
- 4 tests cover SHA, branch, no-ref, and checkout failure cases ✓

**Non-blocking note for future:** Full clone for SHA refs could be slow on large repos. A more optimal approach would be `git init` + `git fetch origin <sha>` + `git checkout FETCH_HEAD`, but this is an acceptable v1. Could be optimized later if users report perf issues.

**Merge order:** PR 2 of 3. Rebase onto main after #176 merges.

PR #287, opened 2026-03-13
fix(windows): resolve Windows CI integration test failures (5/5 validation)

sergio-sisternes-epam:
## Description

Fix Windows CI integration test failures — Codex runtime download 404, Unicode encoding errors on cp1252 consoles, and Hero 2 validation failures (token discovery, subprocess resolution, git config).

After these fixes, the Windows release validation passes **5/5** (previously 2/5).

### Validation Results

| # | Test | Status | Problem | Fix |
|---|------|--------|---------|-----|
| 1 | **Prerequisites** (GITHUB_TOKEN) | ✅ Pass | Token helpers failed when `GITHUB_TOKEN` env var not set, even with `gh` CLI authenticated | Added `gh auth token` fallback in both `github-token-helper.ps1` and `.sh` |
| 2 | **Basic commands** | ✅ Pass | — (already passing) | — |
| 3 | **Runtime setup** (Codex) | ✅ Pass | Codex download returned 404 — Windows assets are named `codex-{platform}.exe.tar.gz` but script used `.tar.gz` | Fixed asset URL in `setup-codex.ps1` (3 occurrences) |
| 4 | **Hero 1** (30s zero-config) | ✅ Pass | Unicode `charmap` encoding error on cp1252 consoles when outputting `✓`, `└─`, `⚠`, etc. | Force UTF-8 console via `SetConsoleOutputCP(65001)` + stream reconfigure in `cli.py` |
| 5 | **Hero 2** (2min guardrailing) | ✅ Pass | (a) `WinError 2` — `subprocess.run(["copilot", ...])` can't find `.cmd` wrappers; (b) `fatal: unable to access 'NUL'` — git config on some Windows git versions | (a) `shutil.which()` resolution in `script_runner.py`; (b) Empty temp file instead of `NUL` for `GIT_CONFIG_GLOBAL` |

### Commits

- `0688a7a` — fix: correct Codex Windows download URL (.exe.tar.gz asset naming)
- `6e39deb` — fix: configure UTF-8 encoding on Windows to prevent UnicodeEncodeError
- `d9034eb` — fix(windows): resolve Hero 2 validation failures

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass (2009 passed, 0 failures)
- [x] Added tests for new functionality (7 encoding tests, 3 subprocess resolution tests, updated 4 existing tests)

### DIFF
--- a/src/apm_cli/deps/github_downloader.py
+++ b/src/apm_cli/deps/github_downloader.py
@@ -200,12 +200,12 @@ class GitHubPackageDownloader:
         env['GIT_ASKPASS'] = 'echo'  # Prevent interactive credential prompts
         env['GIT_CONFIG_NOSYSTEM'] = '1'
         if sys.platform == 'win32':
-            # 'NUL' fails on some Windows git versions; use an empty temp file
+            # 'NUL' fails on some Windows git versions; use an empty temp file.
+            # Always truncate to guarantee emptiness (防 tampering in shared tmp).
             import tempfile
             empty_cfg = os.path.join(tempfile.gettempdir(), '.apm_empty_gitconfig')
-            if not os.path.exists(empty_cfg):
-                with open(empty_cfg, 'w') as f:
-                    pass
+            with open(empty_cfg, 'w') as f:
+                pass
             env['GIT_CONFIG_GLOBAL'] = empty_cfg
         else:
             env['GIT_CONFIG_GLOBAL'] = '/dev/null'
--- a/tests/unit/test_cli_encoding.py
+++ b/tests/unit/test_cli_encoding.py
@@ -1,7 +1,6 @@
 """Tests for CLI entry point encoding configuration."""
 
-import sys
-from unittest.mock import patch, MagicMock, call
+from unittest.mock import patch, MagicMock
 
 from apm_cli.cli import _configure_encoding
 


tests/unit/test_cli_encoding.py
```diff
@@ -0,0 +1,124 @@
+"""Tests for CLI entry point encoding configuration."""
+
+import sys
+from unittest.mock import patch, MagicMock, call
```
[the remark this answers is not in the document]
sergio-sisternes-epam: Fixed in 04e2f2a — removed both unused imports (`sys` and `call`).

src/apm_cli/deps/github_downloader.py
```diff
@@ -199,7 +199,16 @@ def _setup_git_environment(self) -> Dict[str, Any]:
...
+        if sys.platform == 'win32':
+            # 'NUL' fails on some Windows git versions; use an empty temp file
+            import tempfile
+            empty_cfg = os.path.join(tempfile.gettempdir(), '.apm_empty_gitconfig')
+            if not os.path.exists(empty_cfg):
+                with open(empty_cfg, 'w') as f:
+                    pass
+            env['GIT_CONFIG_GLOBAL'] = empty_cfg
+        else:
+            env['GIT_CONFIG_GLOBAL'] = '/dev/null'
```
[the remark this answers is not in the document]
sergio-sisternes-epam: Good catch. Fixed in 04e2f2a — changed to always truncate with `open(empty_cfg, 'w')` instead of the `if not os.path.exists()` guard, so the file is guaranteed empty on every invocation regardless of prior state.

scripts/github-token-helper.sh
```diff
@@ -33,7 +33,18 @@ setup_github_tokens() {
             echo -e "${GREEN}✓ Preserving existing GITHUB_TOKEN for Models access (${#GITHUB_TOKEN} chars)${NC}"
         fi
     else
-        echo -e "${YELLOW}⚠️  No GITHUB_TOKEN found initially${NC}"
+        # Try to obtain a token from the gh CLI if authenticated
```
danielmeppiel, line 36: let's not add this complexity - the github-token-helper.sh is well tested and is not failing on windows since it's not used on windows. Remove changes made to this file.

scripts/windows/github-token-helper.ps1
```diff
@@ -27,7 +27,20 @@ function Initialize-GitHubToken {
             Write-Host "$([char]0x2713) Preserving existing GITHUB_TOKEN for Models access ($($env:GITHUB_TOKEN.Length) chars)" -ForegroundColor Green
         }
     } else {
-        Write-Host "Warning: No GITHUB_TOKEN found initially" -ForegroundColor Yellow
+        # Try to obtain a token from the gh CLI if authenticated
```
danielmeppiel, line 30: remove this block - the script should work if the script user does authenticate with gh cli beforehand - but I do not want to add this block here and rely on gh cli. In CI/CD, the GITHUB_TOKEN is correctly passed in anyways, so this is not the source of error

tests/unit/test_compile_rich_output.py
```diff
@@ -11,7 +11,7 @@
 
 def test_rich_output_contains_table_and_status():
     with temp_project_with_constitution(constitution_text=DEFAULT_CONSTITUTION) as proj:
-        proc = subprocess.run(CLI, cwd=str(proj), capture_output=True, text=True)
+        proc = subprocess.run(CLI, cwd=str(proj), capture_output=True, text=True, encoding="utf-8")
```
danielmeppiel, line 14: why do we need to fix the encoding on this one here? the encoding is configured in cli.py anyways. This will force utf-8 encoding on non-Windows platforms

danielmeppiel (changes requested):

PR #449, opened 2026-03-25
fix: warn when instruction is missing required fields during `apm compile`

alopezsanchez:
## Description

Instructions without `applyTo` were silently dropped during `apm compile` in distributed and claude modes because `validate_primitives()` was only called in the single-file path. This adds validation to all compilation modes and surfaces warnings through the CLI.

- Call `validate_primitives()` in `_compile_distributed()` and `_compile_claude_md()`
- Merge `self.warnings` into dry-run and normal compilation results
- Remove distributed-mode warning suppression in CLI
- Add unit and CLI integration tests for the warning

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [ ] Tested locally
- [x] All existing tests pass

### DIFF
--- a/src/apm_cli/commands/compile/cli.py
+++ b/src/apm_cli/commands/compile/cli.py
@@ -154,8 +154,8 @@ def _get_validation_suggestion(error_msg):
     """Get actionable suggestions for validation errors."""
     if "Missing 'description'" in error_msg:
         return "Add 'description: Your description here' to frontmatter"
-    elif "Missing 'applyTo'" in error_msg:
-        return "Add 'applyTo: \"**/*.py\"' to frontmatter"
+    elif "applyTo" in error_msg and "globally" in error_msg:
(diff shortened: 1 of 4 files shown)


src/apm_cli/commands/compile/cli.py
```diff
@@ -529,16 +529,13 @@ def compile(
                     else:
                         _display_next_steps(output)
 
-        # Common error handling for both compilation modes
-        # Note: Warnings are handled by professional formatters for distributed mode
-        if config.strategy != "distributed" or single_agents:
```
alopezsanchez, line 534: I'm not sure about this line, but it took me some time to realize that I was missing the `applyTo` field. The `apm compile` wasn't adding the instruction to the AGENTS/CLAUDE.md, and I didn't know why.

PR #540, opened 2026-04-02
feat: add Nix support

srid:
## Description

Package apm with uv2nix so it can be run via `nix run github:microsoft/apm` ([what's Nix?](https://nixos.asia/en/nix-first)) on Linux and macOS (and Windows, via WSL). 

To test, run:

```
nix run github:srid/apm/add-nix-flake
```

image

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [ ] All existing tests pass

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,18 @@
+name: Nix
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - run: nix build
+      - run: nix run . -- --version
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ build/tmp/
 scout-pipeline-result.png
 .copilot/
 .playwright-mcp/
+
+# Nix
+result
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ Native release binaries are published for macOS, Linux, and Windows x86_64. `apm
 ```bash
 # Homebrew
 brew install microsoft/apm/apm
+# Nix
+nix run github:microsoft/apm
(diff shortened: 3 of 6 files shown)


danielmeppiel:
Hi @srid -- circling back on this with more concrete context now that #675 has landed.

#675 added a build-time `update_policy` module that lets package-manager distributions disable `apm update` and surface their own update guidance instead. The conda-forge package ([apm-cli-feedstock](https://github.com/conda-forge/apm-cli-feedstock), maintained by @melund) ships using that pattern: feedstock lives downstream, only the upstream enabler lives in this repo.

For Nix, I'd love to support users via the same model rather than carrying `flake.nix` / `flake.lock` in-tree. Concretely: APM packaged in [`nixpkgs`](https://github.com/NixOS/nixpkgs) (or a Nix overlay you maintain in a separate repo we link from our docs).

Reasons:

1. I don't have Nix expertise to maintain `flake.nix` / refresh `flake.lock` reliably, debug `uv2nix` regressions, or audit transitive Nix-side dependencies. An in-tree flake that breaks (advertised as `nix run github:microsoft/apm` from the README) is a worse first-run experience than no flake at all.
2. Membership in `nixpkgs` is the actual signal of "real Nix support" -- it reaches users via `nix-shell -p apm` / `nix profile install nixpkgs#apm` and shows up in the standard Nix discovery surfaces. An in-repo flake from one upstream contributor doesn't carry that signal.
3. It keeps audit responsibility with the Nix community, who are qualified to do it.

If you'd be up for shepherding APM into `nixpkgs` (or maintaining a separate flake repo), I'd happily coordinate any small upstream changes you need -- including using `update_policy` from #675 to disable self-update in the Nix build, the same way conda-forge does. Happy to link the resulting package prominently from the README and docs once it's available.

Leaving #540 closed for now, but very much open to revisiting via this route. Thanks again for the contribution and for caring enough to push on this.

srid:
@danielmeppiel Makes full sense to me.

Somebody will eventually package apm in `nixpkgs`. Third-party packaging also exists, such as https://github.com/numtide/llm-agents.nix/pull/3666. 

All of those tend to be behind, so I've resolved to using `nix shell nixpkgs#uv -c uvx git+...` (uv works pretty well).

Issue #539, opened 2026-04-02
[BUG] sync_integration() hard-codes skill prefixes -- misses user-scope paths during uninstall


sergio-sisternes-epam:
**Describe the bug**
`sync_integration()` in `skill_integrator.py` hard-codes skill directory
prefixes (e.g. `".opencode/skills/"`) when scanning managed files for
removal. User-scope deployments can produce roots with nested path separators
(e.g. `".config/opencode/skills/"`), which do not match the hard-coded
prefixes. This means user-scope skill directories are not cleaned up during
uninstall/sync.

**To Reproduce**
1. `apm install -g --target opencode` (deploys to `~/.config/opencode/skills/`)
2. `apm uninstall -g <package>` (sync_integration tries to remove orphaned skills)
3. Hard-coded prefix `".opencode/skills/"` does not match `".config/opencode/skills/"`
4. Orphaned skill directories remain under `~/.config/opencode/skills/`

**Expected behavior**
`sync_integration()` should derive allowed skill prefixes from `KNOWN_TARGETS`
using `target.effective_root(user_scope=...)` (and `deploy_root` overrides)
instead of hard-coded strings.

**Environment (please complete the following information):**
- OS: All (code-level bug, not platform-specific)
- Python Version: 3.12
- APM Version: 0.8.9
- Affected file: `src/apm_cli/integration/skill_integrator.py`
  (`sync_integration()` method)

**Logs**
N/A -- no runtime error; orphaned directories silently remain.

**Additional context**
Pre-existing issue not introduced by #537, but surfaced during Copilot code
review. Affects any target where `user_root_dir != root_dir` (currently
Copilot and OpenCode).

Suggested approach: build the prefix set dynamically from
`KNOWN_TARGETS.values()` iterating over each target's `effective_root()`
for both scopes, similar to how `get_integration_prefixes()` works in
`base_integrator.py`.

Related: #537, #530

Issue #538, opened 2026-04-02
[BUG] copy_skill_to_target falls back to project-scope active_targets when targets=None at user scope


sergio-sisternes-epam:
**Describe the bug**
`copy_skill_to_target()` accepts `user_scope` but when `targets` is `None`,
it falls back to `active_targets(target_base)` which performs project-scope
auto-detection. A caller passing `user_scope=True` without explicit `targets`
would select project-scope roots instead of user-scope roots.

**To Reproduce**
1. Call `skill_integrator.copy_skill_to_target(...)` with `user_scope=True`
   and `targets=None`
2. The method calls `active_targets()` which scans project-scope directories
3. Deployment uses `root_dir` instead of `user_root_dir`

**Expected behavior**
When `user_scope=True` and `targets=None`, the method should resolve
user-scope targets (e.g. via a `active_targets_user_scope()` helper or by
requiring explicit targets when `user_scope=True`).

**Environment (please complete the following information):**
- OS: All (code-level bug, not platform-specific)
- Python Version: 3.12
- APM Version: 0.8.9
- Affected file: `src/apm_cli/integration/skill_integrator.py` (line ~253)

**Logs**
N/A -- code-level logic issue, no runtime error logs.

**Additional context**
Currently low risk because the install path in `install.py` always passes
explicit `targets`. Flagged during Copilot review of #537. Options:
1. Add an `active_targets_user_scope()` fallback
2. Raise `ValueError` when `user_scope=True` and `targets=None`
3. Document that callers must pass explicit targets at user scope

Related: #537, #530

PR #542, opened 2026-04-02
fix: scope-resolved target profiles + Claude Code instructions

### LINKED ISSUE #530: [BUG] `apm install -g --target opencode` installs skills to `~/.opencode/` instead of `~/.config/opencode/`
**Describe the bug**

When running `apm install -g --target opencode`, skills are silently deployed to `~/.opencode/skills/<skill>/` instead of the correct user-scope path `~/.config/opencode/skills/<skill>/`. No error or warning is raised.

This is caused by two related issues in `skill_integrator.py`:
1. **Wrong root dir used** — `_integrate_native_skill` (line 717) uses `target.root_dir` (`.opencode`) directly instead of calling `target.effective_root(user_scope=True)` (`.config/opencode`). The `user_scope` flag is never propagated into the skill integrator.
2. **`auto_create` guard missing** — All other integrators (agent, command, instruction, prompt) skip deployment when `target.auto_create=False` and the target dir doesn't already exist. `skill_integrator.py` has no such guard, so it unconditionally creates `~/.opencode/skills/` from scratch via `mkdir(parents=True)`.
 
**To Reproduce** 
1. `apm install -g --target opencode <some-skill-package>`
2. Observe `~/.opencode/skills/<skill>/` is created
3. `~/.config/opencode/skills/` is never touched
 
**Expected behavior**

Skill installed to `~/.config/opencode/skills/<skill>/`. If `~/.config/opencode/` doesn't exist and `auto_create=False`, skip with a warning (consistent with other integrators).

**Relevant code**

- `src/apm_cli/integration/skill_integrator.py:717` — `target.root_dir` should be `target.effective_root(user_scope=<flag>)`
- `src/apm_cli/integration/skill_integrator.py:728` — `mkdir` called unconditionally, no `auto_create` guard
- `src/apm_cli/integration/targets.py:99-107` — `effective_root()` exists and handles this correctly, just not called
- Compare with `agent_integrator.py:115`, `command_integrator.py:135` for the correct `auto_create` pattern

**Environment**

- OS: WSL (Linux)
- Python Version: 3.12.3
- APM Version: 0.8.9

**Logs**

No error output — failure is silent.

danielmeppiel:
## Description

Introduces `TargetProfile.for_scope()` and `resolve_targets()` so scope resolution happens **exactly once** -- no parameter threading, no caller responsibility. Supersedes #536 and #537. Builds on #527 (already merged).

### Architecture: scope-resolved targets

The root cause of user-scope bugs: every code path that touches `target.root_dir` must choose the right value for the scope. Both #536 (dataclasses.replace at dispatch) and #537 (thread user_scope through every method) fix the install path but both **miss uninstall Phase 2 re-integration**, and neither handles multi-level `user_root_dir` values like `.config/opencode` in partition routing.

**Solution**: resolve scope on the data object itself, once, at the entry point.

- `for_scope(user_scope)` returns a frozen copy with `root_dir` resolved and unsupported primitives filtered out
- `resolve_targets()` is the single entry point combining detection + scope resolution + filtering
- Integrators read `target.root_dir` directly -- always correct for the current scope
- `partition_managed_files(targets=)` uses trie-based O(depth) routing, fixing `.config/opencode/` multi-level root paths
- `validate_deploy_path(targets=)` derives prefixes from resolved targets
- Uninstall Phase 2 re-integration uses the same resolved targets -- **eliminates wrong-path deployment bug**

### Skill integrator scope fixes (#538, #539)

- `sync_integration()` now derives skill prefixes dynamically from targets instead of hardcoded strings -- handles `.copilot/skills/`, `.config/opencode/skills/`, etc.
- Uninstall Phase 2 passes resolved targets to `integrate_package_skill()`
- `copy_skill_to_target()` adds auto_create guard for targets without existing directories

### Copilot CLI user-scope fix (#536)

- Add `instructions` to `unsupported_user_primitives` (per [official docs](https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-config-dir-reference))
- Fix `resolve_dependencies` to use `apm_dir` at user scope

Fixes #530, #538, #539
Supersedes #536, #537

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Maintenance / refactor

## Testing

- [x] All existing tests pass (3,506 unit tests)
- [x] 40+ new tests covering scope resolution, partition routing, skill sync, and backward compatibility

### DIFF
--- a/src/apm_cli/commands/uninstall/engine.py
+++ b/src/apm_cli/commands/uninstall/engine.py
@@ -308,7 +308,8 @@ def _sync_integrations_after_uninstall(apm_package, project_root, all_deployed_f
     if _skill_dirs_exist:
         integrator = SkillIntegrator()
         result = integrator.sync_integration(apm_package, project_root,
-                                             managed_files=_buckets["skills"] if _buckets else None)
+                                             managed_files=_buckets["skills"] if _buckets else None,
+                                             targets=_resolved_targets)
         counts["skills"] = result.get("files_removed", 0)
 
     # Hooks (multi-target, sync_integration handles all targets)
@@ -369,7 +370,7 @@ def _sync_integrations_after_uninstall(apm_package, project_root, all_deployed_f
                         getattr(_integrator, _method)(
                             _target, pkg_info, project_root,
                         )
-            skill_integrator.integrate_package_skill(pkg_info, project_root)
+            skill_integrator.integrate_package_skill(pkg_info, project_root, targets=_targets)
         except Exception:
             pkg_id = dep_ref.get_identity() if hasattr(dep_ref, "get_identity") else str(dep_ref)
             logger.warning(f"Best-effort re-integration skipped for {pkg_id}")
--- a/src/apm_cli/integration/skill_integrator.py
+++ b/src/apm_cli/integration/skill_integrator.py
@@ -301,6 +301,9 @@ def copy_skill_to_target(
     deployed: list[Path] = []
 
     # Deploy to all active targets that support skills.
+    # When no targets are provided, fall back to project-scope detection.
+    # Callers responsible for user-scope should pass resolved targets
+    # from resolve_targets().
     if targets is None:
         from apm_cli.integration.targets import active_targets
         targets = active_targets(target_base)
@@ -309,6 +312,12 @@ def copy_skill_to_target(
             continue
         skills_mapping = target.primitives["skills"]
         effective_root = skills_mapping.deploy_root or target.root_dir
+
+        # Skip if target dir does not exist and auto_create is disabled
+        target_root_dir = target_base / target.root_dir
+        if not target.auto_create and not target_root_dir.is_dir():
+            continue
+
         skill_dir = target_base / effective_root / "skills" / skill_name
         skill_dir.parent.mkdir(parents=True, exist_ok=True)
         if skill_dir.exists():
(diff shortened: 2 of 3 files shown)


danielmeppiel:
## What this PR does

User-scope deployment (`apm install -g`) was broken across multiple code paths because every caller had to remember to use `target.effective_root(user_scope=True)` instead of `target.root_dir`. Many callers forgot -- causing skills to deploy to `~/.github/` instead of `~/.copilot/`, uninstall to miss cleanup, and partition routing to fail for multi-level paths like `~/.config/opencode/`.

This PR eliminates the problem at its root: **scope resolution happens once, on the data object itself, before any code sees it.**

- `TargetProfile.for_scope(user_scope)` returns a frozen copy with `root_dir` already set to the correct value and unsupported primitives filtered out
- `resolve_targets()` is the single entry point -- callers get profiles where `root_dir` is always correct
- Every downstream consumer (integrators, partition routing, validation, install, uninstall) just reads `target.root_dir` -- no scope awareness needed

### How it solves each linked issue

**#530 -- skills deploy to wrong user-scope path**: The core bug. `resolve_targets(project_root, user_scope=True)` now returns Copilot with `root_dir=".copilot"` and OpenCode with `root_dir=".config/opencode"`. All integrators read `root_dir` directly, so skills (and agents, commands, instructions) land in the right place. Install and uninstall both use the same resolved targets, including Phase 2 re-integration which was previously broken.

**#538 -- `copy_skill_to_target()` ignores user_scope**: Fixed by passing resolved targets from the uninstall engine to `integrate_package_skill()`. Added `auto_create` guard to `copy_skill_to_target()` so targets without existing directories are skipped instead of silently created.

**#539 -- `sync_integration()` uses hardcoded skill prefixes**: Refactored to derive prefixes dynamically from the targets list. The old code had five hardcoded blocks (`.github/skills/`, `.claude/skills/`, `.cursor/skills/`, `.opencode/skills/`, `.agents/skills/`) that couldn't match user-scope paths like `.copilot/skills/`. Now it loops over whatever targets are passed, building prefixes from `deploy_root or target.root_dir`.

Also cherry-picks from superseded PRs: Copilot `instructions` added to `unsupported_user_primitives` (#536), `resolve_dependencies(apm_dir)` fix (#536).

PR #546, opened 2026-04-02
fix: compare content hashes during --update when SHA matches

### LINKED ISSUE #515: [FEATURE] --update should re-resolve transitive dependency tree when content changes at the same commit

When a package author updates the dependency tree (e.g. removes a transitive dependency) without creating a new commit on the default branch, `apm install --update` does not detect the change. It compares the `resolved_commit` in `apm.lock.yaml` against the current upstream ref — and since the commit SHA is identical, it skips re-resolution entirely. The only workaround is to delete `apm.lock.yaml` and run a fresh `apm install`.

`--update` should also compare the **content hashes** of resolved packages against the lock file, not just the commit SHA. If the content hash differs at the same ref, the transitive dependency tree should be re-resolved.

Proposed behaviour:
1. Fetch the upstream ref (as today)
2. If the commit SHA matches the lock file, additionally compare content hashes
3. If any content hash has changed, invalidate that subtree and re-resolve its transitive dependencies
4. Update `apm.lock.yaml` with the new tree

- **Manual lock file deletion** — works but discards all pinned resolutions, not just the affected subtree.
- **`--force` flag** — could force full re-resolution, but doesn't address the root cause.
- **Upstream discipline** — requiring package authors to always create new commits for dependency tree changes. This is fragile and hard to enforce.

**Additional context**

- The lock file already stores `content_hash` per dependency, so the data needed for this check is already available.

sergio-sisternes-epam:
## Summary

Fixes #515

When `apm install --update` / `apm deps update` resolves the same commit SHA as the lockfile, the package is skipped (no re-download). This misses content changes invisible to SHA comparison -- notably when registry proxies re-publish different content at the same version identifier.

## Changes

After the SHA match sets `lockfile_match = True`, we now additionally compute the local content hash and compare with the lockfile's stored `content_hash`. If they differ, `lockfile_match` is reset to `False`, forcing re-download and re-resolution of the transitive dependency tree.

### Behavior matrix

| SHA match? | Hash in lockfile? | Hash match? | Result |
|---|---|---|---|
| No | -- | -- | Re-download (existing) |
| Yes | No | -- | Skip (backward compat) |
| Yes | Yes | Yes | Skip |
| Yes | Yes | No | **Re-download** (new) |

### Changed files

| File | Change |
|---|---|
| `src/apm_cli/commands/install.py` | Content hash comparison after SHA match in `update_refs` block (+14 lines); fix pre-existing non-ASCII em dash |
| `tests/unit/test_install_update.py` | 10 new test cases: hash match/mismatch, no-hash backward compat, SHA-differs bypass, corrupted checkout |
| `CHANGELOG.md` | Entry under `[Unreleased] > Fixed` |

### Notes

- Existing integrity verification (lines 1937-1949, corruption detection on normal installs) is unchanged
- Existing tamper detection (lines 2225-2245, supply-chain protection) is unchanged
- `compute_package_hash()` only runs when SHA matches AND `content_hash` exists in lockfile -- no performance impact on the common path
- Backward compatible: packages without `content_hash` in lockfile behave exactly as before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `apm install --update` now compares content hashes when the remote SHA matches the lockfile, detecting content changes invisible to SHA comparison (#515)
 - Hook integrator now processes the `windows` property in hook JSON files, copying referenced scripts and rewriting paths during install/compile (#311)
 - `apm install -g` now correctly deploys to user-scope directories (e.g., `~/.copilot/` instead of `~/.github/`) across all integrators and uninstall paths (#542)
 - `apm install -g` no longer deploys instructions to `~/.copilot/instructions/` (unsupported by Copilot CLI) (#542)
--- a/src/apm_cli/commands/install.py
+++ b/src/apm_cli/commands/install.py
@@ -1906,6 +1906,20 @@ def _install_apm_dependencies(
                                     local_repo = GitRepo(install_path)
                                     if local_repo.head.commit.hexsha == locked_dep.resolved_commit:
                                         lockfile_match = True
+                                        # Content hash comparison: detect changes
+                                        # invisible to SHA comparison (e.g. registry
+                                        # proxies re-publishing at the same version).
+                                        if lockfile_match and locked_dep.content_hash:
+                                            from ..utils.content_hash import compute_package_hash
+                                            local_hash = compute_package_hash(install_path)
+                                            if local_hash != locked_dep.content_hash:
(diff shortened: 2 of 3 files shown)


sergio-sisternes-epam:
Closing after further analysis. The content hash comparison adds complexity for a near-zero real-world scenario (same SHA, different content only applies to registry proxies). For pure Git repos, SHA changes imply content changes. Additionally, silently re-downloading when local files differ conflicts with the `--force` contract.

The real user-reported problem (new SHA pushed but not downloaded) is tracked in #548 -- caused by `download_callback()` using locked SHAs during `--update`, bypassing re-resolution entirely.

PR #571, opened 2026-04-04
[fix] suppress internal config keys from apm config get output (#564)

edenfunf:
## What

`apm config get` (no argument) printed every key in `~/.apm/config.json`, including the internal `default_client` key. Users saw `default_client: vscode` in the output but got an error if they tried `apm config set default_client vscode`:

```
[x] Unknown configuration key: 'default_client'. Valid keys: auto-integrate
```

This get/set asymmetry is confusing and was flagged as a Medium severity finding in #564.

## Why

The display loop in `commands/config.py` mapped `auto_integrate` → `auto-integrate` correctly, but had an `else` fallthrough that printed any other key verbatim. `default_client` is an internal key used by the package manager adapter; it is not user-settable and should not appear in user-facing output.

## How

Remove the `else` fallthrough so only explicitly mapped, user-settable keys are printed. Internal keys are silently skipped. This is the minimal change — no new abstraction, no changes to the underlying config storage or `default_client` itself (still used internally by `default_manager.py`).

## Test

- Updated `test_get_all_config` to assert `default_client` is **not** present in `apm config get` output when the config contains it.
- Replaced `test_get_all_config_unknown_key_passthrough` (which validated the buggy behaviour) with `test_get_all_config_internal_keys_suppressed`, which asserts internal keys are silently hidden.

All 23 existing config command tests pass.


danielmeppiel (approved):

sergio-sisternes-epam (approved):
**Looks good, @edenfunf!** Clean fix — calling `get_auto_integrate()` directly is the right approach, avoids the leaky fallthrough and handles the fresh-install case correctly.

One thing before we can merge: the branch is behind `main`. Could you update it by merging main into your branch? You can do this via the "Update branch" button on this PR page. That will trigger CI and unblock the merge.

PR #606, opened 2026-04-07
fix: improve apm init onboarding — fix repeated confirm, correct next steps, create start.prompt.md

edenfunf:
## Summary

Fixes three first-run issues discovered when following the `apm init → apm run start` flow end-to-end on a fresh Windows installation.

Closes #602, closes #603.

### Changes

**`src/apm_cli/commands/init.py`**
- Replace Rich `Confirm.ask()` + try/except fallback with a single `click.confirm()` call for the "Continue and overwrite?" prompt. Rich fails to read the input on CP950/CP936 terminals, retries silently, then raises, causing the user to see the prompt three times.
- Remove `apm compile` from the non-plugin Next Steps. It generates `AGENTS.md`/`CLAUDE.md` for a separate multi-agent workflow and has nothing to do with `apm run`.
- Add `start.prompt.md` to the Created Files panel so users know it was created.

**`src/apm_cli/commands/_helpers.py`**
- Create `start.prompt.md` automatically inside `_create_minimal_apm_yml()` for non-plugin projects. Without this file `apm run start` fails immediately with a "prompt not found" error, making the advertised workflow unreachable for new users.
- Use the correct `${input:param_name}` placeholder syntax in the template. The previous `{{name}}` format is never substituted by `_substitute_parameters`.

**`tests/unit/test_init_command.py`**
- Assert that `start.prompt.md` is created during `apm init` (was previously untested).
- Remove references to "minimal mode" (no longer accurate after this change).

## Test Plan

- [ ] `uv run pytest tests/unit/test_init_command.py -x -v` — all tests pass
- [ ] `apm init` on Windows CP950 — confirm prompt appears once
- [ ] `apm init` on a project that already has `apm.yml`, answer `n` — exits cleanly after one prompt
- [ ] Fresh `apm init my-project` — verify `start.prompt.md` is created and listed in output
- [ ] Follow the printed Next Steps — `apm run start` succeeds without extra setup


sergio-sisternes-epam:
Thank you for this contribution @edenfunf! The three issues you identified here (triple confirm on CP950, incorrect next steps, missing `start.prompt.md`) are real pain points for new users.

We've since addressed these same fixes in #649 which covers the same scope (closes #602, #603). To avoid merge conflicts and duplicated work, we're closing this PR in favor of #649.

Your bug reports and analysis were instrumental in getting these fixed -- appreciate the effort!

PR #609, opened 2026-04-07
fix: rewrite Windows backslash paths in hook commands

### LINKED ISSUE #520: Bug: Hook integrator does not process 'windows' property – scripts not copied and paths not rewritten
## Bug Description

When a hook file (`.apm/hooks/*.json`) contains a `windows` property alongside `command`, `bash`, or `powershell`, the `windows` property is **not processed** during compilation/installation. This means:

1. Script files referenced in the `windows` property are **not copied** to `.github/hooks/scripts/`
2. Paths in the `windows` property are **not rewritten** to the installed location (e.g. `.github/hooks/scripts/dotnet/secrets-scanner/scan-secrets.ps1`)

Only the `command`, `bash`, and `powershell` properties work correctly.

## Root Cause Analysis

**File:** `src/apm_cli/integration/hook_integrator.py` — **lines 234 and 248**

The hook path-rewriting code iterates over a hardcoded tuple of keys that does not include `"windows"`:

```python
# Line 234 – GitHub Copilot flat format (matcher-level)
for key in ("command", "bash", "powershell"):
    if key in matcher:
        new_cmd, scripts = self._rewrite_command_for_target(
            matcher[key], package_path, package_name, target,
            hook_file_dir=hook_file_dir,
        )
        matcher[key] = new_cmd
        all_scripts.extend(scripts)

# Line 248 – Claude nested format (hooks array)
for key in ("command", "bash", "powershell"):
    if key in hook:
        new_cmd, scripts = self._rewrite_command_for_target(
            hook[key], package_path, package_name, target,
            hook_file_dir=hook_file_dir,
        )
        hook[key] = new_cmd
        all_scripts.extend(scripts)
```

Because `"windows"` is missing from the tuple in both locations, the `_rewrite_command_for_target()` function is never called for that key, so scripts are not copied and paths are not rewritten.

### History

Commit `39e4c4d` previously fixed this same class of bug by adding `bash` and `powershell` to the tuple (previously only `command` was handled). The `windows` key was overlooked in that fix.

## Suggested Fix

Add `"windows"` to the key tuple in both locations (lines 234 and 248):

```python
for key in ("command", "bash", "powershell", "windows"):
```

Or more robustly, consider dynamically processing all keys that may contain script path references to prevent similar omissions in the future.

## Steps to Reproduce

1. Create a package with a hook in `.apm/hooks/` that has a `windows` property pointing to a script
2. Run `apm install` or `apm compile`
3. Observe that files from `command` are copied to `.github/hooks/` but files from `windows` are not
4. Observe that `windows` paths remain as original relative paths instead of being rewritten

danielmeppiel:
Fixes #520

## Problem

The `windows` key in hook JSON uses backslash paths (`.\\secrets-scanner\\scan-secrets.ps1`), but `_rewrite_command_for_target()` only matched Unix-style `./` references. Result: `command` paths were rewritten correctly but `windows` paths were silently skipped -- scripts not copied, paths not rewritten.

Reported by @dstamand-msft with a concrete reproduction showing the deployed hook JSON has the `command` path rewritten but `windows` still pointing to the original relative path.

## Fix

Extend the relative-path regex from `r'(\./[^\s]+)'` to `r'(\.[/\\\\][^\s]+)'` to match both `./` and `.\` prefixes. Backslashes are normalized to forward slashes for path resolution and file lookup.

**Before:**
```json
{
  "command": ".github/hooks/scripts/pkg/scan.sh",
  "windows": "pwsh -File .\\\\secrets-scanner\\\\scan.ps1"
}
```

**After:**
```json
{
  "command": ".github/hooks/scripts/pkg/scan.sh",
  "windows": "pwsh -File .github/hooks/scripts/pkg/secrets-scanner/scan.ps1"
}
```

## Tests

4 new tests covering:
- Backslash relative path rewrite (`.\\path\\script.ps1`)
- End-to-end flat format with mixed forward/backslash
- End-to-end nested Claude format with mixed forward/backslash
- Forward-slash regression (still works)

3678 tests pass.


sergio-sisternes-epam (approved):
Clean fix -- regex extended to catch both `./` and `.\` prefixes, backslashes normalized before resolution, traversal guard in place. Tests are thorough. LGTM.

PR #630, opened 2026-04-08
Add gh auth token fallback before git credential fill

awakecoding:
## Summary

This PR is intended to disambiguate GitHub authentication based on the repository URL when APM resolves credentials for package fetches.

The motivating issue is Git Credential Manager account-selection prompts on machines that use more than one GitHub account. In my case, APM was attempting authentication in a way that was not specific enough to the target repository, which caused GCM to ask which account to use.

With these changes, I was able to configure different default user accounts for different GitHub repository URLs and use APM to fetch from private repositories without getting the account-selection prompt.

## What changed

- threads repository path context through APM auth resolution
- includes repo-path context in git credential fill lookups when available
- adds a GitHub CLI fallback before the generic credential-helper path for GitHub hosts
- updates tests and auth documentation accordingly

## Why this exists

The goal is to avoid ambiguous host-only credential resolution for GitHub package fetches, especially on developer machines that are signed into multiple GitHub accounts.

The key idea is that the repository URL contains enough context to help select the correct identity, and in practice these changes worked for me.

## Notes for reviewers

These patches were vibe coded while investigating the issue end to end. Please feel free to close this PR in favor of a different implementation if the same behavior should be achieved another way.

If the changes are acceptable as-is, great. If they require adjustments, I can make them.


sergio-sisternes-epam (changes requested):
**Thanks for tackling this** — multi-account credential ambiguity is a real pain point, and the `gh auth token` fallback is a genuinely great idea that we'd love to see land.

That said, we have concerns about the current scope. Auth resolution is our most security-sensitive code path, and this PR changes 8 files, alters cache key semantics from `(host, org)` to `(host, org, repo_path)`, and threads a new input (`repo_path`) into `git credential fill` stdin — which is a protocol that interprets newlines as field delimiters. That's a large review surface for a security-critical area, and we need to be extra cautious here.

**Our recommendation: land this incrementally.**

The `gh auth token --hostname <host>` fallback alone solves ~90% of multi-account cases in a fraction of the code. It doesn't require `repo_path` threading, cache key changes, or new subprocess input — just a clean new step in the resolution chain.

We'd suggest:

1. **Slim this PR** to just the `gh auth token` fallback (before `git credential fill` in the chain). That's a focused, reviewable change we can merge with confidence.
2. If users later report issues in the "no `gh` CLI, multiple GCM credentials" scenario, we can add `repo_path` pass-through to `git credential fill` as a **follow-up PR** — with proper input sanitization via our `path_security` utilities.

Would you be open to scoping down to just the `gh` fallback? We're happy to help define the boundaries if useful.

Thanks again for the contribution — the core insight here is solid.

awakecoding:
Thanks for taking the time to review the pull request and suggest modifications to be made. I'll look into reducing the scope of changes in this pull request this week

awakecoding:
Reduced the PR scope as requested in follow-up commit bcf42a4. This version keeps the GitHub CLI fallback (gh auth token --hostname <host>) ahead of git credential fill, and removes the repo-path threading, cache key expansion, and repo-path-specific tests/docs so the first batch stays focused. Focused auth validation still passes locally: 	ests/test_token_manager.py and 	ests/unit/test_auth.py.

danielmeppiel:
Hi @awakecoding -- friendly check-in: this PR has been in `changes-requested` state since 2026-04-10, and we are working through the wave-3 merge backlog. If you are still able to address the review feedback in the next ~7 days, we would love to land this; otherwise we will close it as stale and you can re-open later (no hard feelings -- contribution graveyard happens to everyone).

Scope the change down for this security-critical auth path -- the current diff touches 8 files, alters the resolver cache key from `(host, org)` to `(host, org, repo_path)`, and threads `repo_path` into `git credential fill` stdin (a newline-delimited protocol); reviewers asked to split / shrink before merge.

Thanks for the contribution!

awakecoding:
Current branch scope is narrower than the original review summary described.

The current PR diff is 6 files and no longer includes the repo-path work that prompted the earlier concern:
- no `repo_path` threading into `git credential fill`
- no `(host, org, repo_path)` resolver cache key
- no install/marketplace plumbing for repo-context auth

What remains in this PR is the GitHub CLI fallback before `git credential fill`:
- `GH_TOKEN` remains in the GitHub module token chain
- `gh auth token --hostname <host>` is tried before `git credential fill`
- matching tests/docs were updated to reflect that narrower chain

Current changed files:
- `src/apm_cli/core/token_manager.py`
- `src/apm_cli/core/auth.py`
- `tests/test_token_manager.py`
- `tests/unit/test_auth.py`
- `docs/src/content/docs/getting-started/authentication.md`
- `packages/apm-guide/.apm/skills/apm-usage/authentication.md`

Focused validation still passes locally:
- `$env:PYTHONPATH='.;src'; uv run python -m pytest tests/test_token_manager.py tests/unit/test_auth.py -x`

If the remaining 6-file diff still feels too broad, I can split out the doc/test-only follow-up separately, but the repo-context auth changes are already removed from this PR.

danielmeppiel:
Acted on the panel review in commit 74d4e782 (pushed to `fix/disambiguate-git-credential-fill`):

**Blocking (doc-writer)**
- Fixed the orphaned table row + mid-table prose in `packages/apm-guide/.apm/skills/apm-usage/authentication.md`. Table is now contiguous (7 rows), prose moved below.

**Code (auth-expert)**
- Folded `_supports_gh_cli_host` into `resolve_credential_from_gh_cli` so both call sites (`_resolve_token` + `_try_credential_fallback`) share one eligibility path; closes the GHES-default-host divergence.
- Added `'gh-auth-token'` to the `_try_credential_fallback` short-circuit set; no double-invoke when gh CLI was already the source.
- Hardened the gh subprocess: `stdin=subprocess.DEVNULL`, `GH_NO_UPDATE_NOTIFIER=1`, debug-log of stderr on non-zero exit.
- Per-step verbose logging ('trying gh auth token for X' then 'trying git credential fill for X') and updated docstring.

**Tests (test-coverage-expert -- 3 gaps)**
- `TestSupportsGhCliHost`: None / empty / ADO / github.com / *.ghe.com / GHES match / GHES mismatch / no `GITHUB_HOST`.
- `test_gh_cli_source_label`: asserts `AuthContext.source == 'gh-auth-token'` when gh CLI supplies the token.
- `test_try_with_fallback_uses_gh_cli`: exercises the gh-CLI branch through `try_with_fallback`, escaping the autouse disable fixture.
- Augmented existing `TestResolveCredentialFromGhCli` with assertions for `stdin=DEVNULL` + `GH_NO_UPDATE_NOTIFIER=1`, plus a guard test that ineligible hosts skip the subprocess entirely.

**Docs (devx-ux + oss-growth-hacker)**
- Reconciled prose-vs-table priority numbering in `docs/.../authentication.md`.
- Updated 'Package source behavior' table to mention `gh auth token`.
- Annotated mermaid F node 'GitHub-like hosts only'; added silent-skip note.
- Added gh CLI `:::tip` callout to `quick-start.md`.
- Reconciled stale '30s' timeout reference to '60s' to match `DEFAULT_CREDENTIAL_TIMEOUT`.

**CHANGELOG**
- Added entry under `[Unreleased] -> Added` framed as user promise, tagged `(#630)`.

**Verified**
- `uv run --extra dev ruff check src/ tests/` -- silent.
- `uv run --extra dev ruff format --check src/ tests/` -- 709 files already formatted.
- `uv run pytest tests/test_token_manager.py tests/unit/test_auth.py -q` -- 140 passed.

PR #635, opened 2026-04-09
Standardize Unicode encodings to explicitly use UTF-8

dvmukul:
### Summary of Changes

This Pull Request standardizes file operations across the codebase by adding explicit `encoding='utf-8'` to `open()` calls. 

### Rationale

On Windows, the default system encoding is often locale-specific (e.g., CP1252 or CP950), which causes `UnicodeDecodeError` when reading or writing files that contain UTF-8 characters (like prompt templates or localized marketplace metadata). This fix ensures that `apm` behaves consistently across Windows, macOS, and Linux.

### Key Fixes:
- **`script_runner.py`**: Fixed issue in prompt compilation and script reading (resolves #604).
- **`config.py`**: Standardized global configuration file operations.
- **`marketplace/client.py` & `registry.py`**: Fixed local caching of marketplace manifests.
- **`models/plugin.py`**: Ensured stable reading of `plugin.json` metadata.
- **`runtime/copilot_runtime.py`**: Fixed MCP configuration retrieval.

### Verification

Changes were audited for text-based text vs binary-based operations to avoid data corruption. Existing unit tests were prioritized to ensure no regressions in core functionality.


src/apm_cli/config.py
```diff
@@ -16,7 +16,7 @@ def ensure_config_exists():
         os.makedirs(CONFIG_DIR)
 
     if not os.path.exists(CONFIG_FILE):
-        with open(CONFIG_FILE, "w") as f:
+        with open(CONFIG_FILE, "w", encoding="utf-8") as f:
```
sergio-sisternes-epam, line 19: 👍 Good fix. Note that an AST scan of the codebase found five more text-mode `open()` calls that still lack explicit `encoding='utf-8'`:

| File | Line | Operation |
|---|---|---|
| `adapters/client/codex.py` | 65 | `toml.dump()` — write config |
| `adapters/client/codex.py` | 80 | `toml.load()` — read config |
| `adapters/client/copilot.py` | 68 | `json.dump()` — write config |
| `adapters/client/copilot.py` | 83 | `json.load()` — read config |
| `deps/github_downloader.py` | 184 | write empty gitconfig |

Since the PR title says "Standardize … across the codebase," it would be great to cover these too (same pattern, same Windows risk).

src/apm_cli/core/script_runner.py
```diff
@@ -916,7 +916,7 @@ def compile(self, prompt_file: str, params: Dict[str, str]) -> str:
         # Now ensure compiled directory exists
         self.compiled_dir.mkdir(parents=True, exist_ok=True)
 
-        with open(prompt_path, "r") as f:
+        with open(prompt_path, "r", encoding="utf-8") as f:
```
sergio-sisternes-epam, line 919: This is the core fix for #604. Consider adding a small test that round-trips a non-ASCII string (e.g. CJK characters) through `PromptCompiler.compile()` to lock in the fix — the existing `test_cli_encoding.py` only covers console stream reconfiguration, not file I/O.

sergio-sisternes-epam (commented):
**Looks good overall** — the changes are correct and directly fix the Windows `UnicodeDecodeError` from #604. Two suggestions in the inline comments (missed `open()` calls + test coverage).

danielmeppiel:
@dvmukul without you accepting the CLA as per the bot comment, we cannot merge this

PR #642, opened 2026-04-09
[fix] include dependency instructions in claude compile (#631)

edenfunf:
## What
Fix compile path that silently skipped dependency instructions stored in
`.github/instructions/` when running `apm compile --target claude` without
`--local-only`.

## Why
Issue #631: `scan_directory_with_source` only scanned `{dep}/.apm/` for
primitives. Packages that store instructions in `.github/instructions/`
(instead of `.apm/instructions/`) were completely invisible to
`discover_primitives_with_dependencies`. The `--local-only` and `--validate`
paths used `LOCAL_PRIMITIVE_PATTERNS` (which includes `**/.github/instructions/`)
so they worked, but the normal distributed compile path did not.

Symptoms:
- `apm compile --target claude --verbose` showed `Instruction Processing: 0.0ms`
- No `CLAUDE.md` was generated
- `apm compile --target claude --local-only` worked correctly
- `apm compile --validate` could find the instructions

## How
Add `DEPENDENCY_GITHUB_PRIMITIVE_PATTERNS` (mirroring `DEPENDENCY_PRIMITIVE_PATTERNS`
for the `.github/` directory layout) and scan `{dep}/.github/` in
`scan_directory_with_source` alongside the existing `.apm/` scan.

The original early-return guard `if not apm_dir.exists(): return` is replaced
by independent `if apm_dir.exists():` / `if github_dir.exists():` checks so
both directories are always considered regardless of which one is present.

No CLI interface changes. No architecture changes. Existing `.apm/`-based
packages are unaffected.

## Test
- Reproduced issue #631: `.github/`-only dependency → 0 instructions before fix
- Verified fix: `.github/`-only dependency → instruction found and `CLAUDE.md` generated
- `--local-only` still works correctly
- `--target copilot` unaffected
- `--validate` unaffected
- Regression tests added to `TestScanDirectoryWithSource`:
  - `test_github_instructions_discovered_when_no_apm_dir`
  - `test_github_instructions_discovered_alongside_apm_dir`
- All 348 existing unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pin codex setup to `rust-v0.118.0` for security and reproducibility; update config to `wire_api = "responses"` (#663)
 - Propagate headers and environment variables through OpenCode MCP adapter with defensive copies to prevent mutation (#622)
+- Fix `apm compile --target claude` silently skipping dependency instructions stored in `.github/instructions/` (#631)
 ### Changed
 
 - `apm marketplace browse/search/add/update` now route through the registry proxy when `PROXY_REGISTRY_URL` is set; `PROXY_REGISTRY_ONLY=1` blocks direct GitHub API calls (#506)
--- a/src/apm_cli/primitives/discovery.py
+++ b/src/apm_cli/primitives/discovery.py
@@ -311,6 +311,27 @@ def get_dependency_declaration_order(base_dir: str) -> List[str]:
         return []
 
 
+def _scan_patterns(base_dir: Path, patterns: Dict[str, List[str]], collection: PrimitiveCollection, source: str) -> None:
+    """Glob-scan-parse loop for one base directory and one patterns dict.
+
+    Args:
+        base_dir (Path): Directory to scan (e.g., dep/.apm or dep/.github).
+        patterns (Dict[str, List[str]]): Primitive-type → glob-pattern mapping.
+        collection (PrimitiveCollection): Collection to add primitives to.
+        source (str): Source identifier for discovered primitives.
+    """
+    for _primitive_type, type_patterns in patterns.items():
+        for pattern in type_patterns:
+            for file_path_str in glob.glob(str(base_dir / pattern), recursive=True):
+                file_path = Path(file_path_str)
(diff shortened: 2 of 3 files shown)


sergio-sisternes-epam (changes requested):
Nice fix — this is a real bug and the regression tests are well-targeted. The PR description is excellent too, clear reproduction and rationale.

Two requests before we can merge:

**1. Reduce code duplication in `scan_directory_with_source`**

The glob-scan-parse loop is copy-pasted for `.apm/` and `.github/`. Consider extracting a small helper:

```python
def _scan_patterns(base_dir: Path, patterns: Dict[str, List[str]], collection: PrimitiveCollection, source: str) -> None:
    for primitive_type, type_patterns in patterns.items():
        for pattern in type_patterns:
            for file_path_str in glob.glob(str(base_dir / pattern), recursive=True):
                file_path = Path(file_path_str)
                if file_path.is_file() and _is_readable(file_path):
                    try:
                        primitive = parse_primitive_file(file_path, source=source)
                        collection.add_primitive(primitive)
                    except Exception as e:
                        print(f"Warning: Failed to parse dependency primitive {file_path}: {e}")
```

Then the main function becomes:

```python
if apm_dir.exists():
    _scan_patterns(apm_dir, DEPENDENCY_PRIMITIVE_PATTERNS, collection, source)
if github_dir.exists():
    _scan_patterns(github_dir, DEPENDENCY_GITHUB_PRIMITIVE_PATTERNS, collection, source)
```

This keeps the logic in one place for future maintenance.

**2. Missing CHANGELOG entry**

Please add an entry under `## [Unreleased]` in `CHANGELOG.md`, e.g.:

```
### Fixed
- Fix `apm compile --target claude` silently skipping dependency instructions stored in `.github/instructions/` (#631)
```

Thanks for the contribution!

edenfunf:
@sergio-sisternes-epam Thanks for the thorough review! Done — both points addressed in the latest commit (f67705d):

1. Extracted `_scan_patterns()` helper — the duplicated glob-scan-parse loop now lives in one place; `scan_directory_with_source` calls it for both `.apm/` and `.github/`.
2. Added the `### Fixed` entry under `[Unreleased]` in CHANGELOG.md.

All 63 unit tests still pass.

sergio-sisternes-epam (approved):

PR #653, opened 2026-04-09
Enable SSH timeout to fallback to HTTP.

### LINKED ISSUE #652: [BUG] Corporate networks blocking port 22 cause SSH clone attempt to hang
**Describe the bug**

`apm install` hangs indefinitely when restoring packages from an Azure DevOps repository on a network that silently drops SSH traffic (e.g., corporate firewalls, VPN/Zscaler). The clone fallback chain in [`_clone_with_fallback()`](v) tries SSH ([Method 2](https://github.com/microsoft/apm/blob/b56920e543c0f41e8ab97a157d6394f0ac15b5d5/src/apm_cli/deps/github_downloader.py#L680)) before plain HTTPS ([Method 3](https://github.com/microsoft/apm/blob/b56920e543c0f41e8ab97a157d6394f0ac15b5d5/src/apm_cli/deps/github_downloader.py#L691)), but `Repo.clone_from()` via GitPython has no connection timeout. When the firewall silently drops packets on port 22 instead of rejecting them, the SSH attempt blocks forever and the HTTPS fallback is never reached.

This affects users who rely on git credential helpers (e.g., Git Credential Manager, Azure CLI) rather than explicit `ADO_APM_PAT` tokens, since having no token means [Method 1](https://github.com/microsoft/apm/blob/b56920e543c0f41e8ab97a157d6394f0ac15b5d5/src/apm_cli/deps/github_downloader.py#L666) (authenticated HTTPS) is skipped.

**To Reproduce**

1. Be on a corporate/VPN network that silently drops outbound SSH traffic (port 22) like Zscaler.
2. Have an `apm.yml` with an Azure DevOps dependency.
3. Do not set `ADO_APM_PAT` (rely on git credential helpers instead).
4. Run `apm install`.
5. The command hangs indefinitely at the SSH clone attempt.

**Expected behavior**

The SSH clone attempt should fail within a reasonable timeout (e.g., 30 seconds) and fall back to plain HTTPS, where git credential helpers can provide authentication.

Alternatively, HTTPS could be attempted first, before SSH, to allow credential helpers to kick in.

- OS: macOS (Apple Silicon), though the issue affects any platform behind a firewall that drops SSH
- Python Version: 3.13
- APM Version: 0.8.11
- VSCode Version: N/A (CLI issue)

**Logs**

Running `apm install -v` gives me this output...

apm install -v output

**Additional context**

- The [`_try_sparse_checkout()`](https://github.com/microsoft/apm/blob/b56920e543c0f41e8ab97a157d6394f0ac15b5d5/src/apm_cli/deps/github_downloader.py#L1685) method already uses [`subprocess.run(timeout=120)`](https://github.com/microsoft/apm/blob/b56920e543c0f41e8ab97a157d6394f0ac15b5d5/src/apm_cli/deps/github_downloader.py#L1715), so sparse-checkout operations are protected. Only the GitPython `Repo.clone_from()` path lacks a timeout.
- The `GIT_SSH_COMMAND` environment variable is not set in the git environment, so there is no ConnectTimeout to bound the SSH connection attempt.
- A workaround is to set `url.<base>.insteadOf` in gitconfig to rewrite SSH URLs to HTTPS, but this only works when APM uses the relaxed env (no token). When APM has a token and uses the locked-down env (`GIT_CONFIG_GLOBAL=/dev/null`), user gitconfig is suppressed.

tillig:
## Description

Allow SSH to timeout in the case of corporate VPN/firewalls that silently drop packets for port 22 so Git credential helpers and HTTPS can execute.

Fixes #652

Additional positive side effects:

- Authentication: You are no longer forced to use a PAT to authenticate with Azure DevOps or GitHub Enterprise. The fallback makes it "just work."
- GitHub Enterprise:
  - You no longer need to set `GITHUB_HOST` for working with internal GitHub Enterprise solutions. That also means...
  - If you have _multiple_ internal GitHub Enterprise endpoints (we do!) you don't have to mess with switching this around.

I've verified both of these additional scenarios on our internal GHE instance. Things start to "just work."

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

Note: There is no `.editorconfig` in the repo and there is a mix/match of "trim trailing whitespace" and "leave trailing whitespace alone" in the code. When I made my updates, my editor was set to trim, so you'll see whitespace changes.

I'm happy to create a follow-up PR with an `.editorconfig` to help standardize formatting in the repo. Or, if the whitespace truncation appears to be "too much," I can roll those portions of the PR back.


sergio-sisternes-epam (changes requested):
Excellent contribution — this is a real problem on corporate networks and the fix is clean and well-tested. The docs updates (both user-facing and skill guide) are a nice touch.

The whitespace changes inflate the diff but the core logic is surgical. No concerns there — an `.editorconfig` follow-up would be welcome to standardize going forward (we'll create a tracking issue for that).

One request before merge: please add a CHANGELOG entry under `## [Unreleased]`:

```
### Fixed
- Fix `apm install` hanging indefinitely when corporate firewalls silently drop SSH packets by setting `GIT_SSH_COMMAND` with `ConnectTimeout=30` (#652)
```

Thanks for the thorough testing on your GHE instance — that kind of real-world verification is invaluable.

tillig:
Thanks! CHANGELOG entry added. Let me know if there's anything else I can do.

sergio-sisternes-epam (approved):

sergio-sisternes-epam (approved):

PR #660, opened 2026-04-10
feat: make install.sh configurable for air-gapped environments

chkp-roniz:
## Summary
- `APM_INSTALL_DIR` env var to override install path (default `/usr/local/bin`)
- `GITHUB_URL` env var to override GitHub base URL for mirrors/GHE
- `REPO` env var to override repository (default `microsoft/apm`)
- `VERSION` env var or `@v1.2.3` arg to pin a specific version — skips GitHub API entirely when set

## Usage
```bash
# Specific version (no GitHub API call needed)
curl -sSL https://aka.ms/apm-unix | sh -s -- @v0.8.10

# Custom install directory
curl -sSL https://aka.ms/apm-unix | APM_INSTALL_DIR=~/.local/bin sh

# Air-gapped / internal mirror
GITHUB_URL=https://artifactory.corp.com/artifactory/github VERSION=v0.8.10 sh install.sh
```

## Test plan
- [x] Tested with custom `GITHUB_URL` + `VERSION=v0.8.10` — downloads from mirror, skips API
- [x] Tested with unreachable API — confirmed API is skipped when VERSION is set
- [x] Tested with `APM_INSTALL_DIR=$HOME/.local/bin` — installs to custom dir without sudo
- [x] `bash -n install.sh` syntax check passes
- [x] No hardcoded `https://github.com` outside Configuration defaults


sergio-sisternes-epam (commented):
**Great contribution, @chkp-roniz!** Air-gapped and GHE support for the installer is a real need. A few things to address before we can merge:

### Must-fix

1. **GitHub API URL is hardcoded** — `https://api.github.com/repos/$REPO/releases/latest` doesn't change when `GITHUB_URL` points to GHE (API lives at `$GITHUB_URL/api/v3/...`). Either derive a `GITHUB_API_URL` or document that `VERSION` is mandatory for GHE setups.

2. **`REPO` env var is too generic** — Rename to `APM_REPO` for consistency with `APM_INSTALL_DIR` and to avoid collisions with the user's environment.

3. **Missing CHANGELOG entry** — Please add a line under `## [Unreleased] / ### Added` per our [contributing guide](https://github.com/microsoft/apm/blob/main/CONTRIBUTING.md).

### Should-fix

4. **Document `APM_LIB_DIR` derivation** — The bundle goes to `$(dirname "$APM_INSTALL_DIR")/lib/apm`, which is reasonable but undiscoverable. Add it to the docs table so users know where the full binary directory lands.

5. **Non-ASCII characters** (`✓`, `⚠`, `🎉`) — These pre-date your PR, but since you're touching those lines anyway, would you mind swapping them for ASCII equivalents (`[+]`, `[!]`, etc.)? We keep output within printable ASCII for Windows cp1252 compatibility.

### Follow-ups (we'll open issues for these)

6. **`apm config` integration** — `install.sh` is the bootstrap, so env vars are the right call here. But once installed, users will expect `apm update` to remember their custom install dir and mirror URL without re-exporting env vars. We should persist these settings via `apm config` (e.g., `install-dir`, `mirror-url`) and have the self-update path read them. Env vars would still override for CI/unattended scenarios.

7. **Windows parity** — `install.ps1` should support the same configurability (`APM_INSTALL_DIR`, `GITHUB_URL`, `APM_REPO`, `VERSION`) so Windows users in air-gapped/GHE environments get the same experience.

sergio-sisternes-epam (approved):

sergio-sisternes-epam (dismissed):
Apologies, my previous feedback was not fixed yet. Please review my it.

danielmeppiel (changes requested):
Thanks @chkp-roniz — this is a real user need and the diff is in good shape overall. CI is green and most of @sergio-sisternes-epam's earlier feedback is addressed (`APM_REPO` rename, CHANGELOG entry, docs table). Three small items to fix before merge:

### Must-fix

**1. `APM_LIB_DIR` is documented as overridable but the script ignores user-supplied values.**

The docs table in `installation.md` lists `APM_LIB_DIR` with a default of `$(dirname APM_INSTALL_DIR)/lib/apm`, implying users can override it. But in `install.sh`:

```bash
APM_LIB_DIR="$(dirname "$APM_INSTALL_DIR")/lib/apm"
```

This unconditionally overwrites any value the user sets. Either:
- Honour the env var: `APM_LIB_DIR="${APM_LIB_DIR:-$(dirname "$APM_INSTALL_DIR")/lib/apm}"`, **or**
- Change the doc row to mark it derived/read-only.

The first option is more useful — a user who sets `APM_INSTALL_DIR=$HOME/bin` today gets the bundle at `$HOME/lib/apm`, which is surprising.

**2. Status symbols are missing the trailing space.**

Per the codebase convention (`STATUS_SYMBOLS` in `src/apm_cli/utils/console.py`), bracket symbols are always followed by a single space. Current output reads:

```
[+]APM installed successfully!
[!]APM installed but not found in PATH
[!]Compatibility Issue Detected
[!]glibc version incompatibility detected
[!]Container/Dev Container environment detected
[+]Download successful
[+]Extraction successful
[+]Binary test successful
```

Should be `[+] APM installed successfully!`, etc. ~10 sites in `install.sh`.

**3. Spurious "will need sudo" notice for non-existent custom dirs.**

The writability check runs before `mkdir -p`:

```bash
if [ ! -w "$APM_INSTALL_DIR" ]; then
    echo -e "${YELLOW}Note: Will need sudo permissions to install to $APM_INSTALL_DIR${NC}"
fi
```

If a user sets `APM_INSTALL_DIR=$HOME/.local/bin` and that dir doesn't exist yet, `[ ! -w ... ]` is true and they see a misleading sudo warning even though `mkdir -p` later succeeds without it. Suggest:

```bash
if [ -e "$APM_INSTALL_DIR" ] && [ ! -w "$APM_INSTALL_DIR" ]; then
    ...
fi
```

### Acceptable as-is

- Hardcoded `api.github.com` for latest-release lookup. The doc note explicitly tells GHE/air-gapped users to set `VERSION=`, which is a reasonable workaround for a bootstrap installer. A proper `GITHUB_API_URL` (or derivation from `GITHUB_URL`) would be cleaner but is fine as a follow-up.

### Nits (optional)

- The header comment block (lines 4–6) documents `APM_INSTALL_DIR`, `GITHUB_URL`, and `@v1.2.3` but omits `APM_REPO` and `VERSION` env var. Quick consistency tidy.

### Follow-ups (not for this PR)

Sergio's earlier suggestions still stand for separate issues:
- `apm config` integration so `apm update` remembers custom install dir / mirror without re-exporting env vars.
- `install.ps1` parity for Windows users in air-gapped/GHE environments.

Once items 1–3 land, ping a maintainer to dismiss the stale `CHANGES_REQUESTED` and approve.

danielmeppiel (approved):

PR #683, opened 2026-04-12
fix: marketplace add fails to authenticate for private repos

### LINKED ISSUE #669: apm marketplace add fails to authenticate when fetching marketplace.json from private repositories
## Summary

`apm marketplace add` cannot register a private GitHub repository as a marketplace. It reports "No marketplace.json found" despite the file existing at the repo root, because the command does not apply authentication when fetching the file. Other commands like `apm install` authenticate correctly against the same private repo.

## Environment

- APM CLI version: 0.8.11
- OS: macOS (darwin, arm64)
- Auth: `gh auth login` active, token available via `git credential fill`, `GITHUB_TOKEN`, and `GITHUB_APM_PAT`

## Steps to reproduce

1. Have a **private** GitHub repository with a valid `marketplace.json` at the root
2. Ensure authentication is configured (any of: `GITHUB_TOKEN`, `GITHUB_APM_PAT`, `GITHUB_APM_PAT_{ORG}`, or `git credential fill`)
3. Run:

```bash
GITHUB_TOKEN=ghp_valid_token apm marketplace add OWNER/REPO -v
```

## Expected behavior

APM uses the available token to fetch `marketplace.json` from the private repo and registers the marketplace.

## Actual behavior

```
[*] Registering marketplace 'REPO'...
    Repository: OWNER/REPO
    Branch: main
[x] No marketplace.json found in 'OWNER/REPO'. Checked:
marketplace.json, .github/plugin/marketplace.json,
.claude-plugin/marketplace.json
```

## Evidence that auth works for other commands

The same token and repo work fine with `apm install`:

```bash
GITHUB_TOKEN=ghp_valid_token apm install OWNER/REPO --dry-run
# [+] OWNER/REPO — validates successfully
```

And the file is confirmed to exist via the GitHub API:

```bash
gh api repos/OWNER/REPO/contents/marketplace.json
# Returns valid marketplace.json content
```

## Tokens tested (all fail identically)

- `GITHUB_TOKEN`
- `GITHUB_APM_PAT`
- `GITHUB_APM_PAT_{ORG}`
- `git credential fill` (confirmed working)

## Root cause analysis

We traced this through the source code. The bug is a semantic mismatch between "file not found" and "repo not accessible" in `_fetch_file()`.

### The auth fallback doesn't trigger on 404

`_fetch_file()` in `src/apm_cli/marketplace/client.py:138-176` uses the GitHub Contents API with `unauth_first=True`:

```python
return auth_resolver.try_with_fallback(
    source.host,
    _do_fetch,
    org=source.owner,
    unauth_first=True,    # <--- tries without token first
)
```

The fallback mechanism in `AuthResolver.try_with_fallback()` (`auth.py:278-290`) only escalates to an authenticated retry **on exceptions**:

```python
if unauth_first:
    try:
        return operation(None, git_env)          # 1. Try with NO token
    except Exception:
        if auth_ctx.token:
            return operation(auth_ctx.token, git_env)  # 2. Retry WITH token
        raise
```

But `_do_fetch` handles HTTP 404 by returning `None` instead of raising:

```python
# client.py:158-159
if resp.status_code == 404:
    return None
```

For private repos, GitHub returns **404** (not 403) to avoid leaking repo existence. Since `_do_fetch` returns `None` rather than raising an exception, the auth fallback never fires. The `None` propagates back and APM concludes the file doesn't exist — when it actually never tried with a token.

### Why `apm install` works with the same private repo

`apm install` uses `GitHubPackageDownloader` which:
- Resolves tokens **eagerly** per-dependency via `_resolve_dep_token()`
- Uses `git clone` with the **token embedded in the HTTPS URL**
- Tries **authenticated first** (`auth → SSH → unauth`) — the inverse of marketplace's order
- Git raises actual exceptions on failure, so the fallback chain works correctly

| Aspect | `marketplace add` | `apm install` |
|--------|-------------------|---------------|
| Protocol | GitHub REST API (Contents) | `git clone` over HTTPS |
| Auth strategy | `unauth_first=True` | Auth-first (token in URL) |
| Fallback direction | unauth → auth → credential-fill | auth → SSH → unauth |
| 404 handling | Returns `None` (no retry) | Git raises exception (triggers fallback) |

### Suggested fixes (in order of preference)

1. **Make `_do_fetch` raise on 404 during the unauthenticated phase**, so `try_with_fallback` escalates to the authenticated retry. Only treat 404-with-a-valid-token as "file not found." This preserves the rate-limit optimization for public repos while fixing private repos.

2. **Use `unauth_first=False`** for the marketplace fetch — try authenticated first, like `apm install` does. Simplest change but uses a token request even for public repos.

3. **Two-phase approach**: first verify repo accessibility (`GET /repos/{owner}/{repo}`), then probe for files.

## Workaround

Install plugins directly using subdirectory syntax, bypassing the marketplace:

```bash
apm install OWNER/REPO/plugin-subdir
```

lawrence3699:
## Summary

Fixes #669

- `apm marketplace add` fails with "No marketplace.json found" for private GitHub repos because `_do_fetch()` returns `None` on HTTP 404, preventing `try_with_fallback()` from retrying with credentials
- GitHub returns 404 (not 403) for private repos on unauthenticated requests to avoid leaking repo existence
- Fix: when `_do_fetch` receives a 404 **without a token**, raise via `raise_for_status()` so the auth fallback triggers. Authenticated 404s still return `None` (file genuinely not found)

## Changes

**`src/apm_cli/marketplace/client.py`** (5 lines added):
- In `_do_fetch`, distinguish unauthenticated 404 (raise to trigger fallback) from authenticated 404 (return None)

**`tests/unit/marketplace/test_marketplace_client.py`** (76 lines added):
- `test_unauthenticated_404_triggers_auth_retry`: verifies the auth fallback fires on unauth 404
- `test_authenticated_404_returns_none`: verifies genuine 404 with token returns None

## Test plan

- [x] All 3824 existing unit tests pass
- [x] New regression tests cover both code paths (unauth 404 → retry, auth 404 → None)
- [ ] Manual test with a private repo: `GITHUB_TOKEN=... apm marketplace add OWNER/PRIVATE_REPO`

### DIFF
--- a/src/apm_cli/marketplace/client.py
+++ b/src/apm_cli/marketplace/client.py
@@ -229,6 +229,11 @@ def _fetch_file(
             headers["Authorization"] = f"token {token}"
         resp = requests.get(url, headers=headers, timeout=30)
         if resp.status_code == 404:
+            if not token:
+                # Unauthenticated 404 may mean "private repo" (GitHub
+                # returns 404 instead of 403 to avoid leaking existence).
+                # Raise so try_with_fallback retries with credentials.
+                resp.raise_for_status()
             return None
         resp.raise_for_status()
         return resp.json()
--- a/tests/unit/marketplace/test_marketplace_client.py
+++ b/tests/unit/marketplace/test_marketplace_client.py
@@ -5,6 +5,7 @@ import time
 from unittest.mock import MagicMock, patch
 
 import pytest
+import requests
 
 from apm_cli.marketplace.errors import MarketplaceFetchError
 from apm_cli.marketplace.models import MarketplaceSource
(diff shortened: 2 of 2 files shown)


lawrence3699:
Closing — Microsoft CLA not signed. Will resubmit after completing the CLA process if still needed.

PR #688, opened 2026-04-13
fix: marketplace add authenticates for private repos

### LINKED ISSUE #669: apm marketplace add fails to authenticate when fetching marketplace.json from private repositories

jacobokeeffe-ow:
## Description

Fix `apm marketplace add` failing to fetch `marketplace.json` from private repositories, even when valid authentication tokens are available.

Fixes #669

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Root cause

`_do_fetch()` in `client.py` returned `None` on HTTP 404 regardless of whether a token was present. Since `try_with_fallback(unauth_first=True)` only escalates to authenticated retry on **exceptions** (not `None` returns), the auth fallback never fired for private repos. GitHub returns 404 (not 403) for private repos to avoid leaking existence, making the unauthenticated 404 indistinguishable from "file not found."

## Fix

When `_do_fetch()` receives a 404 **without a token**, it now raises via `resp.raise_for_status()` so that `try_with_fallback` catches the exception and retries with authentication. When a 404 is received **with a token**, it still returns `None` (the file genuinely doesn't exist).

This preserves the rate-limit optimization for public repos (unauthenticated requests succeed on first try) while correctly handling private repos.

## Testing

- [x] Tested locally
- [x] All existing tests pass (3827 passed)

4 new tests in `TestPrivateRepoAuth`:
- `test_private_repo_unauthenticated_404_retries_with_token` — core regression test
- `test_authenticated_404_returns_none` — genuinely missing files still return `None`
- `test_public_repo_unauthenticated_success` — no regression for public repos
- `test_auto_detect_private_repo_finds_marketplace_json` — end-to-end through `_auto_detect_path`

Also verified against a real private marketplace repo — the fix correctly resolves `marketplace.json` and returns both plugins.

### DIFF
--- a/src/apm_cli/marketplace/client.py
+++ b/src/apm_cli/marketplace/client.py
@@ -18,7 +18,12 @@ from typing import Dict, List, Optional
 import requests
 
 from .errors import MarketplaceFetchError
-from .models import MarketplaceManifest, MarketplacePlugin, MarketplaceSource, parse_marketplace_json
+from .models import (
+    MarketplaceManifest,
+    MarketplacePlugin,
+    MarketplaceSource,
+    parse_marketplace_json,
+)
 from .registry import get_registered_marketplaces
 
 logger = logging.getLogger(__name__)
@@ -173,7 +178,9 @@ def _try_proxy_fetch(
     except (json.JSONDecodeError, ValueError):
         logger.debug(
             "Proxy returned non-JSON for %s/%s %s",
-            source.owner, source.repo, file_path,
+            source.owner,
+            source.repo,
+            file_path,
         )
         return None
 
@@ -213,7 +220,9 @@ def _fetch_file(
     if cfg is not None and cfg.enforce_only:
         logger.debug(
             "PROXY_REGISTRY_ONLY blocks direct GitHub fetch for %s/%s %s",
-            source.owner, source.repo, file_path,
+            source.owner,
+            source.repo,
+            file_path,
         )
         return None
 
@@ -229,6 +238,12 @@ def _fetch_file(
             headers["Authorization"] = f"token {token}"
         resp = requests.get(url, headers=headers, timeout=30)
         if resp.status_code == 404:
+            if not token:
+                # Unauthenticated 404 is ambiguous: could be a genuinely
+                # missing file *or* a private repo hiding its existence.
+                # Raise so that ``try_with_fallback`` retries with a token.
+                resp.raise_for_status()
+            # Authenticated 404 means the file genuinely does not exist.
             return None
         resp.raise_for_status()
         return resp.json()
--- a/tests/unit/marketplace/test_marketplace_client.py
+++ b/tests/unit/marketplace/test_marketplace_client.py
@@ -5,6 +5,7 @@ import time
 from unittest.mock import MagicMock, patch
 
 import pytest
+import requests
 
 from apm_cli.marketplace.errors import MarketplaceFetchError
 from apm_cli.marketplace.models import MarketplaceSource
@@ -16,7 +17,9 @@ def _isolate_cache(tmp_path, monkeypatch):
     """Point cache and config to temp directories."""
     config_dir = str(tmp_path / ".apm")
     monkeypatch.setattr("apm_cli.config.CONFIG_DIR", config_dir)
-    monkeypatch.setattr("apm_cli.config.CONFIG_FILE", str(tmp_path / ".apm" / "config.json"))
+    monkeypatch.setattr(
+        "apm_cli.config.CONFIG_FILE", str(tmp_path / ".apm" / "config.json")
+    )
     monkeypatch.setattr("apm_cli.config._config_cache", None)
     monkeypatch.setattr("apm_cli.marketplace.registry._registry_cache", None)
     yield
@@ -85,7 +88,9 @@ class TestFetchMarketplace:
         }
         mock_resolver = MagicMock()
         mock_resolver.try_with_fallback.return_value = raw_data
-        mock_resolver.classify_host.return_value = MagicMock(api_base="https://api.github.com")
+        mock_resolver.classify_host.return_value = MagicMock(
+            api_base="https://api.github.com"
+        )
 
         manifest = client_mod.fetch_marketplace(
             source, force_refresh=True, auth_resolver=mock_resolver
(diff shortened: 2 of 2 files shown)


jacobokeeffe-ow:
@microsoft-github-policy-service agree company="Oliver Wyman"

jacobokeeffe-ow:
@sergio-sisternes-epam apologies for the formatting changes but when I run `uv run black .` I end up with 308 files reformatted. I followed the instructions in the contributing guide and don't see anything obvious I'm doing wrong.

I might be mistaken but I don't see `black` being run as part of CI so I'm wondering if the whole repo genuinely needs reformatting or if maybe the instructions just need updating - let me know!

sergio-sisternes-epam (approved):
Clean fix -- the 6-line change in `_do_fetch()` correctly distinguishes "private repo hiding existence" from "file not found" by raising on unauthenticated 404 to trigger `try_with_fallback` auth retry. Good test coverage with 4 regression tests.

Re: formatting -- we don't enforce `black` in CI, so the reformatting here is fine (contained to files you touched). We should update the contributing guide to clarify this. Thanks for flagging it!

**Note:** CI hasn't run yet (fork PR). Would be good to get Build & Test triggered before merge.

jacobokeeffe-ow:
thanks @sergio-sisternes-epam.

I think you might need to give approval in order the get the github action to run? https://github.com/microsoft/apm/actions/runs/24339327316 : "This workflow is awaiting approval from a maintainer"

sergio-sisternes-epam:
Hi @jacobokeeffe-ow -- thank you for this contribution, and especially for the exceptional root cause analysis in #669. Your detailed investigation of the `unauth_first` / 404 behavior, the comparison table with `apm install`, and the three suggested fix approaches directly informed our fix.

While this PR was in review, we merged #701 which addresses the same bug via your suggested fix #2 (`unauth_first=False`). A duplicate report (#693) was also filed and closed by #701. Since the two approaches are incompatible -- your raise-on-unauth-404 logic targets the `unauth_first=True` flow that no longer exists on `main` -- I'm closing this PR.

I'm also closing #669 with credit to your root cause analysis, which was the foundation for the fix.

Your contributions here have been genuinely valuable to the project. We'd love to keep collaborating with you on future issues -- feel free to pick up anything that catches your eye: https://github.com/microsoft/apm/issues

Thanks again for your time and thoroughness!

PR #700, opened 2026-04-14
feat: support allow-insecure HTTP dependencies

arika0093:
## Description
This PR adds explicit support for HTTP (insecure) APM dependencies behind an opt-in flow, and makes lockfile replays preserve that HTTP source information.

When adding a package, you can explicitly allow HTTP dependencies by providing the `--allow-insecure` flag.

```bash
$ apm install --allow-insecure http://my-server.example.com/owner/repo

# if not provided, it will error like this:
# [*] Validating 1 package...
# [x] http://my-server.example.com/owner/repo -- 'http://my-server.example.com/owner/repo' uses HTTP (insecure). Pass '--allow-insecure' or run 'apm config set allow-insecure true' to allow HTTP dependencies.
# All packages failed validation. Nothing to install.
```

When restoring packages, you also need to provide the `--allow-insecure` flag unless the global config is enabled. If there are HTTP URLs in the dependency chain, it will error without `--allow-insecure` or `apm config set allow-insecure true`.

```bash
$ apm install --allow-insecure

# if not provided, it will error like this:
# [>] Installing dependencies from apm.yml...
# [x] Dependency 'my-server.example.com/owner/repo' uses HTTP (insecure). Pass '--allow-insecure' to apm install, or run 'apm config set allow-insecure true' to allow HTTP dependencies globally.
```

The `apm.yml` file will record it in the following format:

```yaml
dependencies:
  apm:
    - git: http://my-server.example.com/owner/repo
      allow_insecure: true
```

HTTP dependencies require two explicit signals:
- `allow_insecure: true` on the dependency entry in `apm.yml`
- either `apm install --allow-insecure` or `apm config set allow-insecure true`

The global configuration only removes the need to pass `--allow-insecure` on the CLI. It does not remove the `allow_insecure: true` requirement in `apm.yml`.

```bash
apm config set allow-insecure true
```

Fixes #636

### TODO
- [x] Support for `--allow-insecure` flag when adding dependencies
- [x] Support for `--allow-insecure` flag when installing dependencies
- [x] Support for `allow_insecure: true` field in `apm.yml` and lockfile
- [x] Support for `allow-insecure` global config

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass


arika0093:

src/apm_cli/models/dependency/reference.py
```diff
@@ -986,8 +1001,29 @@ def parse(cls, dependency_str: str) -> "DependencyReference":
...
+            str or dict: String for HTTPS/SSH/local deps; dict for HTTP deps.
+        """
+        if self.is_insecure:
+            host = self.host or default_host()
+            entry = {"git": f"http://{host}/{self.repo_url}"}
+            if self.reference:
+                entry["ref"] = self.reference
+            if self.alias:
+                entry["alias"] = self.alias
+            entry["allow_insecure"] = True
```
sergio-sisternes-epam: **Blocking:** This hardcodes `allow_insecure = True` for all HTTP deps, ignoring `self.allow_insecure`. If this method is reused in a code path that hasn't applied the opt-in check, it silently grants insecure access.

```suggestion
            entry["allow_insecure"] = self.allow_insecure
```

src/apm_cli/models/dependency/reference.py
```diff
@@ -986,8 +1001,29 @@ def parse(cls, dependency_str: str) -> "DependencyReference":
             ado_project=ado_project,
             ado_repo=ado_repo,
             artifactory_prefix=artifactory_prefix,
+            is_insecure=dependency_str.startswith("http://"),
```
sergio-sisternes-epam: **Suggestion:** `startswith("http://")` is case-sensitive, but URL schemes are case-insensitive per RFC 3986. `HTTP://host/repo` would bypass insecure detection.

```suggestion
            is_insecure=urllib.parse.urlparse(dependency_str).scheme.lower() == "http",
```

src/apm_cli/deps/github_downloader.py
```diff
@@ -619,11 +619,18 @@ def _build_repo_url(self, repo_ref: str, use_ssh: bool = False, dep_ref: Depende
...
             is_github = is_github_hostname(host)
+            # Use http:// only for dependencies explicitly marked as insecure.
+            is_insecure = False
+            if dep_ref is not None:
+                is_insecure = bool(getattr(dep_ref, "is_insecure", False))
             if use_ssh:
                 return build_ssh_url(host, repo_ref)
             elif is_github and github_token:
-                # Only send GitHub tokens to GitHub hosts
+                # # Only send GitHub tokens to GitHub hosts
```
sergio-sisternes-epam: **Nit:** Double `#` — typo from the refactor commit.

```suggestion
                # Only send GitHub tokens to GitHub hosts
```

src/apm_cli/models/dependency/reference.py
```diff
@@ -470,6 +483,7 @@ def parse_from_dict(cls, entry: dict) -> "DependencyReference":
         sub_path = entry.get("path")
         ref_override = entry.get("ref")
         alias_override = entry.get("alias")
+        allow_insecure = bool(entry.get("allow_insecure", False))
```
sergio-sisternes-epam: **Suggestion:** `bool(entry.get("allow_insecure", False))` treats any non-empty string (e.g., `"false"`) as `True`. Since this is a security-relevant field, consider validating the type explicitly:

```suggestion
        allow_insecure = entry.get("allow_insecure", False)
        if not isinstance(allow_insecure, bool):
            raise ValueError("'allow_insecure' field must be a boolean")
```

sergio-sisternes-epam (changes requested):
Solid feature -- good security model with dual opt-in (dep-level + CLI/config), thorough test coverage (30+ tests), clean lockfile round-trip, and well-written docs. Nice work!

One blocking issue: `to_apm_yml_entry()` hardcodes `allow_insecure = True`, which could silently escalate security posture. See inline comment.

Also: this PR silently fixes a bug where `artifactory_prefix` was missing from `DependencyReference` construction in `_helpers.py`, `plugin_exporter.py`, and `lockfile.py` -- directly related to #614. Consider referencing that issue in the PR description.

src/apm_cli/models/dependency/reference.py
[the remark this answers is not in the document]
arika0093: Addressed in commit 47b87ab1e3ffc2975dfba90c7e72737a85fd39e2.

src/apm_cli/models/dependency/reference.py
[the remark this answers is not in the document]
arika0093: address in commit 3c0122cb48f9ee994e529ae5fa0bf7beb5f39edd.

src/apm_cli/models/dependency/reference.py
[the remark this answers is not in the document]
arika0093: address in commit cd8bed0c07c5e9592dde69047954eebe0969b6e3.

src/apm_cli/commands/config.py
```diff
@@ -14,6 +14,48 @@
...
+
+
+def _parse_bool_value(value: str) -> bool:
+    """Parse a CLI boolean value."""
+    normalized = value.lower()
+    if normalized in _BOOLEAN_TRUE_VALUES:
+        return True
+    if normalized in _BOOLEAN_FALSE_VALUES:
+        return False
+    raise ValueError(f"Invalid value '{value}'. Use 'true' or 'false'.")
```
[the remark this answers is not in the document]
arika0093: address in commit 5260fbd84d376de6021dbe757dc7e2d94f69c82c.

src/apm_cli/deps/github_downloader.py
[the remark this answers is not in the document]
arika0093: address in fe7179f4cead0401feb356751a5dc0aa7f47c2ba.

src/apm_cli/models/dependency/reference.py
arika0093: addressed in commit cd8bed0c07c5e9592dde69047954eebe0969b6e3.

src/apm_cli/models/dependency/reference.py
arika0093: addressed in commit 3c0122cb48f9ee994e529ae5fa0bf7beb5f39edd.

src/apm_cli/deps/github_downloader.py
arika0093: addressed in fe7179f4cead0401feb356751a5dc0aa7f47c2ba.

src/apm_cli/models/dependency/reference.py
arika0093: addressed in commit 47b87ab1e3ffc2975dfba90c7e72737a85fd39e2.

arika0093:
@sergio-sisternes-epam 
Thanks for the review. I addressed the points you raised, so could you please take another look?
   
On #614: this PR now explicitly passes `artifactory_prefix=dep.registry_prefix` when reconstructing `DependencyReference`, which fixes a related missing-prefix propagation path. But I do not think it resolves the core #614 issue around `_parse_artifactory_base_url()` ignoring `PROXY_REGISTRY_URL`; that still appears to be unresolved.

sergio-sisternes-epam (approved):

sergio-sisternes-epam (approved):
All findings addressed — nice work, @arika0093! Each fix is clean and well-tested.

Thanks for the clarification on #614 — you're right that the `artifactory_prefix` propagation fix here is a partial improvement but the core `_parse_artifactory_base_url()` issue remains separate.

LGTM!

danielmeppiel (changes requested):
First — thanks for tackling this, @arika0093. The use case is real (internal mirrors without HTTPS exist in the wild), the dual opt-in design (per-dep `allow_insecure: true` + CLI flag) is the right shape, and the test coverage is thorough. Sergio's APPROVE isn't wrong on the implementation quality.

I'm requesting changes from a security/UX lens because APM is a supply-chain tool, and a few aspects of the current design create footguns that I don't think we can ship to the broader community. None of these are deal-breakers — they're scope tightening. If we land them, this becomes a feature I'd be proud to ship.

## Blockers

### 1. Drop `apm config set allow-insecure true` (the persistent global config)

This is the biggest single risk in the PR and, IMO, the only true blocker.

Per-invocation `--allow-insecure` is the right friction ceiling. A persistent global setting:

- Means a developer enables it once for one internal project and then **forever silently allows HTTP** for every subsequent `apm install` on that machine, including in unrelated repos.
- Turns shared CI runners into a network-MITM surface for any repo that lands `allow_insecure: true` in `apm.yml` — the runner doesn't need to opt in per-job, the maintainer who set the config months ago already did.
- Is invisible at install time. There's no per-install reminder. Users forget they set it.
- Converts a per-action security decision into ambient permission, which is the textbook supply-chain footgun.

The per-invocation flag is the correct model. It's friction, but the friction is the security feature: every HTTP install requires an active, conscious choice. Please drop the `apm config set allow-insecure` path entirely (and the corresponding `get_allow_insecure` / `set_allow_insecure` config functions).

### 2. Be loud at install time about which URLs are insecure

Today, the user only sees output about HTTP when `--allow-insecure` is **missing** (the error path). On the success path there's no equivalent loud diagnostic listing exactly which URLs are about to be fetched over an unauthenticated channel.

Compare to pip's behavior on insecure indexes: a clear warning per fetch. APM should do the same.

Concretely:
- Print one `[!]` line per HTTP dependency at install-prepare time, with the full URL: `[!] Fetching insecurely (no transport auth): http://internal.example.com/team/foo`
- Make this print whether or not `--allow-insecure` was passed (i.e. the warning is informational; the flag is the gate).
- This is what makes Alice notice when Bob's `apm.yml` grew an unexpected HTTP entry between `git pull`s.

### 3. Transitive HTTP dependencies need a separate gate

This is the most subtle one and the one I'd most appreciate your thinking on. Scenario:

- Bob adds a top-level dep `http://internal.bob.example/foo/bar` with `allow_insecure: true` to the project's `apm.yml`. Alice can see this in the diff and consents by passing `--allow-insecure`.
- But Bob's package itself has an `apm.yml` that pulls a **transitive** dep `http://attacker-controlled.example/lib/x` over HTTP, also with `allow_insecure: true`.
- When Alice runs `apm install --allow-insecure`, that transitive dep gets fetched too. Alice never saw it in the project's own `apm.yml`. Her consent at the root extends silently to a graph she didn't review.

The lockfile makes this auditable *after* it's pinned, but the first install (and any update) doesn't surface this distinction.

Suggested approach (open to alternatives):
- Track per-dep whether the HTTP+`allow_insecure` choice originated in the **root** `apm.yml` or in a **transitive** dep.
- For transitive HTTP, require an additional explicit acknowledgment: either a separate `--allow-insecure-transitive` flag, or an interactive prompt listing the transitive HTTP URLs and their introducing parent, with `--yes` to skip in CI.
- At minimum, the per-URL warning from #2 should annotate `(transitive, introduced by <parent-dep>)` so users can see it.

Without this, the dual-opt-in only protects the immediate dep — the transitive surface is a single root flag away from being silently expanded.

## Non-blocking but please consider

- **Threat-model docs**: the new HTTP section in `dependencies.md` documents the *mechanism* but not the *why*. A short paragraph explaining that HTTP has no transport auth (so an MITM can substitute the package contents — the SHA in the lockfile is itself MITM-able when fetched over HTTP) would help users make informed choices instead of copy-pasting the example.
- **Audit affordance**: there's no way to ask "list every insecure dep in my dependency graph" before running install. With the lockfile already preserving `is_insecure`, an `apm deps list --insecure` (or `apm install --dry-run` listing them) would be cheap and high-value. Could be a follow-up.
- **CHANGELOG entry**: this is a security-sensitive feature. The CHANGELOG entry should be explicit that this introduces an opt-in HTTP path with the threat-model implications, not just "feat: support http deps."
- **Hostname allowlist (future, not now)**: a future iteration could let `apm.yml` declare an `insecure_hosts: [...]` allowlist so that HTTP is only permitted from explicitly-named hosts. Not for this PR.

## What I'd merge

If you land #1 (drop global config) and #2 (loud per-URL output), I'd merge. #3 (transitive gating) is the one I feel strongest about from a supply-chain perspective; if you'd prefer to defer it, I'd accept that as a follow-up issue *only if* the per-URL warning from #2 explicitly annotates transitive HTTP deps so the information is at least visible.

If you'd rather not drop the global config, I'd respectfully decline this PR and recommend users set up HTTPS on their internal mirrors. APM is a supply-chain tool and persistent ambient "allow insecure" is incompatible with that role.

Thanks again for the thoughtful design and the thorough tests — this is salvageable and worth landing well.

arika0093:
Thanks for the thoughtful review. I think your concerns are reasonable, especially given that APM is a supply-chain tool.

I agree on the first two blockers:
- I will drop the persistent global `allow-insecure` config.
- I will make insecure fetches loud on the success path as well, with one warning per HTTP dependency showing the full URL.

On transitive HTTP dependencies: I agree this is the trickiest part. A root-level opt-in should not silently bless an unreviewed transitive graph.

After thinking more about it, I do not think a simple `--allow-insecure-transitive` flag is the right long-term shape, because it broadens trust based on graph position rather than which host is actually being trusted.

For this PR, my plan is:
- a root HTTP dependency allowed via `--allow-insecure` will also allow transitive HTTP dependencies from the same host
- transitive HTTP dependencies from a different host will still be blocked by default
- for non-interactive environments, I will add an explicit host-level opt-in for additional hosts (for example, a repeatable `--allow-insecure-host <host>` flag)
- install output will clearly warn for every HTTP dependency, including whether it is transitive and which parent introduced it

I think this is a better narrow fix for the security concern raised here, while also fitting the common internal-repository use case more naturally than a boolean transitive flag.

I will also update the docs and changelog accordingly. The audit affordance (`apm deps list --insecure`) also makes sense to me.

Thanks again for the careful review.

danielmeppiel (changes requested):
## APM Review Panel -- PR #700

[x] **Verdict: REQUEST CHANGES** -- one CRITICAL credential-leak finding plus a converging architectural/security issue around HTTP identity. The dual-opt-in design is sound and Sergio's earlier blocker is correctly fixed; the items below are what stand between this and merge.

@arika0093 -- this is a substantial, security-sensitive contribution and the design instincts are right. Five reviewers (python-architect, cli-logging-expert, devx-ux-expert, supply-chain-security-expert; growth-hacker side-channel) converged on a tight set of must-fixes. Thank you for the depth here -- 1338 lines + 30+ tests on a class of feature most projects ship sloppily.

### What's excellent
- **Dual opt-in (manifest `allow_insecure: true` + CLI `--allow-insecure`)** is the right mental model. Matches npm `overrides` + `--legacy-peer-deps`. Neither alone is sufficient -- correct.
- **Persistent global config dropped** per earlier review feedback. Right call; every install run should be a conscious act.
- **`--allow-insecure-host` for transitive deps** with same-host auto-propagation for direct deps is well-designed. Good FQDN validation prevents flag injection.
- **`parse_from_dict()` validates `allow_insecure` is boolean** -- prevents YAML type confusion.
- **Doc surface coverage** (cli-commands.md, dependencies.md, apm-guide skill resources, CHANGELOG) is exactly the discipline this codebase demands.

### Required changes (must land in this PR)

1. **[x] CRITICAL -- HTTP clone leaks credentials in plaintext.** `github_downloader.py` `_env_for(use_token=False)` strips `GIT_ASKPASS` and `GIT_CONFIG_NOSYSTEM` for HTTP attempts, **re-enabling system git credential helpers (macOS Keychain, Windows Credential Manager, `gh auth`)**. When git clones `http://internal.example.com/owner/repo`, credential helpers may resolve stored tokens for that host and embed them as plaintext `Authorization` headers. The existing test `test_insecure_http_dep_is_strict_by_default` asserts `"GIT_ASKPASS" not in env_used` -- that test is verifying the vulnerability.

   **Fix:** HTTP attempts must actively block credential resolution:
   ```python
   if attempt.scheme == "http":
       env = dict(self.git_env)
       env['GIT_ASKPASS'] = 'echo'
       env['GIT_TERMINAL_PROMPT'] = '0'
       env['GIT_CONFIG_NOSYSTEM'] = '1'
       # Suppress credential helpers explicitly
       env['GIT_CONFIG_COUNT'] = '1'
       env['GIT_CONFIG_KEY_0'] = 'credential.helper'
       env['GIT_CONFIG_VALUE_0'] = ''
       return env
   ```
   Update the test to assert credential-helper suppression for HTTP rather than its absence.

2. **[x] BLOCKER -- HTTP/HTTPS identity collision enables transport downgrade.** `get_unique_key()` and `get_identity()` are scheme-blind, and `drift.py` documents "Source/host/scheme changes -- not detected." That means `https://gitlab.com/team/rules` and `http://gitlab.com/team/rules` collide as the same dep. An attacker submitting a PR that flips the scheme + adds `allow_insecure: true` produces zero lockfile drift signal. Combined with finding #1, this is a clean transport-downgrade exploit chain.

   **Fix:** Either (a) `get_identity()` includes scheme when `is_insecure=True`, or (b) `drift.py` treats a `is_insecure` flip as drift requiring `--update`. Option (b) is the smaller change and preserves identity stability for non-HTTP deps.

3. **[x] BLOCKER -- Inverted import: `install/phases/resolve.py` imports from `commands/install.py`.** `resolve.py:297-300`:
   ```python
   from apm_cli.commands.install import (
       _check_insecure_dependencies, _collect_insecure_dependency_infos,
       _guard_transitive_insecure_dependencies, _warn_insecure_dependencies,
   )
   ```
   Domain code (`install/phases/`) must never import from CLI command code. Extract the ~200 lines of insecure-policy logic + `_InsecureDependencyInfo` dataclass into a new `src/apm_cli/install/insecure_policy.py`. Both `commands/install.py` and `install/phases/resolve.py` then import from the new module. Pure move, zero logic change.

4. **[x] BLOCKER -- `drift.py` propagates `is_insecure` but not `allow_insecure` during lockfile replay.** Lines +250-252: `build_download_ref()` copies `is_insecure=True` but drops `allow_insecure`. The replayed `DependencyReference` has `is_insecure=True, allow_insecure=False`, which `_check_insecure_dependencies()` rejects. Lockfile replays of HTTP deps are broken (or worse, bypass the check via a different code path).
   ```python
   if getattr(locked_dep, "is_insecure", False) is True:
       overrides["is_insecure"] = True
       overrides["allow_insecure"] = getattr(locked_dep, "allow_insecure", False)
   ```

5. **[x] BLOCKER -- `to_canonical()` no longer strips `http://`, breaking the canonical identity contract.** Architect + DevX both flagged this. The docstring update silently removes "no https://" and HTTP deps now produce `http://host/owner/repo` while HTTPS still strips. Every consumer that does `dep.to_canonical().split("/")` or assumes no `://` will silently diverge for HTTP deps. **Fix:** keep `to_canonical()` scheme-free. The transport-aware string already exists as `to_apm_yml_entry()` -- use that for serialization paths and any surface that needs the scheme.

6. **[!] HIGH -- Add `LockedDependency.to_dependency_ref()` factory.** Three sites reconstruct `DependencyReference` from `LockedDependency`: `_helpers.py:138`, `plugin_exporter.py:399`, `lockfile.py:365`. This PR already had to fix the missing `artifactory_prefix` (#614 root cause) AND add `is_insecure`/`allow_insecure` at all three. Per the "abstract when 3+ call sites" rule, add a factory method that maps every field once. This is the structural fix Sergio asked about -- the artifactory_prefix patch is a band-aid without it.

7. **[!] HIGH -- Unify the three competing error messages into one canonical recipe.** Today a user sees three different errors depending on which gate failed:
   - validation-time ("Pass `--allow-insecure`"),
   - install-time/manifest-missing ("set allow_insecure: true"),
   - install-time/flag-missing ("Pass `--allow-insecure`").

   None give the complete recipe. Collapse to one message that always names both requirements:
   ```
   [x] http://my-server.example.com/owner/repo -- HTTP dependency (no transport encryption)
       To install:
         1. Set allow_insecure: true on the dep in apm.yml
         2. Pass --allow-insecure to apm install
   ```
   And use the full URL (with scheme) in every error -- the user needs to see why it's insecure.

8. **[!] HIGH -- Update `docs/src/content/docs/enterprise/security.md`.** This PR introduces an entirely new transport surface and threat model; per the repo rule ("If a code change weakens or contradicts any guarantee in security.md, the doc must be updated in the same PR"), security.md needs an "HTTP (insecure) dependencies" section covering: what `allow_insecure` does/doesn't protect against, MITM impact on content-hash provenance, credential-helper isolation requirement, the transitive-host boundary model.

9. **[!] HIGH -- Logging routing: dead `else: _rich_*()` fallback branches.** `_warn_insecure_dependencies`, `_guard_transitive_insecure_dependencies`, `_check_insecure_dependencies` repeat `if logger: logger.X() else: _rich_X(...)` 7 times. Callers always have a logger; the else branches are dead code and bypass `DiagnosticCollector` if ever triggered. Make `logger` a required positional arg; drop every fallback branch.

### Polish (same PR if cheap; follow-up otherwise)

10. **[i] Library code calls `sys.exit(1)`.** `_check_insecure_dependencies` and `_guard_transitive_insecure_dependencies` should raise an exception; CLI layer catches and exits. Improves testability and reuse.

11. **[i] `_format_insecure_dependency_warning`** uses jargon ("no transport auth"). Reword:
    ```
    [!] Insecure HTTP fetch (unencrypted): http://...
    ```

12. **[i] Transitive-block error is a policy lecture.** Drop the middle sentence; lead with the command. Move the "why" to docs.

13. **[i] `apm deps list --insecure`** column label "HTTP" misleads (values are `direct` / `via <parent>`). Rename to `Origin`.

14. **[i] Unrelated cleanups** (config.py refactor, `_setup_git_environment` SSH cleanup) belong in separate commits for bisectability.

### Worth a follow-up issue (don't block this PR)

- `APM_ALLOW_INSECURE=1` env var counterpart for CI ergonomics (mirror `APM_ALLOW_PROTOCOL_FALLBACK=1` pattern). Not adding this is fine for v1; tracking issue acceptable.
- Drift-detection enhancement: warn on first-time HTTP install without lockfile SHA pin (defense-in-depth against the hash-circular-trust limitation).

### Merge call

This is solid first-PR work on a security-sensitive feature -- the dual-opt-in design and host scoping are correct. The blockers are concentrated in three areas: **credential leakage (#1), identity model (#2, #5), and architectural cleanup (#3, #4, #6)**. Items #1-#5 are the merge-blockers; #6-#9 should land in the same PR; the rest is polish or follow-up. Please address #1-#9 in this PR and re-request review.

### Growth angle
- Real adoption blocker unlocked: corporate Gitea/Gogs/Artifactory-on-HTTP and air-gapped enterprise users -- the segment that quietly worked around APM's "install from anywhere" claim with manual `git clone`. Worth a release-notes call-out.
- First-time NONE-association contributor shipping 1338 lines + 30+ tests on a security feature is the contributor signal worth amplifying once this lands. Suggested release-notes line: *"APM now supports private HTTP git hosts with explicit dual-gate opt-in (manifest + CLI). Insecure by default? Never. Insecure when you need it, with a clear audit trail."* Repostable hook: **"APM doesn't block your infra -- it makes insecure choices visible."**
- Once shipped, security.md becomes the authoritative reference for the threat model -- making the doc update in #8 above doubly load-bearing.

<sub>Reviewed via the apm-review-panel skill: python-architect, cli-logging-expert, devx-ux-expert, supply-chain-security-expert; arbitrated by apm-ceo with oss-growth-hacker side-channel.</sub>

arika0093:
Thanks for the detailed review. I addressed all requested changes from this review, rewrote the stack so the commits are easier to map back to the review items, and kept the split granularity intact.

Review item mapping:

1. HTTP git attempts now suppress credential helpers: b5ce27e
2. HTTP/HTTPS transport flips now show up as drift: c67cfde
3. Insecure policy logic was extracted out of CLI command code into `install/insecure_policy.py`: 9032b9e
4. Lockfile replay now preserves `allow_insecure` for HTTP dependencies: 1fca10b
5. `to_canonical()` is scheme-free again: caeb13d
6. `LockedDependency.to_dependency_ref()` now centralizes the mapping: 85be6d1
7. HTTP remediation errors were unified into one canonical recipe: f08a716
8. `docs/src/content/docs/enterprise/security.md` was updated for the HTTP threat model: 17e40a3
9. The insecure policy helpers now require a logger and no longer keep dead fallback branches: 32cb21f
10. Library-level insecure policy helpers now raise exceptions instead of calling `sys.exit(1)`: b067ab7
11. The insecure HTTP warning wording was tightened: 2dd90af
12. The transitive HTTP remediation error was shortened and now leads with the command: 783450f
13. `apm deps list --insecure` now uses `Origin` instead of `HTTP`: b8a5c5f

For review item 14:

- The unrelated `commands/config.py` refactor was split into its own commit for bisectability: 333e8f0
- The `_setup_git_environment()` / `GIT_SSH_COMMAND` SSH cleanup mentioned in that item was not introduced by this PR stack. It predates this branch, so there was no PR-local change to split out for that part.


At this point, all review items from this review have been addressed in the branch.

danielmeppiel:
## APM Review Panel — Round 2 Verdict (PR #700)

Re-review following @arika0093's commit-by-item response mapping. Panel: Python Architect, CLI Logging, DevX UX, Supply-Chain Security, OSS Growth. Five specialists, full convergence, no disagreements to arbitrate.

**Verdict: APPROVE with minor nits. Ready to merge after one small fix.**

All 14 items from the previous round are **substantively addressed**. Evidence verified commit-by-commit.

---

### Items verified resolved

| # | Blocker | Fix commit | Verdict |
|---|---------|------------|---------|
| 1 | Credential-helper leak on HTTP clone | `b5ce27eb` | Defence-in-depth: `GIT_ASKPASS=echo`, `GIT_TERMINAL_PROMPT=0`, `GIT_CONFIG_NOSYSTEM=1`, `GIT_CONFIG_COUNT/KEY/VALUE` triple. All 4 independent channels closed. Verified at `github_downloader.py:505-534` + tests `test_auth_scoping.py:277-326, 854-859`. |
| 2 | HTTP/HTTPS identity drift | `c67cfded` | XOR `is_insecure` comparison at `drift.py:100-105` catches both upgrade and downgrade. Tests cover both directions + no-drift case. |
| 3 | Inverted imports commands→install | `9032b9ef` | `install/insecure_policy.py` (209 LOC) with correct import direction. Zero residual duplicates in `commands/install.py`. Layering clean. |
| 4 | `allow_insecure` dropped on lockfile replay | `1fca10b9` | `drift.py:259-263` replays both `is_insecure` and `allow_insecure` from lockfile. Legacy lockfiles fail-closed (missing `allow_insecure` → False → re-prompt). |
| 5 | `to_canonical()` scheme-leak | `caeb13d9` | Scheme-free invariant restored. All consumers audited: display/identity paths scheme-free; only `to_apm_yml_entry()` keeps scheme where needed. |
| 6 | Missing `to_dependency_ref()` factory | `85be6d14` | Single source of truth at `lockfile.py:204-218`. All 3 call sites migrated. Zero hand-rolled `DependencyReference(` reconstructions outside the factory. |
| 7 | Unified error recipe | ✅ | Canonical `_format_insecure_dependency_requirements()` used across all three paths. **See nit N1 below.** |
| 8 | HTTP threat-model in `security.md` | `17e40a3b` | New section (L64-82) covers allow_insecure semantics, MITM exposure, content-hash provenance limitation, credential-helper isolation, transitive host model. Sales-enablement quality. |
| 9 | Dead `_rich_*` fallback branches | ✅ | `insecure_policy.py` has zero `_rich_*` imports/calls; all three policy functions call `logger.warning/error` bare (no `if logger:` guard, no `else:` fallback). |
| 10 | `sys.exit(1)` in library code | `b067ab7a` | `insecure_policy.py` raises `InsecureDependencyPolicyError(RuntimeError)`; CLI layer at `install.py:1329, 1419` owns the exit. Library-level is pure. |
| 11 | Jargon in warning copy | ✅ | Warning now reads `Insecure HTTP fetch (unencrypted)`. **See nit N1 — sibling jargon persists in error recipe.** |
| 12 | Transitive-block error | ✅ | `insecure_policy.py:172-178` leads with `Re-run with --allow-insecure-host HOST`. One-step fix, copy-pasteable. |
| 13 | `apm deps list --insecure` column | ✅ | Renamed to "Origin" in Rich and text-fallback tables. Values `direct` / `via <parent>`. Docs match. |
| 14 | Commit hygiene / bisectability | ✅ | 20 commits, one-per-review-item, descriptive `(review item N)` tags. `333e8f00` config.py refactor cleanly isolated. |

Architecture invariant (`install.py <= 1525 LOC`) preserved: **1496 LOC** after the merge-commit extraction of `mcp_warnings.py`. Full unit suite **4767/4767 passing**.

---

### Nits (should-fix; not merge-blockers)

**N1 — Double URL in validation-time path (CLI Logging, one-line fix)**
When `apm install <http-url>` fails the insecure check, the reason text already starts with `{url} --` but is passed to `logger.validation_fail(package, reason)` which prepends `{package} -- `. Result: `[x] http://server/owner/repo -- http://server/owner/repo -- HTTP dependency...`. Fix: call `logger.error(reason)` directly on the insecure-reject path at `install.py:292-297`, matching the other two failure paths. One-line change.

**N2 — Wording inconsistency between warning and error copy (CLI Logging)**
Warning says `(unencrypted)` (good, post-#11). Error recipe at `insecure_policy.py:67` still says `(no transport encryption)`. Unify to `(unencrypted)` for consistent register across both surfaces.

**N3 — Context-aware hint when only one gate is missing (DevX)**
`_check_insecure_dependencies` emits the full two-step recipe even when `allow_insecure: true` is already set on the dep in apm.yml. Users who already edited their manifest see "1. Set allow_insecure: true" redundantly. Conditionally emit only the missing step. Cosmetic; track as fast-follow.

**N4 — `_build_noninteractive_git_env` intent comment (Security)**
The `env.pop("GIT_ASKPASS", None)` at `github_downloader.py:514` unconditionally removes the APM default, then conditionally restores it only when `suppress_credential_helpers=True`. The intent (allow user keychains for HTTPS/SSH fallback, block them for HTTP) is correct but fragile. Add a short comment to prevent a future refactor from inverting the logic. No behaviour change.

**N5 — `--insecure` sample output missing from `cli-commands.md` (DevX)**
The `apm deps list` reference doc shows a sample output block for default mode but none for `--insecure`. Add a second block with the "Origin" column to help users preview the output without installing an HTTP dep.

---

### Tracked follow-ups (out of scope; open as issues)

- **F1** — Advisory SHA-pin for HTTP initial fetch (SRI-like `expected-sha:` in apm.yml) to break the first-fetch circular-trust loop documented in `security.md:81`. Hardening item.
- **F2** — `is_insecure` field not yet documented in `lockfile-spec.md` schema table. Doc-writer owns this.
- **F3** — `packages/apm-guide/.apm/skills/apm-usage/dependencies.md` doesn't mention `allow_insecure` or the transitive policy (commands.md does). Makes LLM-agent consumers slightly less informed on the dependencies.md page.
- **F4** — Pre-existing back-references in `install/presentation/dry_run.py:15` and `install/phases/finalize.py:24` violate the strict "no `install/` → `commands/`" layering. Inherited from #764, not this PR's regression. Worth a small follow-up extraction.
- **F5** — Host-only changes without a scheme flip still bypass drift detection (documented limitation at `drift.py:41-44`). Not a regression; worth an issue.

---

### Growth / positioning (side-channel)

PR #700 converts the README's "Install from anywhere" bullet from aspirational into defensible: corporate Gitea / Gogs / Artifactory / air-gapped Artifactory users can now install. The `enterprise/security.md` threat-model section is a sales-enablement asset that answers the enterprise-review question before it's asked. The `apm deps list --insecure` audit surface compounds that.

Contributor-velocity narrative: **@arika0093** (first-time external contributor, NONE-association) shipped a security-sensitive 1,300+ LOC feature with 204 new test functions, through two panel reviews, one atomic commit per review item. This is the contributor-funnel story APM should amplify in the release post — it sets the quality bar for future complex contributions.

Recommended release-notes hook:
> APM now installs from HTTP mirrors — and makes every insecure choice visible. Corporate Gitea, Gogs, and air-gapped Artifactory users can `apm install --allow-insecure` with per-host transitive controls, credential-helper suppression, and `apm deps list --insecure` auditability.

README "Install from anywhere" bullet should be updated in a follow-up to mention HTTP mirrors with explicit opt-in.

---

### Conflict resolution

Merged cleanly against `main` across two rounds (PR #809 landed mid-review):
- `c8bcc700` — initial conflicts + extract `mcp_warnings.py` to keep `install.py` under the 1525-LOC budget
- `35d64631` — second-round merge after #809 landed; took main's authoritative `mcp_warnings.py` (has FU4's `command=` parameter) and clean import ordering

Full suite green on HEAD. No functional changes from these merge commits — pure conflict resolution.

---

### Final call

**Merge-ready.** Single actionable ask before merge: fix N1 (one-line: swap `logger.validation_fail` → `logger.error` on the insecure-reject path). N2–N5 can ship as a tiny follow-up. F1–F5 → open as issues.

Outstanding work by @arika0093. Thank you for seeing this through two review rounds with the care each commit shows.

---

<sub>Panel: python-architect, cli-logging-expert, devx-ux-expert, supply-chain-security-expert, oss-growth-hacker. Synthesis via apm-review-panel skill. Conflicts resolved in `c8bcc700` + `35d64631`.</sub>

danielmeppiel:
Follow-up on [the panel verdict](https://github.com/microsoft/apm/pull/700#issuecomment-4290979672): I've pushed `2c7e81bf` addressing N1-N5, F2, and F3 directly on top of your branch so the PR stays a single-author squash-merge under your name. Co-authorship trailers preserve the full attribution.

**Nits addressed**

- **N1** -- Fixed the double-URL render. `install.py` now calls `logger.error(reason)` on the insecure-reject path instead of `logger.validation_fail` (which prepends the package name, producing `pkg -- url -- HTTP dependency...`).
- **N2** -- Unified the jargon register: error recipe now says `(unencrypted)` to match the warning copy.
- **N3** -- `_format_insecure_dependency_requirements` now takes `missing_dep_allow` / `missing_cli_flag` kwargs and only emits the steps the user actually needs. If `allow_insecure: true` is already on the dep entry, the error only asks for `--allow-insecure` (and vice versa). The CLI add-time path keeps both steps by default. Unit tests updated to assert the per-branch wording.
- **N4** -- Added an explicit docstring to `_build_noninteractive_git_env` documenting the intentional two-stage credential-helper policy (pop-then-conditionally-restore of `GIT_ASKPASS`), preventing a future refactor from inverting the logic.

**Follow-ups addressed (doc-writer agent)**

- **N5** -- Added an `apm deps list --insecure` sample output block to `docs/.../reference/cli-commands.md` showing the `Origin` column.
- **F2** -- Documented `is_insecure` and `allow_insecure` in `docs/.../reference/lockfile-spec.md` with replay semantics and legacy fail-closed behaviour.
- **F3** -- Added an "HTTP dependencies (opt-in)" section to `packages/apm-guide/.apm/skills/apm-usage/dependencies.md` so LLM agents consuming the skill know the dual opt-in model.

**Validation**

- Full unit suite: `4767/4767 passing` (one pre-existing deselect unrelated to this PR).
- `install.py` at 1497 LOC -- under the 1525-LOC architecture invariant.

Remaining follow-ups (`F1` advisory SHA-pin, `F4` `install/ -> commands/` inherited back-references, `F5` host-only drift) are out of scope and should be filed as separate issues. Ready for maintainer merge whenever CI gates clear.

Thank you for the iteration quality -- the one-commit-per-review-item hygiene made this second round fast and high-confidence.

danielmeppiel (approved):

Issue #245, opened 2026-03-11
[FEATURE] Enable Scoop bucket auto-update on release


sergio-sisternes-epam:
## Summary

Enable the Scoop bucket auto-update job in the release workflow so that `apm` is installable via `scoop install apm` on Windows.

## Current State

The release workflow already contains a fully scaffolded `update-scoop` job:

- **File:** `.github/workflows/build-release.yml`
- **Job definition:** starts at line 653
- **Disabled guard to remove:** line 659 — `if: false && github.ref_type == 'tag' && startsWith(github.ref_name, 'v')`
- **Target repo:** `microsoft/scoop-apm` (needs to be created)

The job dispatches a `repository_dispatch` event to the Scoop bucket repo with the version and download URL.

## Tasks

- [ ] Create the `microsoft/scoop-apm` Scoop bucket repository with the manifest template
- [ ] Remove the `false &&` guard from line 659 in `build-release.yml`
- [ ] Test the end-to-end flow: tag a release -> Scoop manifest updated -> `scoop install apm` works
- [ ] Update install instructions in README.md and docs/getting-started.md

## Related

- Parent tracking issue: #244
- Original feature request: #88
- PR #227 (Phase 1 implementation)

danielmeppiel:
[i] Closing -- the user-facing need is already community-served, and a first-party bucket is not on the current investment path.

Two reasons:

1. `scoop install apm` is already available via the community ScoopInstaller/Main entry (ScoopInstaller/Main#7762 -- thanks @sitiom). The core goal, Windows users installing `apm` through Scoop, is met today.
2. Standing up and auto-updating a dedicated first-party `microsoft/scoop-apm` bucket is release-automation work that is not on our current portability / security / governance investment path. It adds a second manifest to keep in sync on every release without a corresponding demand signal.

Clearing `status/accepted` + milestone and closing to unclutter the board. The scaffolded `update-scoop` job stays in `build-release.yml` if we revisit. If a first-party signed-manifest Scoop bucket becomes a priority (e.g. an enterprise Windows distribution requirement), reopen and we will scope it then. Thanks @sergio-sisternes-epam for the well-specified write-up.

PR #705, opened 2026-04-14
Fix Scoop bucket URL for apm installation

vandenberghev:
## Description

The current scoop url points to `https://github.com/microsoft/scoop-apm` and gives a 404.
Fixed the url.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Maintenance / refactor

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ pip install apm-cli
 
 ```powershell
 # Scoop
-scoop bucket add apm https://github.com/microsoft/scoop-apm
+scoop bucket add apm https://github.com/microsoft/apm
 scoop install apm
 # pip
 pip install apm-cli


sergio-sisternes-epam:
Thanks for catching this, @vandenberghev! You're absolutely right that the Scoop URL returns a 404 — good eye.

The reason is that the `microsoft/scoop-apm` bucket repository hasn't been created yet. Our CD pipeline already has a placeholder job for it ([`build-release.yml:753-789`](https://github.com/microsoft/apm/blob/main/.github/workflows/build-release.yml#L753)), but it's currently disabled while we finish setting it up. The Homebrew equivalent ([`microsoft/homebrew-apm`](https://github.com/microsoft/homebrew-apm)) is already live, and the Scoop bucket will follow the same pattern.

Unfortunately, pointing to `microsoft/apm` wouldn't work because the main repo isn't structured as a Scoop bucket (Scoop expects a `bucket/` directory with JSON app manifests).

This is tracked in #245 (assigned to @danielmeppiel) — once the bucket repo is created and the pipeline enabled, the existing README instructions will just work. Closing this for now, but thanks again for flagging it!

vandenberghev:
My bad, I missed that issue.
I tried adding the `scoop bucket` directly with `microsoft/apm` and it just worked.
As did the install. So I assumed it was okay.
`apm` seems to have be added to `ScoopInstaller/Main` so I guess it was all just a lucky coincidence ;)

Good to know it's in the works, thanks!

PR #709, opened 2026-04-14
fix(hooks): make apm install idempotent for hook entries

### LINKED ISSUE #708: Re-running `apm install` duplicates hook entries in settings.json
## Summary

Running `apm install` a second time (without wiping `.claude/`) appends a duplicate hook entry for every `_apm_source`-tagged hook already present in `.claude/settings.json`. The integrator extends the hook list unconditionally instead of upserting by `_apm_source`, so every re-install grows the file by one entry per hook per package.

## Reproducer

apm version: 0.8.11 (current main).

Starting from a clean `.claude/settings.json` produced by `apm install` (one Stop hook, `_apm_source: "agency"`, 16 lines), running `apm install` a second time with no other change produces:

```diff
         ],
         "_apm_source": "agency"
+      },
+      {
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": ".claude/hooks/agency/scripts/do-stop-guard.sh"
+          }
+        ],
+        "_apm_source": "agency"
       }
```

The second entry is byte-identical to the first. A third `apm install` adds a third, and so on.

In my case this is reproducible with the `srid/agency` APM package (has a single Stop hook), consumed from a kolu checkout. Any package declaring a hook should reproduce it.

## Root cause

`src/apm_cli/integration/hook_integrator.py:534` — inside `_integrate_merged_hooks` — unconditionally extends the per-event hook list:

```python
# Mark each entry with APM source for sync/cleanup
for entry in entries:
    if isinstance(entry, dict):
        entry["_apm_source"] = package_name

json_config["hooks"][event_name].extend(entries)
```

There is no check for existing entries tagged with the same `_apm_source`. The cleanup path elsewhere in the file (line 730, 775) already knows how to filter by `_apm_source` marker — the install path doesn't use the same guard.

## Expected behaviour

`apm install` should be idempotent: rerunning it with no changes should leave `settings.json` byte-identical. Specifically, before `.extend(entries)`, existing entries where `entry["_apm_source"] == package_name` for this event should be removed (or the new entries merged in place).

## Proposed fix

Before the `extend` at line 534, strip existing entries for the same `_apm_source` from `json_config["hooks"][event_name]`. Roughly:

```python
json_config["hooks"][event_name] = [
    e for e in json_config["hooks"][event_name]
    if not (isinstance(e, dict) and e.get("_apm_source") == package_name)
]
json_config["hooks"][event_name].extend(entries)
```

This matches the cleanup semantics already used by the uninstall/sync paths.

## Note on #561 / #562

[PR #562](https://github.com/microsoft/apm/pull/562) (which closed #561) has "hook dedup" in the title but the body clarifies it is code-level deduplication (collapsing the three copy-pasted `integrate_package_hooks_{claude,cursor,codex}` methods) with "zero behavioral change". It did not address this issue. Downstream users currently reference #561 in workaround comments believing it tracks this bug — worth clarifying in docs once fixed.

## Current workarounds

Consumers work around this today by wiping `.claude/settings.json` (or the entire `.claude/` tree) before `apm install`, which is destructive and requires a dedicated recipe. See [juspay/kolu's `agents/ai.just`](https://github.com/juspay/kolu/blob/master/agents/ai.just) for an example.

srid:
## Description

Re-running `apm install` appended a duplicate hook entry per package on every run because `_integrate_merged_hooks` extended the per-event list unconditionally.

Upsert by `_apm_source` before appending — the cleanup code path already treats that marker as the ownership key; the install path now matches.

Fixes #708

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [ ] All existing tests pass

### DIFF
--- a/src/apm_cli/integration/hook_integrator.py
+++ b/src/apm_cli/integration/hook_integrator.py
@@ -491,6 +491,11 @@ class HookIntegrator(BaseIntegrator):
         hooks_integrated = 0
         scripts_copied = 0
         target_paths: List[Path] = []
+        # Events whose prior-owned entries have already been cleared on
+        # this install run. Packages can contribute to the same event
+        # from multiple hook files — we must only strip once so earlier
+        # files' fresh entries aren't wiped by later iterations.
+        cleared_events: set = set()
 
         # Read existing JSON config
         json_path = target_dir / config.config_filename
@@ -535,10 +540,16 @@ class HookIntegrator(BaseIntegrator):
                 # package before appending fresh ones. Without this, every
                 # `apm install` re-run duplicates the package's hooks
                 # because `.extend()` is unconditional. See microsoft/apm#708.
-                json_config["hooks"][event_name] = [
-                    e for e in json_config["hooks"][event_name]
-                    if not (isinstance(e, dict) and e.get("_apm_source") == package_name)
-                ]
+                # Only strip once per event per install run — a package
+                # with multiple hook files targeting the same event
+                # contributes each file's entries in turn, and stripping
+                # on every iteration would erase earlier files' work.
+                if event_name not in cleared_events:
+                    json_config["hooks"][event_name] = [
+                        e for e in json_config["hooks"][event_name]
+                        if not (isinstance(e, dict) and e.get("_apm_source") == package_name)
+                    ]
+                    cleared_events.add(event_name)
                 json_config["hooks"][event_name].extend(entries)
 
             hooks_integrated += 1
--- a/src/apm_cli/integration/hook_integrator.py
+++ b/src/apm_cli/integration/hook_integrator.py
@@ -531,6 +531,14 @@ class HookIntegrator(BaseIntegrator):
                     if isinstance(entry, dict):
                         entry["_apm_source"] = package_name
 
+                # Idempotent upsert: drop any prior entries owned by this
+                # package before appending fresh ones. Without this, every
+                # `apm install` re-run duplicates the package's hooks
+                # because `.extend()` is unconditional. See microsoft/apm#708.
+                json_config["hooks"][event_name] = [
+                    e for e in json_config["hooks"][event_name]
+                    if not (isinstance(e, dict) and e.get("_apm_source") == package_name)
+                ]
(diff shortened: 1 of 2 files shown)


srid:
Thanks for the review @Copilot — all three addressed in [ac7a840](https://github.com/juspay/apm/commit/ac7a840):

**Multi-hook-file regression (`hook_integrator.py:542`)** — good catch. The filter ran inside the per-hook-file loop, so a package contributing to the same event from two JSON files would lose the first file's entries on the second iteration. Fixed by tracking cleared `event_name`s in a set and skipping subsequent filters.

**Multi-file test (`test_hook_integrator.py:597`)** — added `test_reinstall_preserves_multiple_hook_files_same_event`. It drops two hook files (`hooks-a.json`, `hooks-b.json`) both targeting `Stop` under the same package, asserts both entries survive install, and that the combined output is byte-identical across re-runs. Verified the test fails without the `cleared_events` gate (only the second file's entry survives) and passes with it.

**Defensive `parent.mkdir` (`test_hook_integrator.py:658`)** — added. The fixture already creates `.claude/`, so the test was passing, but the guard is cheap and makes the test robust to fixture changes.

sergio-sisternes-epam (approved):
Thorough review of the upsert logic and all 3 test scenarios. The `cleared_events` gate correctly handles multi-hook-file packages contributing to the same event — Copilot caught the original issue and @srid addressed it cleanly in ac7a840.

Code logic verified: strip-once-then-extend pattern is correct for single-file, multi-file, re-install, and legacy-duplicate scenarios. User-owned hooks (no `_apm_source`) are properly preserved by the filter.

Excellent contribution — clean fix with comprehensive regression tests.

sergio-sisternes-epam:
Thanks for the fix and the thorough test coverage, @srid — LGTM!

Could you rebase onto `main` when you get a chance? The branch is currently behind and CI needs to run on the updated base before we can merge.

srid:
@sergio-sisternes-epam Thanks! I've rebased the PR.

PR #715, opened 2026-04-15
fix: deploy root project .apm/ primitives without a sub-package stub

### LINKED ISSUE #714: apm install ignores root-level .apm/ primitives when project has external dependencies
## Description

Running `apm install` on a project that has both external APM dependencies in `apm.yml` and local rules in the root `.apm/` directory only deploys the external packages' rules -- the root-level `.apm/` is silently ignored.

## Reproduction

```
my-project/
├── apm.yml                              # has external dep
└── .apm/
    └── instructions/
        └── my-rules.instructions.md    # never deployed
```

```yaml
# apm.yml
name: my-project
version: 1.0.0
dependencies:
  apm:
    - some-org/some-package
```

```bash
apm install -t claude
# some-org/some-package rules deployed -> .claude/rules/  OK
# my-rules.instructions.md             -> NOT deployed
```

The same problem occurs even with no external deps at all -- `apm install` exits with "All dependencies are up to date" and deploys nothing.

## Workaround (from Discussion #595)

Move rules into a sub-directory and add a stub `apm.yml`:

```
my-project/
├── apm.yml           # depends on ./agent
└── agent/
    ├── apm.yml       # required stub
    └── .apm/instructions/my-rules.instructions.md
```

This is unnecessary boilerplate and the expected behavior should work without it.

## Root Cause

`_install_apm_dependencies()` has two early-return guards that check whether the resolved dependency list is empty. Both return before the target detection and integrator setup happen, so the project root's own `.apm/` directory is never processed. The integration pipeline only ever runs against paths inside `apm_modules/`.

## Expected Behavior

`apm install` should deploy primitives from the project root's own `.apm/` directory alongside any external dependencies, without requiring a sub-package stub.

Refs: Discussion #595

edenfunf:
## What

`apm install` now deploys primitives from the project root's own `.apm/` directory, alongside any declared external dependencies.

**Before:** running `apm install` on a project with root `.apm/instructions/` rules either exited silently ("All dependencies are up to date") or deployed only the external packages' rules -- the root-level `.apm/` was never touched.

**After:**
```
apm install
  [+] some-org/some-package
  |-- 1 rule(s) integrated -> .claude/rules/

  [+] <project root> (local)
  |-- 1 rule(s) integrated -> .claude/rules/
```

## Why

Users with both external dependencies and local rules in root `.apm/` were forced to create a dummy `./agent/apm.yml` stub and add it as a local-path dependency just to get their own rules deployed. This is the issue raised in Discussion #595 and tracked in #714.

The root cause is two early-return guards inside `_install_apm_dependencies()` that check whether the resolved dependency list is empty and return before target detection and integrator setup ever run. The integration pipeline only operates on paths inside `apm_modules/`, so the project root itself was never a candidate.

## How

Three changes to `src/apm_cli/commands/install.py`:

1. **`install()` entry point** -- detect root `.apm/` subdirectories and enter the APM install path even when `has_any_apm_deps` is False.
2. **First early return** -- move `project_root` computation before the guard and skip the return when root primitives exist.
3. **Second early return** -- same treatment for the `deps_to_install == []` guard.
4. **Root integration step** -- after the dependency loop, create a `PackageInfo(install_path=project_root)` and run it through the existing `_integrate_package_primitives` pipeline. No new integration logic; uses the same path all other packages take.

The existing `./agent/apm.yml` workaround continues to work without changes (verified by regression test).

## Test

Added `TestRootProjectPrimitives` in `tests/integration/test_local_install.py`:

- **no deps case** -- root `.apm/` + no external deps: rules deployed
- **mixed case** -- external dep + root `.apm/`: both rule sets deployed
- **regression** -- old `./agent/apm.yml` workaround still works
- **idempotency** -- running `apm install` twice produces the same result

All four new tests pass. Full test suite: 4309 passed, 0 new failures.

Fixes #714
Refs Discussion #595


src/apm_cli/commands/install.py
```diff
@@ -1400,12 +1412,24 @@ def _install_apm_dependencies(
...
+    # Check whether the project root itself has local .apm/ primitives (#714).
+    # Users should be able to keep root-level .apm/ rules alongside their apm.yml
+    # without creating a dummy sub-package stub.
+    _root_apm_dir = project_root / ".apm"
+    _LOCAL_PRIM_SUBDIRS = ('instructions', 'agents', 'chatmodes', 'context', 'memory', 'prompts')
+    _root_has_local_primitives = (
+        _root_apm_dir.exists() and _root_apm_dir.is_dir() and any(
+            (_root_apm_dir / d).is_dir() for d in _LOCAL_PRIM_SUBDIRS
+        )
+    ) or (project_root / "SKILL.md").exists()
```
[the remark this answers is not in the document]
danielmeppiel: We should be processing it in any case most likely, the integrators do check this. Probably just treating the .apm folder as a package itself.

edenfunf:
@danielmeppiel Thank you for taking the time to review this — your feedback was really helpful.

I've pushed a follow-up commit that addresses all three points raised in the review:

1. **Root primitive detection simplified** — removed the `_LOCAL_PRIM_SUBDIRS` allowlist from both `install()` and `_install_apm_dependencies()` as you suggested. The check now just verifies that `.apm/` exists and treats the project root as a package, letting each integrator decide whether it has content to deploy. This also eliminates the duplication Copilot flagged and avoids future drift if new primitive types are added.

2. **Error handling hardened** — the `except` block now passes the full traceback via `diagnostics.error(detail=...)`, and emits an explicit `logger.error()` when root integration is the only action (no external deps), so failures are visible rather than silently swallowed.

3. **Tests expanded** — added `test_root_apm_hooks_deployed` and `test_root_skill_md_detected` to cover the hooks and SKILL.md detection paths. All 6 tests in `TestRootProjectPrimitives` pass.

Let me know if there's anything else you'd like me to adjust. Thanks again!

src/apm_cli/commands/install.py
```diff
@@ -797,8 +797,17 @@ def install(ctx, packages, runtime, exclude, only, update, dry_run, force, verbo
...
             old_local_deployed = builtins.list(_existing_lock.local_deployed_files)
 
+        # Also enter the APM install path when the project root has local .apm/
+        # primitives, even if there are no external APM dependencies (#714).
+        from apm_cli.core.scope import get_deploy_root as _get_deploy_root
+        _cli_project_root = _get_deploy_root(scope)
+        _cli_root_apm = _cli_project_root / ".apm"
+        _cli_has_root_primitives = (
+            _cli_root_apm.exists() and _cli_root_apm.is_dir()
+        ) or (_cli_project_root / "SKILL.md").exists()
```
danielmeppiel: Why would we be processing it in case the project contains a Skill.md in root? I think that would be confusing UX

src/apm_cli/commands/install.py
```diff
@@ -1400,12 +1409,21 @@ def _install_apm_dependencies(
...
     project_root = get_deploy_root(scope)
     apm_dir = get_apm_dir(scope)
 
+    # Check whether the project root itself has local .apm/ primitives (#714).
+    # Users should be able to keep root-level .apm/ rules alongside their apm.yml
+    # without creating a dummy sub-package stub.
+    _root_apm_dir = project_root / ".apm"
+    _root_has_local_primitives = (
+        _root_apm_dir.exists() and _root_apm_dir.is_dir()
+    ) or (project_root / "SKILL.md").exists()
```
danielmeppiel: Second time the check appears in the file, seems error prone. Could extract to function.

danielmeppiel (changes requested):

src/apm_cli/commands/install.py
edenfunf: Good point — removed. The SKILL.md branch was carried over from the original implementation but there's no clear reason why its mere presence at the root should trigger the integration pipeline. Dropped it so the only trigger is the .apm/ directory.

src/apm_cli/commands/install.py
edenfunf: Done — extracted to a module-level helper _project_has_root_primitives(). Both call sites now just call the function, so there's a single place to update if the logic ever changes.

danielmeppiel (approved):

PR #724, opened 2026-04-15
fix: validate project name to reject path separators in apm init

### LINKED ISSUE #723: [BUG] apm init treats project name containing '/' as a filesystem path
**Describe the bug**
When running \`apm init\` with a project name that contains \`/\` (e.g. \`apm init 4/15\`), the argument is passed directly into \`Path(project_name)\`, causing \`/\` to be interpreted as a filesystem path separator. On Windows this produces a low-level OS error instead of a clear validation message.

**To Reproduce**
1. Run \`apm init 4/15\`
2. Observe: \`[x] Error initializing project: [WinError 3] The system cannot find the path specified: '4\15'\`

**Expected behavior**
\`apm init\` should reject project names containing path separators (\`/\` or \`\\`) with a clear error: \`Invalid project name '4/15': project names must not contain path separators\`.

**Environment:**
 - OS: Windows 11
 - Python Version: 3.13.4
 - APM Version: 0.8.11

**Additional context**
Discussed in https://github.com/microsoft/apm/discussions/718. Maintainer confirmed the preferred direction: strict project name semantics with input validation.

edenfunf:
## Description

`apm init` passed the `project_name` argument directly into `Path(project_name)`, so `/` and `\` were silently treated as filesystem path separators. On Windows, `apm init 4/15` produced a low-level OS error:

```
[x] Error initializing project: [WinError 3] The system cannot find the path specified: '4\15'
```

instead of a clear validation message.

Fixes #723
Closes discussion #718

## Changes

- **`_helpers.py`** — adds `_validate_project_name(name)`: returns `False` if the name contains `/` or `\`, making it clear these are invalid in a project name context.
- **`init.py`** — calls `_validate_project_name` before `Path(project_name)` is constructed and exits with a descriptive error. Also re-prompts in interactive mode when the user types an invalid name.
- **`tests/unit/test_init_command.py`** — adds `TestProjectNameValidation` (unit tests for the helper) and `TestInitProjectNameValidation` (integration tests verifying CLI exit code and output for `/` and `\` inputs).

## Type of change

- [x] Bug fix

## Testing

- [x] Tested locally
- [x] All existing tests pass


src/apm_cli/commands/init.py
```diff
@@ -201,7 +218,15 @@ def _interactive_project_setup(default_name, logger):
...
         logger.progress("Press ^C at any time to quit.")
 
-        name = click.prompt("Project name", default=default_name).strip()
+        while True:
+            name = click.prompt("Project name", default=default_name).strip()
+            if _validate_project_name(name):
+                break
+            click.echo(
+                f"{ERROR}Invalid project name '{name}': "
+                "project names must not contain path separators ('/' or '\\\\').{RESET}"
```
sergio-sisternes-epam: **Bug:** `{RESET}` is not interpolated here. The second string literal is a plain string (not an f-string), so `{RESET}` will print literally instead of applying the ANSI reset code. Either make both parts f-strings or consolidate into one:

```suggestion
                f"project names must not contain path separators ('/' or '\\\\').{RESET}"
```

src/apm_cli/commands/init.py
```diff
@@ -179,7 +188,15 @@ def _interactive_project_setup(default_name, logger):
...
 
-        name = Prompt.ask("Project name", default=default_name).strip()
+        while True:
+            name = Prompt.ask("Project name", default=default_name).strip()
+            if _validate_project_name(name):
+                break
+            console.print(
+                f"[error]Invalid project name '{name}': "
+                "project names must not contain path separators ('/' or '\\\\').[/error]"
+            )
```
sergio-sisternes-epam, line 198: **Test gap:** The new `while True` re-prompt loop has no test coverage. Please add a test that feeds an invalid name followed by a valid name in interactive mode (no `--yes`) and asserts the re-prompt behavior works correctly. Both the Rich and colorama paths should ideally be covered.

src/apm_cli/commands/_helpers.py
```diff
@@ -414,6 +414,17 @@ def _validate_plugin_name(name):
...
 
+def _validate_project_name(name):
+    """Validate that a project name does not contain path separators.
+
+    Project names are used directly as directory names and must not contain
+    '/' or '\\' to prevent unintended filesystem path traversal.
+
+    Returns True if valid, False otherwise.
+    """
+    return "/" not in name and "\\" not in name
```
sergio-sisternes-epam: **Architecture:** Per project conventions, user input used as filesystem paths must go through the centralized guards in `src/apm_cli/utils/path_security.py`. Consider using or complementing with `validate_path_segments(name, context="project name")`, which already handles backslash normalization and also rejects `..` traversal sequences.

As written, `apm init ..` passes `_validate_project_name` but would create a project in the parent directory. Note that `project_name == "."` is already special-cased at line 48, so no conflict there -- but `..` is not handled.

A simple approach: keep your separator check, and add a `..` guard (or delegate to `validate_path_segments`). Either way, please add a test case for `apm init ..`.

sergio-sisternes-epam (changes requested):
Good fix -- catching path separators before `Path()` prevents the confusing WinError on Windows. Three items need attention before this can merge (see inline comments).

edenfunf:
@sergio-sisternes-epam  Thanks for the detailed review! I've addressed all your feedback in the latest commit:

f-string bug: fixed the click.echo call so {RESET} is now part of an f-string and interpolates correctly
.. traversal: extended _validate_project_name to also reject .., and added a test for apm init ..
docstring: updated the wording to accurately reflect what the guard does
test coverage: added tests for the interactive re-prompt loop (both the invalid-name and .. cases)
test structure: refactored TestInitProjectNameValidation to use CliRunner.isolated_filesystem(), removing the manual os.chdir / try-finally bookkeeping
All 25 tests pass locally. Let me know if anything else needs attention!

PR #726, opened 2026-04-15
Fix duplicated skills/ directory nesting in plugin pack

### LINKED ISSUE #719: [BUG] apm pack --format plugin produces duplicated skills/ directory nesting
**Describe the bug**

When running `apm pack --format plugin`, the resulting build artifact contains a duplicated `skills/skills/` directory structure instead of the expected `skills/`. The `plugin.json` declares `"skills": ["skills/"]`, but the pack command places the skill folders one level too deep, resulting in `skills/skills/<skill-name>/SKILL.md` instead of `skills/<skill-name>/SKILL.md`.

**To Reproduce**

1. Run command `apm pack`
2. With parameters `--format plugin` inside a project containing an `apm.yml` with `dependencies.apm` pointing to a GitHub-hosted skill (e.g. `github/<org>/<repo>/skills/<skill-name>`)
3. Inspect the build output directory: `build/<name>-<version>/`
4. See error: the skills are placed under `build/<name>-<version>/skills/skills/<skill-name>/` instead of `build/<name>-<version>/skills/<skill-name>/`

**Expected behavior**

The packed plugin should produce the following structure:

```
build/halo-angular-1.0.0/
├── plugin.json
└── skills/
    └── javascript-typescript-jest/
        └── SKILL.md
```

**Actual behavior**

```
build/halo-angular-1.0.0/
├── plugin.json
└── skills/              ← declared in plugin.json as "skills": ["skills/"]
    └── skills/          ← duplicated nesting introduced by apm pack
        └── javascript-typescript-jest/
            └── SKILL.md
```

**Environment**

- OS: Linux (Debian GNU/Linux 13 trixie, dev container)
- APM Version: 0.8.11 (81082e2)
- VSCode Version: 1.115.0

**Logs**

`apm pack --format plugin` exits with code `0` (no error reported), making the issue silent. The malformed output is only discoverable by inspecting the build directory manually.

**`apm.yml`:**

```yaml
name: halo-angular
version: 1.0.0
description: APM project for halo-angular
author: Antonin Rouxel

dependencies:
  apm:
    - github/awesome-copilot/skills/javascript-typescript-jest
  mcp: []
scripts: {}
```

**Generated `plugin.json`:**

```json
{
  "name": "halo-angular",
  "version": "1.0.0",
  "description": "APM project for halo-angular",
  "author": {
    "name": "Antonin Rouxel"
  },
  "skills": [
    "skills/"
  ]
}
```

**Additional context**

The root cause appears to be that `apm pack` resolves the skill destination path as `skills/` (from `plugin.json`) and then copies the skill folders into a subdirectory also named `skills/`, resulting in double nesting. The pack command should either flatten the resolved path or not append an extra `skills/` segment when the destination folder is already named `skills/`.

Ai-chan-0411:
## Description

When running `apm pack --format plugin` with a dependency whose `virtual_path` starts with `skills/` (e.g. `github/awesome-copilot/skills/javascript-typescript-jest`), the bare skill collector (`_collect_bare_skill`) used the full `virtual_path` as the slug. Since the function already prepends `skills/` to the output path, this produced `skills/skills/javascript-typescript-jest/SKILL.md` instead of the expected `skills/javascript-typescript-jest/SKILL.md`.

I ran into this while testing a plugin that depends on `github/awesome-copilot/skills/javascript-typescript-jest` — the packed output had the skill files nested one level too deep, which meant the plugin host couldn't discover them.

The fix strips the leading `skills/` prefix from `virtual_path` before using it as the directory slug, preventing the double nesting. Added a regression test that verifies virtual paths with a `skills/` prefix produce the correct flat structure.

Fixes #719

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass (65/65 in test_plugin_exporter.py)

### DIFF
--- a/src/apm_cli/bundle/plugin_exporter.py
+++ b/src/apm_cli/bundle/plugin_exporter.py
@@ -131,6 +131,10 @@ def _collect_bare_skill(
     # Derive a slug: prefer virtual_path (e.g. "frontend-design"), else last
     # segment of repo_url (e.g. "my-skill" from "owner/my-skill")
     slug = (getattr(dep, "virtual_path", "") or "").strip("/")
+    # Strip leading "skills/" to avoid double nesting (skills/skills/…)
+    # when virtual_path already contains the skills/ prefix.
+    if slug.startswith("skills/"):
+        slug = slug[len("skills/"):]
     if not slug:
         slug = dep.repo_url.rsplit("/", 1)[-1] if dep.repo_url else "skill"
     for f in sorted(install_path.iterdir()):
--- a/src/apm_cli/bundle/plugin_exporter.py
+++ b/src/apm_cli/bundle/plugin_exporter.py
@@ -133,8 +133,9 @@ def _collect_bare_skill(
     slug = (getattr(dep, "virtual_path", "") or "").strip("/")
     # Strip leading "skills/" to avoid double nesting (skills/skills/…)
     # when virtual_path already contains the skills/ prefix.
+    # Also strip any extra leading slashes (e.g. "skills//foo" → "foo").
     if slug.startswith("skills/"):
-        slug = slug[len("skills/"):]
+        slug = slug[len("skills/"):].lstrip("/")
     if not slug:
         slug = dep.repo_url.rsplit("/", 1)[-1] if dep.repo_url else "skill"
     for f in sorted(install_path.iterdir()):
--- a/tests/unit/test_plugin_exporter.py
+++ b/tests/unit/test_plugin_exporter.py
@@ -308,7 +308,8 @@ class TestCollectBareSkill:
         )
         out: list = []
         _collect_bare_skill(tmp_path, dep, out)
-        assert any(r.startswith("skills/frontend-design/") for _, r in out)
+        rel_paths = [r for _, r in out]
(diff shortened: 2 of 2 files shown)


Ai-chan-0411:

Ai-chan-0411:
Thanks for the Copilot review — both points addressed in f75ad61:

**1. Leading slash after `skills/` strip** (`plugin_exporter.py`):
Added `.lstrip("/")` after the slice so `"skills//foo"` → `"foo"` instead of `"/foo"`. The slug computation now reads:
```python
slug = slug[len("skills/"):].lstrip("/")
```

**2. Loose assertion in `test_virtual_path_used_as_slug`** (`test_plugin_exporter.py`):
Changed from prefix check to exact equality:
```python
assert rel_paths == ["skills/frontend-design/SKILL.md"]
```
Applied the same tightening to `test_virtual_path_with_skills_prefix_no_double_nesting`, and added a new regression test `test_virtual_path_with_skills_double_slash_no_leading_slash` that exercises the double-slash edge case directly.

All 66 tests pass locally.

PR #734, opened 2026-04-16
fix: strip 'skills/' prefix in _collect_bare_skill to prevent double nesting

### LINKED ISSUE #719: [BUG] apm pack --format plugin produces duplicated skills/ directory nesting

Jah-yee:
## Summary

When a dependency has `virtual_path='skills/javascript-typescript-jest'`, the bare skill collector in `_collect_bare_skill` was producing output paths like `skills/skills/javascript-typescript-jest/SKILL.md` instead of `skills/javascript-typescript-jest/SKILL.md`.

This happens because the slug (derived from `virtual_path`) already contains the `skills/` prefix, and the code unconditionally prepends another `skills/` to the output path.

## Fix

Strip the `skills/` prefix from the slug before prepending `skills/` in the output path:

```python
# Strip "skills/" prefix to avoid double nesting like "skills/skills/..."
if slug.startswith("skills/"):
    slug = slug[len("skills/"):]
```

## Test added

`test_virtual_path_with_skills_prefix_no_double_nesting` in `tests/unit/test_plugin_exporter.py` covers this exact bug scenario.

Fixes #719.

### DIFF
--- a/src/apm_cli/bundle/plugin_exporter.py
+++ b/src/apm_cli/bundle/plugin_exporter.py
@@ -133,6 +133,9 @@ def _collect_bare_skill(
     slug = (getattr(dep, "virtual_path", "") or "").strip("/")
     if not slug:
         slug = dep.repo_url.rsplit("/", 1)[-1] if dep.repo_url else "skill"
+    # Strip "skills/" prefix to avoid double nesting like "skills/skills/..."
+    if slug.startswith("skills/"):
+        slug = slug[len("skills/"):]
     for f in sorted(install_path.iterdir()):
         if f.is_file() and not f.is_symlink() and f.name not in (
             "apm.yml", "apm.lock.yaml", "plugin.json",
--- a/tests/unit/test_plugin_exporter.py
+++ b/tests/unit/test_plugin_exporter.py
@@ -310,6 +310,25 @@ class TestCollectBareSkill:
         _collect_bare_skill(tmp_path, dep, out)
         assert any(r.startswith("skills/frontend-design/") for _, r in out)
 
+    def test_virtual_path_with_skills_prefix_no_double_nesting(self, tmp_path):
+        """virtual_path prefixed with 'skills/' does not produce 'skills/skills/...'."""
+        from apm_cli.bundle.plugin_exporter import _collect_bare_skill
+
+        (tmp_path / "SKILL.md").write_text("# Jest Typescript")
+        dep = LockedDependency(
+            repo_url="github/awesome-copilot",
+            resolved_commit="abc123",
+            depth=1,
+            virtual_path="skills/javascript-typescript-jest",
+            is_virtual=True,
+        )
+        out: list = []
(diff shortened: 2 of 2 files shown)


sergio-sisternes-epam:
Thank you for this contribution @Jah-yee! Your fix correctly identifies the root cause and the test is well-written.

We received two independent fixes for #719 -- yours and #738. After comparing both, we're going with #738 because it adds cross-platform path normalization (backslash handling) and includes an additional end-to-end integration test alongside the unit test.

Your effort is appreciated -- hope to see you on future issues!

PR #736, opened 2026-04-16
feat: add package variables (${var:...}) for deployed primitives

nathantoews:
Hello! APM team. This a feature I think would be super helpful to make apm a little more modular for abstractions of agents for installing. Please let me know what you think would be happy to make some changes.

## Summary

Adds support for declaring and resolving variables in APM packages. Package authors can define `${var:name}` placeholders in their primitive files, declare variables with defaults in `apm.yml`, and consumers can override values per-dependency. Variables are resolved at install time and substituted into all deployed text-based primitives.

## Motivation

Packages often need to be customizable without requiring consumers to fork or manually edit files after install. Variables provide a first-class mechanism for parameterization -- things like model names, API endpoints, project-specific paths, or behavioral toggles can be exposed as named variables with sensible defaults.

## How it works

1. **Declaration** -- Package authors add a `variables` section to `apm.yml` with names, defaults, and optional descriptions.
2. **Usage** -- Authors place `${var:name}` placeholders anywhere in their primitive text files (instructions, prompts, agents, skills, commands).
3. **Override** -- Consumers specify per-package variable values in their own `apm.yml` under `dependencies[].variables`.
4. **Resolution** -- At install time, APM resolves each variable (consumer override > package default) and substitutes placeholders in all deployed files.
5. **Lockfile** -- Resolved values are recorded in `apm-lock.yml` for reproducibility.

## Changes

- New `src/apm_cli/utils/variables.py` -- substitution engine (pattern matching, resolution, file walking)
- `models/apm_package.py` -- `variables` field on `APMPackage`
- `deps/lockfile.py` -- `resolved_variables` on `LockedDependency`
- `integration/base_integrator.py` -- `set_variables()` / `apply_variable_substitution()` on base class
- All integrators (agent, instruction, prompt, command, skill) -- apply substitution during deploy
- `commands/install.py` -- resolve variables per-package and thread through integration
- 66 new unit tests in `tests/unit/test_package_variables.py`
- Documentation: manifest schema reference, dependencies guide, package authoring skill, CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
 ## [Unreleased]
 
 ### Added
 
+- Package variables (`${var:...}`) for deployed primitives -- package authors declare variables in `apm.yml`, consumers override values, and placeholders are substituted at install time in `.agent.md`, `.instructions.md`, `SKILL.md`, and other text files
 - `apm install` now automatically discovers and deploys local `.apm/` primitives (skills, instructions, agents, prompts, hooks, commands) to target directories, with local content taking priority over dependencies on collision (#626, #644)
 
 ### Fixed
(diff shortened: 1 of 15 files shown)


danielmeppiel:
There is some support for input parameters in agent workflows. Adding variables is akin to creating a language and compiler/interpreter. That's something that I recognize as valuable but it's too early and we need discussion on design before. 

The place for variables should be the markdowns themselves and not the manifest I believe. 

Let's move this to a discussion

PR #746, opened 2026-04-17
fix(init): create start.prompt.md on init and fix next-steps guidance

### LINKED ISSUE #747: apm init does not create start.prompt.md, causing apm run start to fail immediately
### Description

After running `apm init`, the next advertised step is `apm run start`, but this fails immediately because `start.prompt.md` does not exist. The "Next Steps" panel also lists `apm install` and `apm compile` as if they are required, but neither is needed for a basic first run.

### Steps to Reproduce

```bash
mkdir my-project && cd my-project
apm init --yes
apm run start
```

**Result:** `apm run start` exits with an error — no prompt file found.

### Expected Behavior

`apm init` should create a starter `start.prompt.md` so the end-to-end flow (`init` → `run`) works without additional steps. The "Next Steps" guidance should reflect what a new user actually needs to do.

### Root Cause

`_create_minimal_apm_yml()` in `_helpers.py` writes only `apm.yml`. There is no code path that creates `start.prompt.md`, so the file is absent after init.

The "Next Steps" list includes:
- `apm install <owner>/<repo>` — not required to run a basic prompt
- `apm compile` — not required unless skills/instructions are installed

These mislead new users into running unnecessary commands before `apm run start`.

### Environment

- OS: Windows 11
- APM version: current `main`
- Python: 3.13

### Proposed Fix

- Create `start.prompt.md` in `_create_minimal_apm_yml()` when `plugin=False`, skipping if the file already exists (so re-running `apm init` on an existing project is safe)
- Trim next-steps to the three essential actions: install runtime → edit prompt → run
- List `start.prompt.md` in the "Created Files" table shown after init

edenfunf:
Closes #747

## What

**Before:** `apm init` created only `apm.yml`. Running `apm run start` immediately after would fail with _prompt file not found_. The "Next Steps" panel also listed `apm install` and `apm compile`, neither of which is required for a basic first run.

**After:** `apm init` creates `start.prompt.md` alongside `apm.yml`. Next Steps now shows the three essential steps: install a runtime, edit the prompt, run.

```diff
- next_steps = [
-     "Install a runtime:       apm runtime setup copilot",
-     "Add APM dependencies:    apm install <owner>/<repo>",
-     "Compile agent context:   apm compile",
-     "Run your first workflow: apm run start",
- ]
+ next_steps = [
+     "Install a runtime:       apm runtime setup copilot",
+     "Edit your prompt:        open start.prompt.md and write your instructions",
+     "Run your first workflow: apm run start",
+ ]
```

## Why

The root problem is that `apm run start` looks for `start.prompt.md` to exist, so `apm init` leaving it absent breaks the advertised end-to-end flow. `apm install` and `apm compile` are listed as if required but a fresh project with a single prompt file needs neither.

The overwrite-confirm path used `_lazy_confirm()` with a bare `except Exception` fallback; replacing it with `click.confirm()` (consistent with the rest of the fallback paths in the same file) avoids swallowing unexpected exceptions.

## How

- `_helpers._create_minimal_apm_yml()`: after writing `apm.yml`, write `start.prompt.md` with a template if it does not yet exist. Skipped for `--plugin` mode (matches the existing pattern where plugin projects don't receive a starter prompt). Pattern mirrors the `_create_plugin_json()` neighbour: create once, never overwrite.
- `init.py`: add `start.prompt.md` to the "Created Files" table (both rich and plain-text paths). Fix next-steps list. Replace `_lazy_confirm()` block with `click.confirm()`.

Reference: `src/apm_cli/commands/_helpers.py` — `_create_plugin_json()` uses the same "create if not exists" pattern for `plugin.json`.

## Test

```
pytest tests/unit/test_init_command.py    15/15  pass  (Python 3.13, Windows 11)
pytest tests/unit/                       3865/3865  pass  (no regression)
```

New test `test_init_starter_prompt_not_overwritten` verifies that re-running `apm init --yes` on an existing project does not clobber a customised `start.prompt.md`.

### DIFF
--- a/src/apm_cli/commands/_helpers.py
+++ b/src/apm_cli/commands/_helpers.py
@@ -21,8 +21,8 @@ from ..constants import (
     GITIGNORE_FILENAME,
 )
 from ..utils.console import _rich_echo, _rich_info, _rich_warning
-from ..version import get_build_sha, get_version
 from ..utils.version_checker import check_for_updates
+from ..version import get_build_sha, get_version
 
 # CRITICAL: Shadow Click commands at module level to prevent namespace collision
 # When Click commands like 'config set' are defined, calling set() can invoke the command
@@ -114,7 +114,10 @@ def _lazy_confirm():
 # Shared orphan-detection helpers
 # ------------------------------------------------------------------
 
-def _build_expected_install_paths(declared_deps, lockfile, apm_modules_dir: Path) -> set:
+
+def _build_expected_install_paths(
+    declared_deps, lockfile, apm_modules_dir: Path
+) -> set:
     """Build expected package paths under *apm_modules_dir*.
 
     Combines direct deps (from ``apm.yml``) with transitive deps
@@ -165,7 +168,9 @@ def _scan_installed_packages(apm_modules_dir: Path) -> list:
     for candidate in apm_modules_dir.rglob("*"):
         if not candidate.is_dir() or candidate.name.startswith("."):
             continue
-        if not ((candidate / APM_YML_FILENAME).exists() or (candidate / APM_DIR).exists()):
+        if not (
+            (candidate / APM_YML_FILENAME).exists() or (candidate / APM_DIR).exists()
+        ):
             continue
         rel_parts = candidate.relative_to(apm_modules_dir).parts
         if len(rel_parts) >= 2:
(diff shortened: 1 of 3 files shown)


edenfunf:
Closing this — I realized I had already submitted PR #606 covering the same fix back on 2026-04-07 and forgot about it. Sorry for the noise.

PR #748, opened 2026-04-17
fix(deps): align _count_package_files to use .apm/context/ (singular) dir

### LINKED ISSUE #704: [BUG] `_count_package_files` and `_get_detailed_context_counts` disagree on contexts directory name
## Description

`_count_package_files()` in `src/apm_cli/commands/deps/_utils.py` looks for `.apm/contexts/` (plural, line 103):

```python
context_dirs = ['instructions', 'chatmodes', 'contexts']
```

But `_get_detailed_context_counts()` in the same file maps the `'contexts'` key to `.apm/context/` (singular, line 138):

```python
context_directories = {
    'instructions': 'instructions',
    'chatmodes': 'chatmodes',
    'contexts': 'context'  # Note: directory is 'context', not 'contexts'
}
```

This means:
- A package with `.apm/context/` (singular) will have its files counted by `_get_detailed_context_counts` but **missed** by `_count_package_files`
- A package with `.apm/contexts/` (plural) will have its files counted by `_count_package_files` but **missed** by `_get_detailed_context_counts`

## Impact

`_count_package_files` feeds into `apm deps list` totals, while `_get_detailed_context_counts` feeds into `apm view` detail breakdown. The two commands could show inconsistent numbers for the same package.

## Discovered in

PR #682 -- the new tests correctly test each function's actual behavior, which made this inconsistency visible.

## Suggested fix

Align both functions on the same directory name. Check which name is canonical (likely `context/` singular based on the comment in `_get_detailed_context_counts`), and update the other function to match.

edenfunf:
## What

`_count_package_files()` in `src/apm_cli/commands/deps/_utils.py` scanned `.apm/contexts/` (plural) for context files, but the canonical directory is `.apm/context/` (singular). This caused `apm deps list` to always report **0 context files** for every package, while `apm view` showed the correct count — a silent, persistent inconsistency.

**Before:**
```python
context_dirs = ['instructions', 'chatmodes', 'contexts']
```

**After:**
```python
context_dirs = ["instructions", "chatmodes", "context"]
```

---

## Why

`.apm/context/` (singular) is the established convention across the codebase:

- `_get_detailed_context_counts()` in the same file already maps correctly, with an explicit comment:
  ```python
  'contexts': 'context'  # Note: directory is 'context', not 'contexts'
  ```
- Primitive discovery in `primitives/discovery.py` uses `"**/.apm/context/*.context.md"`.

`_count_package_files()` was the only function using the wrong plural form, so it always traversed a directory that doesn't exist and returned 0.

---

## How

One-character change: `'contexts'` → `'context'` in `_count_package_files()`, aligning it with `_get_detailed_context_counts()` at `src/apm_cli/commands/deps/_utils.py:149`.

No public CLI interface is affected — `apm --help` output is unchanged.

---

## Tests

```
pytest tests/unit/test_deps_utils.py    43/43   pass  (Python 3.13, Windows 11)
pytest tests/unit/                    3908/3908  pass  (no regression)
```

The test suite covers all public helpers in `_utils.py` across 8 test classes, adapted from the structure proposed in #682. One test was corrected in the process: `test_contexts_dir_counted` in #682 created `.apm/contexts/` (plural) and expected a non-zero count, which validated the buggy behavior. This PR changes it to use `.apm/context/` (singular), so it now correctly documents and pins the fix.

Fixes #704

### DIFF
--- a/src/apm_cli/commands/deps/_utils.py
+++ b/src/apm_cli/commands/deps/_utils.py
@@ -100,7 +100,7 @@ def _count_package_files(package_path: Path) -> tuple[int, int]:
         return 0, workflow_count
     
     context_count = 0
-    context_dirs = ['instructions', 'chatmodes', 'contexts']
+    context_dirs = ['instructions', 'chatmodes', 'context']
     
     for context_dir in context_dirs:
         context_path = apm_dir / context_dir
--- a/tests/unit/test_deps_utils.py
+++ b/tests/unit/test_deps_utils.py
@@ -3,9 +3,7 @@
 Covers the pure helpers that scan, count, and describe installed packages.
 """
 
-import tempfile
 from pathlib import Path
-from unittest.mock import patch
 
 import pytest
 
@@ -233,10 +231,10 @@ class TestCountPackageFiles:
         assert ctx == 1
 
     def test_contexts_dir_counted(self, tmp_path):
-        """Files in .apm/contexts/ are counted as context."""
+        """Files in .apm/context/ (singular) are counted as context."""
         apm = _make_apm_dir(tmp_path)
-        (apm / "contexts").mkdir()
-        (apm / "contexts" / "c.md").write_text("# c")
+        (apm / "context").mkdir()
+        (apm / "context" / "c.md").write_text("# c")
         ctx, _ = _count_package_files(tmp_path)
         assert ctx == 1
 


sergio-sisternes-epam (dismissed):
Good catch on the `contexts` -> `context` typo in `_count_package_files()`! The core fix is correct and the test coverage is solid.

However, this PR bundles several unrelated changes that should be split:

1. **The context dir fix** (the actual bug fix for #704) -- great, keep this and its test
2. **Formatting cleanup** across `_utils.py` and `_helpers.py` (single->double quotes, line wrapping) -- should be a separate PR
3. **`start.prompt.md` creation in `_create_minimal_apm_yml()`** and `init.py` changes -- these overlap with #649 which already addresses this exact feature. Merging both would create conflicts.

Could you please scope this down to just the `_count_package_files()` fix in `_utils.py` and its corresponding test? The formatting and init changes create unnecessary merge conflicts with other in-flight PRs.

The new `test_deps_utils.py` test file is welcome -- but please submit it without the init/helpers changes mixed in.

edenfunf:
@sergio-sisternes-epam Thanks for the thorough review!

Scoped the PR down to just the _count_package_files() fix and the test file — removed the formatting changes from _utils.py / _helpers.py and the init.py changes (noted the overlap with #649). Now it's a single commit touching only src/apm_cli/commands/deps/_utils.py (1-line change) and tests/unit/test_deps_utils.py.

edenfunf:
Rebased onto the latest main and resolved the conflict with #682's merged test file. Two changes applied on top of main's test_deps_utils.py:

1.Removed unused tempfile and patch imports as suggested.
2.Corrected test_contexts_dir_counted to use .apm/context/ (singular) — the version merged from #682 creates .apm/contexts/ (plural) and expects a non-zero count, which validates the bug this PR fixes. The test now reflects the correct behavior.

Regarding the docs finding — the references to contexts/ (plural) in docs/ and packages/apm-guide/ are pre-existing and outside this PR's scope. Happy to open a follow-up PR to update those if that's the preferred approach.

danielmeppiel (approved):

PR #755, opened 2026-04-18
feat: apply manifest namespace to local .apm skills

### LINKED ISSUE #739: [FEATURE] Support package namespaces
As an example, I have a project that looks something like this

- project, with skills
 - apm dependency 1
 - apm dependency 2
   - transitive dependency 1
   - transitive dependency 2

After apm install, the skills folder flattens everything, e.g.,
- project skills
- apm dependency 1
- apm dependency 2
- transitive dependecy 1
- transitive dependency 2

This is of course not APM's fault, it's the design of the .github folder. However, it leads to two distinct problems:
1. It quickly becomes very difficult to understand where various skills came from, particularly differentiating project skills from installed dep skills
2. Whilst developing a project, it's easy for the agent to update the installed skills because they are part of the "codebase" rather than a gitignored dep

Using namespaces could at least help solve problem #1. If the package manifest defined a namespace that was applied (presumably sometihng like namespace.skillname/SKILL.md) then it would be clear where each skill originated. This would also allow for easier copilot instructions indicating not to edit non-project skill prompts unless expressly asked to do so.

A secondary/related solution could be to have apm install add the individual dep skill folders to .gitignore along with apm_modules. This may not be desirable in all cases though, so would likely need to be optional.

Hand-adding a namespace to my skill names is how I'm handling multiple skill repositories in our org, but that doesn't work with third-party skills.

**Additional context**

shreejaykurhade:
## Description

This PR completes support for namespaced skill deployment by propagating the root `apm.yml` `namespace` into local `.apm/skills` integration.

The namespace behavior for skill deployment was already implemented; this change makes project-local skills follow the same rule as dependency-installed skills, so local sub-skills are deployed with names like `acme.design.linting` instead of unprefixed names.

This PR also adds regression coverage for the local `.apm` path and updates docs to reflect the expected namespaced skill folder behavior.

Fixes #739

## Type of change

- [] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [ ] All existing tests pass

Issue-scoped tests passed locally:
- `tests/unit/test_local_content_install.py`
- `tests/unit/integration/test_skill_integrator.py`
- `tests/integration/test_local_install.py -k "namespace or root_apm_skills_use_project_namespace"`

Note:
- The full test suite currently reports `24 failed, 4453 passed, 160 skipped, 9 deselected, 11 warnings`
- Those failures were not introduced as part of this issue-scoped change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `apm install` enforces org `apm-policy.yml` at install time (deps deny/allow/require, MCP deny/transport/trust-transitive, `compilation.target.allow`, `extends:` chains, `policy.fetch_failure` knob, `policy.hash` pin); `--no-policy` / `APM_POLICY_DISABLE=1` escape hatch; `--dry-run` previews verdicts; failed package installs roll back `apm.yml`. New `apm policy status` diagnostic (table / `--json`, exit-0 by default, `--check` for CI). `apm audit --ci` auto-discovers org policy. **Migration**: orgs publishing `enforcement: block` may see installs that previously succeeded now fail -- preview with `apm install --dry-run`. Closes #827, #829, #831, #834 (#832)
+- `apm experimental` command group - a feature-flag registry with `list` / `enable` / `disable` / `reset` subcommands. Opt in to new behaviour before it graduates to default. Ships with one built-in flag (`verbose-version`) and a contributor recipe for proposing new flags (#845)
 - `pr-review-panel` gh-aw workflow: runs the `apm-review-panel` skill on PRs labelled `panel-review` and posts a synthesized verdict (#824)
 
 ### Changed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,6 +144,44 @@ uv run isort .
 
 If your changes affect how users interact with the project, update the documentation accordingly.
 
+## Extending APM
+
+### How to add an experimental feature flag
+
+Use an experimental flag to de-risk rollout of a user-visible behavioural change that may need early adopter feedback. Do not add a flag for a bug fix, internal refactor, or any change that should simply ship as the default behaviour.
+
+Experimental flags MUST NOT gate security-critical behaviour (content scanning, path validation, lockfile integrity, token handling, MCP trust, collision detection). Flags are ergonomic/UX toggles only.
+
+When adding a new experimental flag:
+
+1. Register it in `src/apm_cli/core/experimental.py` in the `FLAGS` dict with a frozen `ExperimentalFlag(name=..., description=..., default=False, hint=...)`.
+2. Gate the code path with a function-scope import (avoids import cycles):
+   ```python
+   def my_function():
+       from apm_cli.core.experimental import is_enabled
+       if is_enabled("my_flag"):
+           ...
+   ```
+3. Add tests that cover both the enabled and disabled code paths.
+4. Update the experimental command reference page at `docs/src/content/docs/reference/experimental.md`.
+
+Naming rules:
+
+- Use `snake_case` in the registry and config.
+- Use `kebab-case` for display and other user-facing strings.
+- The CLI accepts both forms on input.
+
+Graduation and retirement:
+
+1. When a flag becomes the default, remove the gate and remove the matching `FLAGS` entry in the same PR.
+2. Add a `CHANGELOG.md` entry under `Changed` with a migration note if the previous default differed.
+
+Avoid these anti-patterns:
+
+- Do not gate security-critical behaviour behind an experimental flag.
+- Do not read `is_enabled()` at module import time.
+- Do not persist flag state anywhere other than `~/.apm/config.json` via `update_config`.
+
 ## License
 
 By contributing to this project, you agree that your contributions will be licensed under the project's [MIT License](LICENSE).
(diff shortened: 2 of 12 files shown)


sergio-sisternes-epam (commented):
Great first contribution, @shreejaykurhade! Namespace support is a feature our users are actively asking for (#739) and you've covered the core design well -- validation, normalization, test coverage, and thorough doc updates across 5 pages. Solid work.

I'm requesting three focused changes before we can merge. Two are small (~10 lines total) and one is a rebase. Everything else we'll track as follow-up issues that we own -- not your burden.

**Blocking (3 items):**

1. **Rebase onto current `main`** -- The install engine was refactored while your PR was in flight (#764). `_integrate_local_content` now lives in `src/apm_cli/install/services.py`. After rebasing, move your `project_namespace=` parameter to that function. We're happy to help if you get stuck -- just ping us here.

2. **Narrow the bare `except Exception`** in `sync_integration` -- see inline comment. Silent fallback to `None` can cause valid namespaced skills to be deleted as "orphans" when `apm.yml` is corrupted or unreadable. (~5 lines)

3. **Add `validate_path_segments()` defence-in-depth** in `build_deployed_skill_name()` -- see inline comment. The namespace regex is solid, but our security model requires all user-derived path components to pass through `path_security.py` as a second layer. (~3 lines)

**Nice-to-have (address if energy remains, else we'll follow up):**
- Add a one-sentence advisory in `manifest-schema.md` section 3.3: *"Adding `namespace` to an existing project renames deployed skill directories on the next `apm install`. Update any hardcoded skill paths accordingly."*
- Type-check that `namespace` from YAML is actually a string in `from_apm_yml()` (non-string values like lists or booleans will crash the validator)

**Follow-ups we'll file ourselves:**
- Store `namespace` in the lockfile (`LockedDependency`) so sync cleanup survives `apm_modules/` deletion
- Add namespace to `apm init` scaffold template
- Parse-time validation in manifest loading pipeline
- Adversarial test vectors for path traversal edge cases

**Naming convention**: The dot-separated format (`acme.design.skill-name`) is the right call -- it's the most natural namespace separator in computing and all major IDEs handle dotted directories without issue. Well chosen.

---
<sub>Review produced with input from the APM Expert Review Panel: Python Architect, CLI Logging Expert, DevX UX Expert, Supply Chain Security Expert, with CEO synthesis.</sub>

src/apm_cli/integration/skill_integrator.py
```diff
@@ -1003,19 +1070,26 @@ def sync_integration(self, apm_package, project_root: Path,
...
+            namespace = None
+            apm_yml_path = install_path / "apm.yml"
+            if apm_yml_path.exists():
+                try:
+                    from apm_cli.models.apm_package import APMPackage
+                    namespace = APMPackage.from_apm_yml(apm_yml_path).namespace
+                except Exception:
+                    namespace = None
+            skill_name, _ = build_deployed_skill_name(raw_name, namespace)
             installed_skill_names.add(skill_name)
```
sergio-sisternes-epam: **Narrow this exception handler + emit a diagnostic**

A bare `except Exception` here silently drops the namespace when `apm.yml` is corrupted, which changes the computed skill name and can cause `_clean_orphaned_skills()` to delete the correctly-namespaced directory as an "orphan".

Suggested fix:
```python
except (FileNotFoundError, ValueError) as exc:
    import logging
    logging.getLogger(__name__).debug(
        "Could not read namespace from %s: %s", apm_yml_path, exc
    )
    namespace = None
```

At minimum, don't swallow silently -- log at debug level so users can diagnose with `--verbose`.

src/apm_cli/commands/install.py
```diff
@@ -1293,6 +1294,7 @@ def _has_local_apm_content(project_root):
 def _integrate_local_content(
     project_root,
     *,
+    project_namespace=None,
```
sergio-sisternes-epam: **Heads-up: this function moved during a recent refactor**

`_integrate_local_content` was extracted to `src/apm_cli/install/services.py` in #764. After rebasing onto current `main`, apply this `project_namespace` parameter to the function there instead. The call site that passes `project_namespace=apm_package.namespace` likely moved too -- check `src/apm_cli/install/phases/local_content.py`.

Happy to help with the rebase if you need a hand!

src/apm_cli/integration/skill_integrator.py
```diff
@@ -127,6 +127,73 @@ def validate_skill_name(name: str) -> tuple[bool, str]:
...
+        normalized_namespace = namespace if ns_valid else normalize_skill_namespace(namespace)
+        if not ns_valid:
+            warnings.append(
+                f"Skill namespace '{namespace}' normalized to '{normalized_namespace}' ({ns_error})"
+            )
+        if normalized_namespace:
+            skill_name = f"{normalized_namespace}.{skill_name}"
+
+    return skill_name, " ".join(warnings) if warnings else None
+
```
sergio-sisternes-epam: **Add `validate_path_segments()` as defence-in-depth**

The namespace regex blocks traversal today, but APM's security model (`path_security.py`) requires all user-derived path components to pass through the centralised guards. One-liner addition before this return:

```python
from apm_cli.utils.path_security import validate_path_segments
validate_path_segments(skill_name, context="deployed skill name")
return skill_name, " ".join(warnings) if warnings else None
```

This ensures any future regression in the regex is caught by the path-level guard (16 other call-sites already follow this pattern).

sergio-sisternes-epam (commented):
Inline comments with code suggestions for the three blocking items above.

sergio-sisternes-epam (changes requested):
Marking as changes requested per the three blocking items in the review above (rebase, narrow exception handler, add path security guard). Looking forward to the updated PR!

PR #756, opened 2026-04-19
fix: rename manifest from apm.yml to apm.yaml for consistency

### LINKED ISSUE #745: Inconsistent file extensions between apm.yml and apm.lock.yaml
## Summary

APM currently uses:
- `apm.yml` for the manifest
- `apm.lock.yaml` for the lockfile

These two files are both YAML-based, but use different file extensions (`.yml` vs `.yaml`), which feels inconsistent.

## Problem

This inconsistency can be slightly confusing for users, especially since both files are:

- YAML 1.2 documents
- Core project configuration files
- Expected to be committed to version control

Most ecosystems tend to keep naming consistent across manifest and lockfile (e.g. `package.json` / `package-lock.json`, `pnpm-lock.yaml`, etc.).

## Proposal

Unify the file extensions for consistency. For example:

Option A:
- `apm.yml`
- `apm.lock.yml`

Option B:
- `apm.yaml`
- `apm.lock.yaml`

## Notes

- The lockfile was renamed from `apm.lock` to `apm.lock.yaml` for IDE syntax highlighting (per lockfile spec).
- However, the choice of `.yaml` over `.yml`, and the inconsistency with `apm.yml`, is not explained in the current documentation.

## Question

Was this inconsistency intentional?

If not, would you consider aligning the extensions for better consistency and developer experience?

Jah-yee:
Good day! 👋

Fixes #745

## Summary
Renames the manifest file extension from `.yml` to `.yaml` to match the lockfile (`apm.lock.yaml`), resolving the inconsistency reported in the issue.

## Changes
- `APM_YML_FILENAME` constant updated from `'apm.yml'` to `'apm.yaml'`
- All hardcoded `'apm.yml'` string references updated across source files  
- Policy model artifact names updated from `'apm.yml'` to `'apm.yaml'`

## Rationale
Per the issue, `.yaml` is the more widely-recognized standard extension (vs `.yml`), and the lockfile was already using `.yaml` for IDE syntax highlighting. Aligning both files to `.yaml` provides consistency.

This is a straightforward rename with no functional changes beyond the filename.

---
🏷️ labeled: good first issue


danielmeppiel:
Thank you for the effort here, @Jah-yee -- and apologies that the "accepted" label on #745 led you down a path I now think we shouldn't take. This is on me, not on you.

After looking at the full footprint of the rename: this PR changes 16 source files, but ~139 other files in the repo still reference `apm.yml` (tests, all docs under `docs/`, README, CHANGELOG, templates, integration scripts, and -- importantly -- the bundled `packages/apm-guide/.apm/skills/apm-usage/` skill that APM ships to teach users the manifest format). More critically, there's no dual-read fallback in the change, so every existing user's `apm.yml` would become silently invisible to APM on next upgrade, including in CI: any org running `apm policy check` against an existing repo would get a green build with zero policy applied, because "no manifest found" = no checks run.

Stepping back: manifest filenames are a stable contract that mature package managers treat as inviolable once real users exist (`package.json`, `Cargo.toml`, `Gemfile`, `pyproject.toml`, `composer.json` -- none have ever been renamed post-release). The "consistency" gain of matching `apm.lock.yaml`'s extension is purely aesthetic; both `.yml` and `.yaml` are equally valid YAML extensions (GitHub Actions uses `.yml`, Docker Compose uses `.yaml`, neither is "more correct"). The trade is aesthetic gain vs. ecosystem-wide breakage of every existing repo, tutorial, blog post, public package, and the conda-forge fixtures we just landed in #675. I underestimated that cost when I labeled #745 "accepted."

I'm going to close this PR and re-scope #745. If you're interested in continuing, the change that actually solves the underlying issue without the breakage is **dual-accept**:

- `apm_resolver` / discovery accepts either `apm.yaml` or `apm.yml`, prefers `.yaml` if both are present (warn on conflict)
- `apm init` writes `.yaml` going forward for new projects
- Existing user repos see zero breakage; new repos get the consistent extension

That's roughly scoped to `apm_resolver.py`, `init.py`, and `primitives/discovery.py` (~50 LOC + tests). I'll re-scope #745 in that direction. Either way, thank you for the contribution and for engaging with this repo -- the work you put in here is appreciated, even if the conclusion is "we shouldn't do the rename."

danielmeppiel:
Quick correction to my previous comment, @Jah-yee -- I offered dual-accept (`apm_resolver` reading either `apm.yaml` or `apm.yml`) as a fallback path. On reflection that contradicts the core argument and isn't the right call either.

The argument against the rename was that manifest filenames are stable contracts no mature package manager changes. That same argument rules out dual-accept: none of `package.json`, `Cargo.toml`, `Gemfile`, `pyproject.toml` accept fallback filenames either. Dual-accept would leave a permanent dual-path tax in discovery + tooling, fork the ecosystem (new repos on `.yaml`, old on `.yml` forever, the `apm-usage` skill teaching "either works"), and not deliver real consistency -- all to scratch an aesthetic itch.

I've closed [#745](https://github.com/microsoft/apm/issues/745) as won't-fix with the rationale on the record: `apm.yml` is and remains the canonical manifest filename. The `.yaml` extension on the lockfile is for IDE syntax-highlighting reasons specific to lockfiles and doesn't transfer to the manifest.

Thanks again for the work and the engagement -- and apologies for the muddled redirect. If you want a different "good first issue" to take on, happy to point you at one that's actually well-scoped.

Issue #666, opened 2026-04-10
[FEATURE] apm install should clean up stale deployed files when files are renamed within a package


Boubalou:
## Problem

When a file is renamed inside an APM package (e.g., a prompt file my-command.prompt.md → my-new-command.prompt.md), running apm install generates the new target files but leaves the old ones on disk as stale artifacts.

## Steps to reproduce

  1. Have a package with a prompt file my-command.prompt.md
  2. Run apm install — files are deployed to .claude/commands/my-command.md, .opencode/commands/my-command.md, .github/prompts/my-command.prompt.md
  3. Rename the source to my-new-command.prompt.md
  4. Run apm install again

**Expected:** Old files (my-command.md, my-command.prompt.md) are removed, new files (my-new-command.md, my-new-command.prompt.md) are created.

**Actual:** Both old and new files exist on disk. The old files must be manually deleted.

## Why this happens

The install loop writes the new deployed_files list to the lockfile, overwriting the previous one — but never computes the diff between old and new to remove orphaned files. The existing detect_orphans() in drift.py only operates at the package level (is the package still in apm.yml?), not at the file level within a package.

## Suggested approach

The infrastructure is already in place — apm.lock.yaml tracks deployed_files per dependency. During apm install, for each package:

  1. Read the old deployed_files from the existing lockfile
  2. Integrate the package (existing logic)
  3. Collect the new deployed_files
  4. Compute stale = old - new
  5. Delete stale files from disk

This would cover renames, deleted prompts, and any other case where a package's output file set changes between installs.

sergio-sisternes-epam:
Thanks for the clear write-up, @Boubalou -- the repro steps and suggested approach are spot-on.

You're right that the infrastructure is already there (`apm.lock.yaml` tracks `deployed_files` per dependency). The diff between old and new deployed files during install is a clean, well-scoped enhancement.

Accepting this. Good candidate for a community contribution -- the approach you outlined (old - new = stale, delete stale) is exactly what we'd want to see in a PR.

Boubalou:
Thanks for the comments @sergio-sisternes-epam, appreciated! I took this opportunity and went all in with the offer. The PR is up and running, waiting on your review. 

I'm not gonna lie, the code itself was executed through our Agentic flow here at Mirego, but rest assure that we use the same process internally for most of our work. The spec has been reviewed prior to the implementation and the execution, followed by myself along the way, was a great learning adventure of your code base and practices.

PR #762, opened 2026-04-19
fix(install): harden stale-file cleanup with per-file content-hash provenance (#666 follow-up)

danielmeppiel:
## Summary

Hardening follow-up to #750 (issue #666). Three commits, addressing every blocker and reasonable polish surfaced by an internal review team (security expert, Python architect, CLI logging UX expert) across three iteration rounds.

### 1. Per-file content-hash provenance for stale-file deletion

`#750` cleans up files that left a package between installs. This PR makes that cleanup **safe under user edits**:

- `LockedDependency.deployed_file_hashes: Dict[str, str]` records `sha256:<hex>` for every file APM deploys.
- On the next install, before deleting a "stale" file, APM compares the current on-disk hash to the recorded hash. **Mismatch -> file kept**, treated as a user edit, attributed in diagnostics.
- Missing recorded hash (legacy lockfiles) falls through to delete (back-compat).

### 2. Single safety chokepoint: `apm_cli/integration/cleanup.py`

New `remove_stale_deployed_files(...)` helper is the **only** path that deletes lockfile-tracked files on install. Three gates, in order:

1. `BaseIntegrator.validate_deploy_path()` -- rejects unmanaged prefixes.
2. Directory rejection -- defeats poisoned-lockfile rmtree.
3. Recorded-hash comparison -- defeats user-edit clobber.

The orphan-cleanup path (packages removed from `apm.yml` entirely) and the intra-package stale path both route through this helper. Pinned by `test_helper_signature_does_not_accept_logger` (SoC: helper pushes to `DiagnosticCollector`, never logs directly).

### 3. Logger surface for cleanup

Three new `CommandLogger` methods:
- `stale_cleanup(n, package=)` / `orphan_cleanup(n)` -- `[i]` symbol (informational, not progress; cleanup is not a success in itself).
- `cleanup_skipped_user_edit(path, package)` -- yellow inline warning per skipped file (verbose mode).
- `install_summary(stale_cleaned=)` -- final summary appends `(N stale files cleaned).` inside the sentence.

### 4. Orphan-cleanup correctness regression fix (commit 3)

The first refactor introduced a subtle bug caught by security re-review: the orphan loop derived its "still-declared" set from `package_deployed_files.keys()` (integration outcome). A transient integration error for a still-declared package would leave its key absent from that dict, and the loop would silently delete the package's files. **Fix:** derive from `intended_dep_keys` (manifest intent). Pinned by `test_orphan_loop_uses_manifest_intent_not_integration_outcome` (source-level guard that strips comments before checking).

## Review iteration

| Round | Security | Architecture | UX |
|-------|----------|--------------|-----|
| 1 | REQUEST-CHANGES (orphan path bypassed hash gate) | APPROVE-WITH-SUGGESTIONS (5 polish items) | APPROVE-WITH-SUGGESTIONS (1 wiring blocker + 6 polish) |
| 2 | REQUEST-CHANGES (orphan loop misclassification on integration failure) | APPROVE | APPROVE |
| 3 | APPROVE | -- (no code in scope) | -- (no code in scope) |

Deferred suggestions (tracked, non-blocking):
- Validate `sha256:<hex>` format on recorded hashes (fail-safe today).
- Plan a deprecation path for legacy lockfiles missing hashes.
- Apply safety gates to dry-run preview (preview-vs-actual divergence is UX, not security).

## Tests

- New: 14 unit tests in `tests/unit/integration/test_cleanup_helper.py` (gates, orphan path, SoC pin, regression guard).
- New: 9 unit tests in `tests/unit/test_command_logger.py` (cleanup methods + summary punctuation).
- New: 4 unit tests in `tests/test_lockfile.py` (`deployed_file_hashes` round-trip).
- Full suite: **3968 unit + 87 install/orphan integration tests pass.**

## Files changed

- New: `src/apm_cli/integration/cleanup.py`, `src/apm_cli/utils/content_hash.py`
- Modified: `src/apm_cli/commands/install.py`, `src/apm_cli/deps/lockfile.py`, `src/apm_cli/core/command_logger.py`
- Tests: `tests/unit/integration/test_cleanup_helper.py` (new), `tests/unit/test_command_logger.py`, `tests/test_lockfile.py`
- Docs: `docs/src/content/docs/reference/cli-commands.md` (stale-cleanup section + safety contract), `CHANGELOG.md` ([Unreleased] -> Fixed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Harden `apm install` stale-file cleanup (#666 follow-up): refuse to delete directory entries from the lockfile (closes a poisoned-lockfile vector); skip files the user has edited since APM deployed them via per-file SHA-256 provenance recorded in `apm.lock.yaml`; preview package-level orphan cleanup in `--dry-run`; reclassify recoverable cleanup failures as warnings (retried on next install) instead of errors; surface stale-file removals at default verbosity through a new `InstallLogger.stale_cleanup` / `orphan_cleanup` path; extract a shared `apm_cli.integration.cleanup.remove_stale_deployed_files` helper to dedupe the remote- and local-package cleanup blocks in `install.py`
+- Harden `apm install` stale-file cleanup to prevent unsafe lockfile deletions, preserve user-edited files via per-file SHA-256 provenance, and improve cleanup reporting during install and `--dry-run` (#666, #762)
 - Fix `apm marketplace add` silently failing for private repos by using credentials when probing `marketplace.json` (#701)
 - Harden marketplace plugin normalization to enforce that manifest-declared `agents`/`skills`/`commands`/`hooks` paths resolve inside the plugin root (#760)
 - Stop `test_auto_detect_through_proxy` from making real `api.github.com` calls by passing a mock `auth_resolver`, fixing flaky macOS CI rate-limit failures (#759)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Harden `apm install` stale-file cleanup (#666 follow-up): refuse to delete directory entries from the lockfile (closes a poisoned-lockfile vector); skip files the user has edited since APM deployed them via per-file SHA-256 provenance recorded in `apm.lock.yaml`; preview package-level orphan cleanup in `--dry-run`; reclassify recoverable cleanup failures as warnings (retried on next install) instead of errors; surface stale-file removals at default verbosity through a new `InstallLogger.stale_cleanup` / `orphan_cleanup` path; extract a shared `apm_cli.integration.cleanup.remove_stale_deployed_files` helper to dedupe the remote- and local-package cleanup blocks in `install.py`
 - Fix `apm marketplace add` silently failing for private repos by using credentials when probing `marketplace.json` (#701)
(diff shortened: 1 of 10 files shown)


danielmeppiel:
Thanks for the careful review. All six findings addressed in 4356c63. Summary:

**Critical (NameError)** -- `_hash_deployed` was an inner closure of `_install_apm_dependencies` but referenced from the sibling `_integrate_local_content`. Confirmed the call would `NameError` at runtime on the local `.apm/` persist path. Promoted to a module-level helper with explicit `(rel_paths, project_root)` signature; updated all three call sites; pinned by `test_hash_deployed_is_module_level_and_works`. Good catch.

**Security (fail-open on hash-read error)** -- Confirmed and fixed. The provenance gate now fails CLOSED: a `compute_file_hash` exception with a recorded hash present skips the deletion, records `skipped_user_edit`, and emits a "could not verify file content" warning. Pinned by `test_hash_read_failure_fails_closed`.

**Correctness (orphan target scope)** -- Confirmed against the pre-PR code, which explicitly merged `BaseIntegrator.KNOWN_TARGETS` for exactly this reason. Switched the orphan call site to `targets=None`, which `validate_deploy_path` treats as "all KNOWN_TARGETS" -- the cleaner equivalent. Added a comment explaining the trap.

**Docstring** -- Updated to reflect the current SoC (helper records to diagnostics, callers own logging). The mismatch was real -- the test `test_helper_signature_does_not_accept_logger` actively forbids the parameter the docstring still mentioned.

**CHANGELOG** -- Added `(#762)` and trimmed to a single concise sentence per the instructions file.

**`cleanup_skipped_user_edit` verbosity** -- The implementation (always emit) is intentional and correct: this is the visible safety guarantee surfacing -- the user *must* know when APM kept their edit, otherwise the per-file provenance check is invisible. The method's own docstring already says "Yellow inline at default verbosity". The PR description was misleading; that's a description-only correction, no code change.

Tests: 3970 unit + 87 install/orphan integration green.

PR #787, opened 2026-04-20
fix: detect port-like segment in SCP shorthand and suggest ssh:// form

### LINKED ISSUE #784: [BUG] SCP shorthand with numeric port silently misparsed as repo path
## Problem

When a user specifies an SCP-shorthand SSH dependency with a port number — e.g. from a Bitbucket Datacenter server — the parser silently treats the port as part of the repo path instead of raising an actionable error:

```yaml
dependencies:
  apm:
    - git@bitbucket.example.com:7999/project/repo.git
```

**Parsed result:**
| Field | Value |
|-------|-------|
| `host` | `bitbucket.example.com` |
| `port` | `None` |
| `repo_url` | `7999/project/repo` |

The user intended port 7999 with repo `project/repo`, but APM accepted `7999/project/repo` as a valid three-segment repo path. The subsequent `git clone` fails with a 404 from the remote — not an APM error message pointing at the SCP/port mismatch.

## To Reproduce

```yaml
# apm.yml
dependencies:
  apm:
    - git@bitbucket.example.com:7999/project/repo.git
```

1. Run `apm install --verbose`
2. Clone fails with `repository not found` from git
3. No APM-level warning about the SCP/port ambiguity

## Expected behavior

APM should detect the leading numeric segment and raise an actionable error:

```
Error: 'git@bitbucket.example.com:7999/project/repo.git' — '7999' looks like
a port number, but SCP shorthand (git@host:path) cannot carry a port.
Use: ssh://git@bitbucket.example.com:7999/project/repo.git
```

## Root cause

`_parse_ssh_url` (`reference.py:648`) uses the regex `^git@([^:]+):(.+)$` which captures everything after `:` as the repo path. Phase 3 validation (`[a-zA-Z0-9._-]+` per segment) accepts purely numeric segments like `7999`. No heuristic exists to detect the ambiguity.

## Suggested fix

After extracting `repo_url` in `_parse_ssh_url` (line 671), check if the first path segment is a valid port number:

```python
segments = repo_url.split("/")
first = segments[0]
if first.isdigit():
    port_candidate = int(first)
    if 1 <= port_candidate <= 65535:
        remaining = "/".join(segments[1:])
        raise ValueError(
            f"It looks like '{first}' in 'git@{host}:{repo_url}' is a port number, "
            f"but SCP shorthand cannot carry a port. "
            f"Use: ssh://git@{host}:{first}/{remaining}.git"
        )
```

**False positive analysis:** purely numeric org/owner names are not allowed on GitHub and extremely rare on other hosts (GitLab namespaces require at least one letter). If a user does have one, the error message directs them to the `ssh://` form which unambiguously separates port from path.

## Tests to add

- `git@host:7999/project/repo.git` → raises `ValueError` with suggestion
- `git@host:22/project/repo.git` → same (default port, still a port intention)
- `git@host:acme/repo.git` → no error (non-numeric first segment, existing behavior)
- `git@host:v2/repo.git` → no error (starts with letter)

## Context

Follow-up from PR #665 (port preservation). Identified in the review panel's devx-ux-expert finding.

Refs: #661, #665

edenfunf:
Fixes #784.

## Root Cause

`_parse_ssh_url` uses `^git@([^:]+):(.+)$` to parse SCP shorthand. Everything after `:` becomes the repo path. When a user writes:

```
git@bitbucket.example.com:7999/project/repo.git
```

The parser produces `repo_url = "7999/project/repo"` with `port = None`. Phase 3 validation accepts `"7999"` as a valid segment (`[a-zA-Z0-9._-]+`). The subsequent `git clone` fails with a 404 from the remote — no actionable error from APM.

The user intended port 7999 with repo `project/repo`, but SCP shorthand (`git@host:path`) uses `:` as the path separator and cannot carry a port. The correct form is `ssh://git@host:7999/project/repo.git`, which is handled by `_parse_ssh_protocol_url` via `urlparse`.

## Fix

After alias/ref extraction and `.git` stripping in `_parse_ssh_url`, check if the first path segment is a valid TCP port number (1–65535). If so, raise a `ValueError` with an actionable suggestion:

```
It looks like '7999' in 'git@bitbucket.example.com:7999/project/repo' is a port
number, but SCP-style URLs (git@host:path) cannot carry a port. Use the ssh://
URL form instead:
  ssh://git@bitbucket.example.com:7999/project/repo.git
```

### Before
```
git@bitbucket.example.com:7999/project/repo.git
→ repo_url="7999/project/repo", port=None → git clone → 404 ❌
```

### After
```
git@bitbucket.example.com:7999/project/repo.git
→ ValueError with actionable suggestion to use ssh:// form ✅
```

## Design decisions

- **ASCII-only digit check** (`re.fullmatch(r"[0-9]+", ...)` instead of `str.isdigit()`): Python's `isdigit()` returns `True` for Unicode superscript digits (²³) where `int()` raises `ValueError`. Using an ASCII-only regex avoids this edge case.
- **Placement before `validate_path_segments`**: a port-like first segment is a more specific and actionable error than a generic traversal/segment error. It takes priority.
- **Suggestion preserves `.git` suffix**: if the user wrote `.git`, the suggested URL includes it for copy-paste-ability. If omitted, the suggestion omits it too.
- **Port-only input** (`git@host:7999` with no remaining path): detected separately with a distinct message ("no repository path follows").
- **Out-of-range numbers** (0, 99999) fall through to existing validation — they are not valid ports and are unlikely to be port intentions.

## False positive analysis

Purely numeric org/owner names in the range 1–65535 would be a false positive. GitHub explicitly disallows numeric-only usernames; GitLab and Bitbucket require at least one letter. Self-hosted instances could theoretically allow it, but this is vanishingly rare. The error message clearly explains the issue and suggests the `ssh://` form, which unambiguously separates port from path.

## Changes

| File | Change |
|------|--------|
| `src/apm_cli/models/dependency/reference.py` | Add port-like segment detection in `_parse_ssh_url` (~28 lines) |
| `tests/unit/test_generic_git_urls.py` | Add `TestSCPPortDetection` class with 15 test cases |

## Tests

New `TestSCPPortDetection` class covering:

- Port 7999, 22, 1, 65535 (boundary values) — all raise with `ssh://` suggestion
- Leading zeros (`007999`) — detected and normalized
- Port with `#ref` and `@alias` — still detected after extraction
- Port-only / no remaining path — distinct error message
- Suggestion preserves or omits `.git` suffix based on user input
- Port 0 and 99999 (out of range) — NOT detected, fall through
- Normal org names (`acme`, `v2`) — parse normally, no change
- `ssh://` URL with port — continues to work (regression guard)

## Validation

```bash
uv run pytest tests/unit/test_generic_git_urls.py -x     # 128 passed
uv run pytest tests/unit tests/test_console.py -x         # 4120 passed
```

Refs: #661, #665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `apm install` no longer silently drops skills, agents, and commands when a Claude Code plugin also ships `hooks/*.json`. The package-type detection cascade now classifies plugin-shaped packages as `MARKETPLACE_PLUGIN` (which already maps hooks via the plugin synthesizer) before falling back to the hook-only classification, and emits a default-visibility `[!]` warning when a hook-only classification disagrees with the package's directory contents (#780)
 - Preserve custom git ports across protocols: non-default ports on `ssh://` and `https://` dependency URLs (e.g. Bitbucket Datacenter on SSH port 7999, self-hosted GitLab on HTTPS port 8443) are now captured as a first-class `port` field on `DependencyReference` and threaded through all clone URL builders. When the SSH clone fails, the HTTPS fallback reuses the same port instead of silently dropping it (#661, #731)
+- Detect port-like first path segment in SCP shorthand (`git@host:7999/path`) and raise an actionable error suggesting the `ssh://` URL form, instead of silently misparsing the port as part of the repository path (#784)
 
 ## [0.8.12] - 2026-04-19
 
--- a/src/apm_cli/models/dependency/reference.py
+++ b/src/apm_cli/models/dependency/reference.py
@@ -665,10 +665,38 @@ class DependencyReference:
         else:
             repo_part = ssh_repo_part
 
-        if repo_part.endswith(".git"):
+        had_git_suffix = repo_part.endswith(".git")
+        if had_git_suffix:
             repo_part = repo_part[:-4]
 
         repo_url = repo_part.strip()
+
+        # SCP syntax (git@host:path) uses ':' as the path separator, so it
+        # cannot carry a port.  Detect when the first segment is a valid TCP
+        # port number (1-65535) and raise an actionable error instead of
+        # silently misparsing the port as part of the repo path.
+        segments = repo_url.split("/", 1)
+        first_segment = segments[0]
+        if re.fullmatch(r"[0-9]+", first_segment):
+            port_candidate = int(first_segment)
+            if 1 <= port_candidate <= 65535:
+                if len(segments) > 1:
+                    remaining_path = segments[1]
+                    git_suffix = ".git" if had_git_suffix else ""
+                    suggested = f"ssh://git@{host}:{port_candidate}/{remaining_path}{git_suffix}"
+                    raise ValueError(
+                        f"It looks like '{first_segment}' in 'git@{host}:{repo_url}' "
+                        f"is a port number, but SCP-style URLs (git@host:path) cannot "
+                        f"carry a port. Use the ssh:// URL form instead:\n"
+                        f"  {suggested}"
+                    )
+                else:
+                    raise ValueError(
+                        f"It looks like '{first_segment}' in 'git@{host}:{first_segment}' "
+                        f"is a port number, but no repository path follows it. "
+                        f"SCP-style URLs (git@host:path) cannot carry a port. "
+                        f"Use the ssh:// URL form: ssh://git@{host}:{port_candidate}/<owner>/<repo>.git"
+                    )
 
         # Security: reject traversal sequences in SSH repo paths
         validate_path_segments(
--- a/src/apm_cli/models/dependency/reference.py
+++ b/src/apm_cli/models/dependency/reference.py
@@ -683,7 +683,9 @@ class DependencyReference:
                 if len(segments) > 1:
                     remaining_path = segments[1]
                     git_suffix = ".git" if had_git_suffix else ""
-                    suggested = f"ssh://git@{host}:{port_candidate}/{remaining_path}{git_suffix}"
+                    ref_suffix = f"#{reference}" if reference else ""
+                    alias_suffix = f"@{alias}" if alias else ""
+                    suggested = f"ssh://git@{host}:{port_candidate}/{remaining_path}{git_suffix}{ref_suffix}{alias_suffix}"
                     raise ValueError(
                         f"It looks like '{first_segment}' in 'git@{host}:{repo_url}' "
                         f"is a port number, but SCP-style URLs (git@host:path) cannot "
--- a/src/apm_cli/models/dependency/reference.py
+++ b/src/apm_cli/models/dependency/reference.py
@@ -680,8 +680,8 @@ class DependencyReference:
         if re.fullmatch(r"[0-9]+", first_segment):
             port_candidate = int(first_segment)
             if 1 <= port_candidate <= 65535:
-                if len(segments) > 1:
-                    remaining_path = segments[1]
+                remaining_path = segments[1] if len(segments) > 1 else ""
+                if remaining_path:
                     git_suffix = ".git" if had_git_suffix else ""
                     ref_suffix = f"#{reference}" if reference else ""
                     alias_suffix = f"@{alias}" if alias else ""
--- a/tests/unit/test_generic_git_urls.py
+++ b/tests/unit/test_generic_git_urls.py
@@ -781,3 +781,101 @@ class TestNestedGroupSupport:
         """HTTPS URLs can't embed virtual paths even with nested groups."""
         with pytest.raises(ValueError, match="virtual file extension"):
             DependencyReference.parse("https://gitlab.com/group/subgroup/file.prompt.md")
+
+
+class TestSCPPortDetection:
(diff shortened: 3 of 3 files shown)


danielmeppiel:
## Review panel verdict: **Approve** -- with two non-blocking suggestions

External contribution (@edenfunf) closing #784. Fix is small (29/+1 lines), surgical, well-tested (15 new test cases), and turns a silent 404 from a remote into an actionable APM-level error. Ran it past the full panel; key findings below.

---

### Python Architect: pass
- Change is local to `_parse_ssh_url`; no cross-module ripple. Good.
- Placement before `validate_path_segments` is correct -- a port-like first segment is a more specific error and should win.
- Tiny refactor (`had_git_suffix = ...`) is justified by the suggestion needing to mirror the user's `.git` suffix.
- The `re.fullmatch(r"[0-9]+", ...)` + range check is slightly belt-and-braces (`fullmatch` already excludes Unicode digits) but the comment explains the intent. No change requested.

### Supply Chain Security Expert: pass (with a note)
- **Strong identity-correctness win.** Pre-fix, `git@host:7999/project/repo.git` parsed to `repo_url="7999/project/repo"` -- a *different identity key* than the intended `project/repo`. That means lockfile entries keyed on the wrong identity and silent failure to dedup against the same dep added via `ssh://` form. The fix prevents that ambiguity at parse time. (`get_unique_key()` / lockfile keying confirmed elsewhere in the codebase.)
- **False positives**: a numeric-only first segment in the 1-65535 range. The PR's analysis is sound -- GitHub forbids numeric-only usernames, GitLab and Bitbucket require at least one letter, self-hosted edge cases are vanishingly rare and the error message points at a clean remediation. Accept.
- **Fails closed** (raises `ValueError`) -- correct.

### CLI Logging Expert: pass
- No `CommandLogger` paths touched. Error is a `ValueError` that flows through the existing install error handler -- the multi-line `\n  ssh://...` suggestion will render fine.

### DevX UX Expert: pass with one polish suggestion
- Error wording is unusually good for this class of problem: names the offending token, explains *why* SCP can't carry a port, gives a copy-paste-able `ssh://` form, and preserves `.git` based on user input. This is better UX than `git`'s native error and most package managers.
- **Suggestion (non-blocking)**: when the user wrote `git@host:7999/project/repo.git#main` or `...@my-alias`, the suggestion currently drops the `#ref` / `@alias` because they're stripped before the port check. The tests at lines 832 and 838 confirm the detection still fires -- great -- but the suggested `ssh://` URL won't include the user's ref/alias, so they have to re-add it manually. Worth mirroring them into the suggestion.
- Minor (out of scope): the quoted URL in the message uses the post-strip `repo_url` (e.g. `'git@host:7999/project/repo'`) rather than the user's literal input -- a bit confusing when the original had `.git` or `#ref`. Quoting the original `dependency_str` would be more faithful, but again, polish.

### APM CEO: approve, request CHANGELOG line
- Bitbucket Datacenter on non-22 ports is a real enterprise use case; fixing this unblocks adoption inside enterprises that bounced silently before.
- Not a breaking change -- previously users hit a 404 from `git`, now they hit an actionable APM error. Strictly better.
- **Ask**: add a one-line `Fixed` entry under `[Unreleased]` in `CHANGELOG.md` (`Closes #784, #787`). Per `.github/instructions/changelog.instructions.md` every merged PR that changes code needs a line. This is the only thing I'd consider near-blocking.

### OSS Growth Hacker: positive signal
- Story angle worth surfacing in the next release: "APM now plays nice with self-hosted Bitbucket / GitLab / Gitea on non-standard SSH ports." Small but it's the kind of "we listen to enterprise" detail that converts.
- External contributor merged quickly = social proof for the project's responsiveness. Good.

---

### Recommended path to merge

1. **Required**: add CHANGELOG entry under `[Unreleased]` -> `Fixed`.
2. **Optional polish** (can be follow-up if you want speed): preserve `#ref` and `@alias` in the suggested `ssh://` URL when present.

LGTM after #1. Thanks @edenfunf for the clean fix and thorough tests.

edenfunf:
Thanks @danielmeppiel for the thorough panel review. Both items addressed:

**1. CHANGELOG entry (required)** — added under `[Unreleased]` → `Fixed`:

> Detect port-like first path segment in SCP shorthand (`git@host:7999/path`) and raise an actionable error suggesting the `ssh://` URL form, instead of silently misparsing the port as part of the repository path (#784)

**2. Preserve `#ref` and `@alias` in suggestion (polish)** — the suggested `ssh://` URL now threads `reference` and `alias` back into the output. For example:

```
git@host:7999/project/repo.git#v1.0@my-alias
→ ssh://git@host:7999/project/repo.git#v1.0@my-alias
```

Added a dedicated test (`test_scp_port_with_ref_and_alias_preserves_both`) to cover the combined case.

**Also addressed Copilot inline findings:**

- `git@host:7999/` (trailing slash, empty remaining path) now correctly falls into the "no repository path follows" branch instead of building a suggestion ending with `/`. Added `test_scp_port_trailing_slash_no_path_raises` to cover this.
- Switched `test_suggestion_omits_git_suffix_when_absent` from `try/except + assert False` to `pytest.raises` for consistency with the rest of the test file.

All 4208 unit tests pass.

danielmeppiel (approved):
@edenfunf excellent PR - thank you for this!

Issue #780, opened 2026-04-20
Install silently drops skills/agents/commands when package also ships hooks/*.json


danielmeppiel:
## Summary

`apm install obra/superpowers` deploys only the 2 hook files and silently skips the package's 14 skills, 1 agent, and 3 commands. The package is misclassified as `HOOK_PACKAGE` even though it ships a complete Claude Code plugin layout (`.claude-plugin/plugin.json` + `agents/` + `skills/` + `commands/` + `hooks/`).

Reproduced on macOS, with the latest `main` (also confirmed against the released CLI).

## Reproducer

```bash
apm install obra/superpowers
```

Result reported by `apm install`:

```
Hooks   2   ~/.copilot/hooks/
```

Expected: 14 skills + 1 agent + 3 commands also deployed (the package is a fully-featured Claude Code plugin -- see https://github.com/obra/superpowers).

## Root cause

The detection cascade in `src/apm_cli/models/validation.py:138-174` (`detect_package_type`) is **order-sensitive** and the order is wrong when a package ships both hook files and a richer plugin layout:

```python
# Current order:
1. HYBRID            (apm.yml + SKILL.md)
2. APM_PACKAGE       (apm.yml)
3. CLAUDE_SKILL      (SKILL.md)
4. HOOK_PACKAGE      (hooks/*.json)               <-- fires here for superpowers
5. MARKETPLACE_PLUGIN (plugin.json OR agents/ OR skills/ OR commands/)
```

`obra/superpowers` matches **both** rule 4 (`hooks/hooks.json`) and rule 5 (`.claude-plugin/plugin.json`, `agents/`, `skills/`, `commands/`). Because rule 4 fires first, the package is classified `HOOK_PACKAGE`, which **does not** trigger `normalize_plugin_directory()` in `src/apm_cli/install/sources.py:178-180` -- the call that would synthesize `apm.yml` from the plugin layout and surface skills/agents/commands to the rest of the install pipeline. Result: only hooks ship.

Verified by direct invocation against a fresh clone:

```python
from pathlib import Path
from apm_cli.models.validation import detect_package_type
print(detect_package_type(Path("/tmp/superpowers-repro")))
# (<PackageType.HOOK_PACKAGE: 'hook_package'>, None)
```

The repo layout that triggers the bug:

```
.claude-plugin/
  plugin.json           <-- triggers rule 5
  marketplace.json
agents/code-reviewer.md <-- triggers rule 5
[...]

PR #789, opened 2026-04-20
fix(install): warn when --allow-protocol-fallback reuses a custom port across schemes (#786)

### LINKED ISSUE #786: [FEATURE] Cross-protocol fallback with custom port: port mismatch across SSH/HTTPS
## Problem

When `--allow-protocol-fallback` is enabled and a dependency has a custom port, `_build_repo_url` passes the same `dep_ref.port` to both `build_ssh_url` and `build_https_clone_url`. For servers where SSH and HTTPS run on different ports, the cross-protocol fallback produces an incorrect URL.

### Concrete scenario: Bitbucket Datacenter

Bitbucket DC defaults: SSH on port 7999, HTTPS on port 7990.

```yaml
- ssh://git@bitbucket.corp.com:7999/project/repo.git
```

| Attempt | URL produced | Result |
|---------|-------------|--------|
| SSH | `ssh://git@bitbucket.corp.com:7999/project/repo.git` | Correct |
| HTTPS fallback | `https://bitbucket.corp.com:7999/project/repo.git` | **Wrong** — HTTPS is on 7990, not 7999 |

### Other affected scenarios

| Setup | SSH attempt | HTTPS fallback | Result |
|-------|------------|----------------|--------|
| GitLab (SSH:2222, HTTPS:443) | `ssh://host:2222/...` ✅ | `https://host:2222/...` ❌ | Hits SSH daemon |
| Same-port server (e.g. Gitea:3000) | ✅ | ✅ | Works by accident |

## Current mitigations

**Strict mode (default since #778) prevents this from manifesting** for explicit `ssh://` and `https://` URLs — only the declared protocol is attempted. The bug only surfaces when users explicitly opt in via `--allow-protocol-fallback` or `APM_ALLOW_PROTOCOL_FALLBACK=1`.

The test `test_https_attempt_preserves_same_port_across_protocols` (`test_auth_scoping.py:423`) codifies "same port across protocols" as intended behavior — which is correct for same-port servers but wrong for the Bitbucket DC scenario.

## Root cause

`DependencyReference` has a single `port: Optional[int]` field shared across all protocols. There is no `ssh_port` / `https_port` distinction. `TransportAttempt` and `TransportSelector` have no port awareness — port handling lives entirely in `_build_repo_url` which blindly passes `dep_ref.port` to whichever builder it calls.

## Design space

The fundamental issue: **when crossing protocols, we cannot guess the correct port for the other protocol.**

### Option A: Warn + multi-port fallback cascade

When `dep_ref.port` is set and cross-protocol fallback triggers:
1. Try with same port (covers same-port servers)
2. If that fails, try with default port (covers servers where the other protocol uses default port)
3. Emit a warning: "Custom port {port} was specified for {scheme}://; the fallback protocol may use a different port."

**Pro:** no schema change. **Con:** doesn't cover BB DC (HTTPS:7990 is neither 7999 nor 443).

### Option B: `fallback` field in object form

```yaml
- git: ssh://git@bitbucket.corp.com:7999/project/repo.git
  fallback: https://bitbucket.corp.com:7990/project/repo.git
  ref: v1.0
```

**Pro:** fully explicit, covers any URL difference (not just port — also path differences for ADO, proxies, etc.). **Con:** schema expansion.

### Option C: Per-protocol port fields in object form

```yaml
- git: bitbucket.corp.com/project/repo
  ssh_port: 7999
  https_port: 7990
```

**Pro:** explicit. **Con:** only addresses port, not other URL differences; requires `host` field to be meaningful; less general than Option B.

### Option D: Status quo + documentation

Strict mode already prevents the problem for explicit URLs. Document that `--allow-protocol-fallback` with custom ports may produce incorrect fallback URLs, and recommend pinning the protocol that matches the port.

**Pro:** zero code change. **Con:** doesn't help users who genuinely need fallback across different ports.

## Files involved

- `src/apm_cli/deps/github_downloader.py` — `_build_repo_url` (line 664), `_clone_with_fallback`
- `src/apm_cli/deps/transport_selection.py` — `TransportSelector.select`, `TransportAttempt` (no port awareness)
- `src/apm_cli/utils/github_host.py` — `build_ssh_url`, `build_https_clone_url`
- `src/apm_cli/models/dependency/reference.py` — single `port` field

## Context

Follow-up from PR #665 review. The "same port across protocols" behavior was a deliberate choice in #665 to fail loudly rather than silently hitting a different service. Strict-by-default transport (#778) subsequently removed most of the exposure. This issue tracks the residual gap for the `--allow-protocol-fallback` path.

Refs: #661, #665, #731, #778

edenfunf:
## Description

When `--allow-protocol-fallback` (or `APM_ALLOW_PROTOCOL_FALLBACK=1`) is enabled and a dependency carries a custom port, `_build_repo_url` passes the same `dep_ref.port` to both `build_ssh_url` and `build_https_clone_url`. On servers that serve each protocol on a different port (e.g. Bitbucket Datacenter: SSH 7999, HTTPS 7990), the off-protocol URL is wrong.

Strict-by-default (#778) already prevents this for the default path; the bug only surfaces when the user opts into cross-protocol fallback.

Per the maintainer's Option D direction on #786 (warning + docs, no schema change):

- **`_clone_with_fallback`**: when the plan is permissive (fallback active) AND `dep_ref.port is not None` AND the planned attempts include both SSH and HTTPS, emit a default-visibility `[!]` warning before the first clone attempt, directing the user to pin the protocol with an explicit `ssh://` or `https://` URL. The warning routes through the existing `_rich_warning` idiom (same path as the current per-attempt fallback warning) and is deduped per `(host, repo, port)` on the `GitHubPackageDownloader` instance, so a single install run with multiple internal `_clone_with_fallback` calls (ref resolution + actual clone) still emits exactly one warning per dep.
- **`--allow-protocol-fallback` help text**: one-line caveat that fallback reuses the same port across schemes.
- **`docs/.../guides/dependencies.md` + `getting-started/authentication.md`**: short `:::caution` callouts under the transport-selection / fallback sections documenting the limitation and the pin-the-protocol workaround.
- **`CHANGELOG.md`**: one entry under `[Unreleased]` / `Changed`.

Explicit non-goals (per the review panel's rejected options): no `fallback:` schema field, no `ssh_port` / `https_port` fields, no schema migration, no multi-port cascade. The existing `test_https_attempt_preserves_same_port_across_protocols` invariant at `tests/unit/test_auth_scoping.py:423` is unchanged -- the warning is what makes that behaviour loud.

Fixes #786

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

### Manual verification

End-to-end against a custom-port dep on an unreachable host:

- `apm install --allow-protocol-fallback` with `ssh://git@host:7999/owner/repo.git` -> `[!] Custom port 7999 ...` fires exactly once, before any clone attempt (confirmed single emission even though ref-resolution and actual-clone phases each call `_clone_with_fallback`).
- Same dep without `--allow-protocol-fallback` (strict mode) -> no warning.
- Dep without a custom port + `--allow-protocol-fallback` -> no warning.
- `apm install --help` renders the updated `--allow-protocol-fallback` caveat.

### Automated tests

New file `tests/unit/test_protocol_fallback_warning.py` (7 tests):

- Warning fires for `ssh://` URL with port + `allow_fallback`, citing the port and the Bitbucket Datacenter example verbatim.
- Warning fires for `https://` URL with port + `allow_fallback`.
- Warning silent in strict mode even with a custom port (the `#778` invariant).
- Warning silent when `dep_ref.port is None`.
- Warning fires exactly once per `_clone_with_fallback` call (not per planned attempt).
- Warning deduped across repeated `_clone_with_fallback` calls for the same dep identity (regression guard for the ref-resolution + actual-clone multi-call path).
- Warning fires again for a distinct `(host, repo, port)` identity (dedup is per-dep, not global).

Full unit suite: 4170 passed, 0 new failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Strict-by-default transport selection: explicit `ssh://`/`https://` URLs no longer silently fall back to the other protocol; shorthand consults `git config url.<base>.insteadOf` and otherwise defaults to HTTPS. Set `APM_ALLOW_PROTOCOL_FALLBACK=1` (or pass `--allow-protocol-fallback`) to restore the legacy permissive chain; cross-protocol retries then emit a `[!]` warning. Closes #328 (#778)
 
+### Changed
+
+- `apm install --allow-protocol-fallback` now emits a `[!]` warning before the first clone attempt when a dependency has a custom port and both SSH and HTTPS attempts are planned, explaining that fallback reuses the same port across schemes and suggesting a pinned `ssh://` or `https://` URL for servers that use different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999, HTTPS 7990). Docs note the limitation in the transport-selection and authentication guides. Closes #786
+
 ### Added
 
 - `apm install --ssh` / `--https` flags and `APM_GIT_PROTOCOL=ssh|https` env to pick the initial transport for shorthand dependencies (#778)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `apm install --allow-protocol-fallback` now emits a `[!]` warning before the first clone attempt when a dependency has a custom port and both SSH and HTTPS attempts are planned, explaining that fallback reuses the same port across schemes and suggesting a pinned `ssh://` or `https://` URL for servers that use different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999, HTTPS 7990). Docs note the limitation in the transport-selection and authentication guides. Closes #786
+- `apm install --allow-protocol-fallback` now emits a one-shot `[!]` warning before the first clone attempt when a dependency carries a custom port and both SSH and HTTPS attempts are planned, naming the dependency and recommending either pinning the URL scheme (with fallback disabled) or dropping the flag to fail fast. Servers like Bitbucket Datacenter that serve SSH and HTTPS on different ports can otherwise hit a silent port mismatch. Closes #786 (#789)
 
 ### Added
 
--- a/docs/src/content/docs/getting-started/authentication.md
+++ b/docs/src/content/docs/getting-started/authentication.md
@@ -321,10 +321,12 @@ env var, and the `--allow-protocol-fallback` escape hatch, see
 
 :::caution[Custom ports and cross-protocol fallback]
 When `--allow-protocol-fallback` is in effect, APM reuses the
-dependency's port on both SSH and HTTPS attempts. On servers that use
-different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999,
+dependency URL's port on both SSH and HTTPS attempts. On servers that
+use different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999,
 HTTPS 7990), the off-protocol URL will be wrong. APM emits a `[!]`
-warning before the first clone attempt to flag this. Pin the protocol
-with an explicit `ssh://...` or `https://...` URL in `apm.yml` to
-bypass fallback entirely.
+warning before the first clone attempt to flag this. To avoid
+cross-protocol retries, leave `--allow-protocol-fallback` disabled
+(strict mode) and pin the dependency with an explicit `ssh://...` or
+`https://...` URL. If the flag is enabled, APM may still try the
+other protocol even when the URL uses an explicit scheme.
 :::
--- a/docs/src/content/docs/getting-started/authentication.md
+++ b/docs/src/content/docs/getting-started/authentication.md
@@ -318,3 +318,13 @@ URL scheme honored exactly; shorthand uses HTTPS unless
 selection matrix, the `--ssh` / `--https` flags, the `APM_GIT_PROTOCOL`
 env var, and the `--allow-protocol-fallback` escape hatch, see
 [Dependencies: Transport selection](../../guides/dependencies/#transport-selection-ssh-vs-https).
+
+:::caution[Custom ports and cross-protocol fallback]
+When `--allow-protocol-fallback` is in effect, APM reuses the
+dependency's port on both SSH and HTTPS attempts. On servers that use
+different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999,
+HTTPS 7990), the off-protocol URL will be wrong. APM emits a `[!]`
+warning before the first clone attempt to flag this. Pin the protocol
+with an explicit `ssh://...` or `https://...` URL in `apm.yml` to
+bypass fallback entirely.
+:::
--- a/docs/src/content/docs/guides/dependencies.md
+++ b/docs/src/content/docs/guides/dependencies.md
@@ -485,13 +485,16 @@ both protocols. Use this to unblock a pipeline while you fix the root
 cause -- not as a long-term setting.
 
 :::caution[Cross-protocol fallback reuses the same port]
-Fallback reuses `dep_ref.port` for both schemes. On servers that use
-different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999,
-HTTPS 7990), the off-protocol URL will be wrong. APM emits a `[!]`
-warning before the first clone attempt when a custom port is set and
-fallback is enabled. The fix is to pin the protocol with an explicit
-`ssh://...` or `https://...` URL in `apm.yml`, which bypasses the
-fallback chain entirely.
+Fallback reuses the dependency's custom port for both schemes. On
+servers that use different ports per protocol (e.g. Bitbucket
+Datacenter: SSH 7999, HTTPS 7990), the off-protocol URL will be
+wrong. APM emits a `[!]` warning before the first clone attempt when
+a custom port is set and fallback is enabled. To avoid cross-protocol
+retries entirely, leave `--allow-protocol-fallback` disabled (strict
+mode) and pin the dependency with an explicit `ssh://...` or
+`https://...` URL in `apm.yml`. If fallback is enabled, APM may still
+try the other protocol even when the URL uses an explicit scheme --
+pinning only hard-stops cross-protocol retries in strict mode.
 :::
 
 For SSH key selection (ssh-agent, `~/.ssh/config`) and HTTPS token
--- a/docs/src/content/docs/guides/dependencies.md
+++ b/docs/src/content/docs/guides/dependencies.md
@@ -484,6 +484,16 @@ When fallback runs, each cross-protocol retry emits a `[!]` warning naming
 both protocols. Use this to unblock a pipeline while you fix the root
 cause -- not as a long-term setting.
 
+:::caution[Cross-protocol fallback reuses the same port]
+Fallback reuses `dep_ref.port` for both schemes. On servers that use
+different ports per protocol (e.g. Bitbucket Datacenter: SSH 7999,
+HTTPS 7990), the off-protocol URL will be wrong. APM emits a `[!]`
+warning before the first clone attempt when a custom port is set and
+fallback is enabled. The fix is to pin the protocol with an explicit
+`ssh://...` or `https://...` URL in `apm.yml`, which bypasses the
+fallback chain entirely.
+:::
+
 For SSH key selection (ssh-agent, `~/.ssh/config`) and HTTPS token
 resolution, see
 [Authentication](../../getting-started/authentication/#choosing-transport-ssh-vs-https).
--- a/docs/src/content/docs/reference/cli-commands.md
+++ b/docs/src/content/docs/reference/cli-commands.md
@@ -98,7 +98,7 @@ apm install [PACKAGES...] [OPTIONS]
 - `-g, --global` - Install to user scope (`~/.apm/`) instead of the current project. Primitives deploy to `~/.copilot/`, `~/.claude/`, etc.
 - `--ssh` - Force SSH for shorthand (`owner/repo`) dependencies. Mutually exclusive with `--https`. Ignored for URLs with an explicit scheme.
 - `--https` - Force HTTPS for shorthand dependencies. Mutually exclusive with `--ssh`. Default unless `git config url.<base>.insteadOf` rewrites the candidate to SSH.
-- `--allow-protocol-fallback` - Restore the legacy permissive cross-protocol fallback chain (HTTPS-then-SSH or vice-versa). Strict-by-default otherwise. Each retry emits a `[!]` warning naming both protocols.
+- `--allow-protocol-fallback` - Restore the legacy permissive cross-protocol fallback chain (HTTPS-then-SSH or vice-versa). Strict-by-default otherwise. Each retry emits a `[!]` warning naming both protocols. When the dependency URL carries a custom port, APM also emits a one-shot `[!]` warning before the first clone attempt noting that the same port will be reused across schemes (wrong on servers like Bitbucket Datacenter that serve SSH and HTTPS on different ports) -- to avoid the mismatch, omit this flag and pin the dependency with an explicit `ssh://` or `https://` URL.
 
 **Transport env vars:**
 
--- a/src/apm_cli/commands/install.py
+++ b/src/apm_cli/commands/install.py
@@ -415,7 +415,7 @@ def _validate_and_add_packages_to_apm_yml(packages, dry_run=False, dev=False, lo
     "allow_protocol_fallback",
     is_flag=True,
     default=False,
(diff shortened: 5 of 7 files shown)


edenfunf:
@danielmeppiel -- ready for review when you have a moment.

This is the small PR you sketched on #786: the 30-50 line Option D change (warning + docs, no schema impact). It sits downstream of #778's strict-by-default plan without touching the `test_https_attempt_preserves_same_port_across_protocols` invariant, and is independent of #788 (which addresses #785's credential-resolution path).

One deviation worth flagging from your verdict text: the warning emission is deduped per `(host, repo, port)` on the `GitHubPackageDownloader` instance. Without dedup, the "once per dep" rule broke in practice because `_clone_with_fallback` is called multiple times for the same dep in a single install run (ref resolution clone + the actual dep clone). The unit test `test_warning_fires_once_not_per_attempt` only covered the single-call path; the dedup and its regression test (`test_warning_dedup_across_multiple_clone_calls_for_same_dep`) were added after end-to-end `apm install` surfaced the double emission.

Manual verification matrix is in the PR body. Happy to iterate on the wording or move the emission site to `TransportSelector.select` if you'd prefer -- I kept it in `_clone_with_fallback` so the logger stays out of the pure decision engine.

danielmeppiel:
## APM Review Panel — PR #789

[>] **Verdict: APPROVE-WITH-FIXES.** Land after a single polish pass; no re-queue specialist review needed.

Four specialists (python-architect, cli-logging-expert, auth-expert, devx-ux-expert) reviewed. Two approved cleanly; two flagged the same warning-text gap. Convergence is on a quick, mechanical polish pass.

### What's excellent

- **Architecture is right.** Dedup on `(dep_host, repo_url_base, dep_port)` at the per-install downloader is the correct scope, and the trigger placement before the attempt loop is clean.
- **Threat model honest.** Framed as UX/mechanical, not a security claim. No credential or URL-construction surface touched. Strict-mode silence verified.
- **Test coverage proportional.** 7 tests including strict-mode silence, dedup-across-calls, and re-fire-for-different-dep. Right shape.
- **Story is one CHANGELOG line** that reinforces "APM tells you when it's compromising on your behalf."

### Findings

1. **[x] BLOCKER — warning text is unactionable.** `"this dependency"` with N deps in flight = noise. Name the offender; `repo_url_base` and `dep_host` are already in scope. Target shape:

   ```
   [!] Custom port 7999 on bitbucket.example.com/project/repo: SSH unreachable, falling back to HTTPS.
       Pin the URL scheme, or drop --allow-protocol-fallback to fail fast.
       See: <docs URL to the :::caution block>
   ```

   This single edit absorbs the cli-logging blocker, the devx escape-hatch nit, and the devx doc-link nit in one line. Both `:::caution` blocks already exist as the link target.

2. **[x] BLOCKER — `cli-commands.md:101` stale.** Per the cli.instructions contract, a CLI behavior change without the reference doc updated in the same PR is incomplete. The `--allow-protocol-fallback` description must mention the new pre-attempt port warning, not just the per-retry one. Non-negotiable; same-PR fix.

3. **[i] Follow-up — dedup key case sensitivity.** GitHub/Bitbucket hosts are case-insensitive; key is case-sensitive. Worst case: a duplicate warning. Add a `# NOTE: case-sensitive` comment now, file an issue, move on.

4. **[i] Follow-up — `_fallback_port_warned` thread-safety** (auth-expert). Plain set on shared downloader. Benign race, worst case duplicate warning. CLI is single-threaded today; revisit if/when we parallelize installs.

5. **[i] Nice-to-have — CHANGELOG sentence is ~80 words.** Split into "Added" + one-line behavior note. Author discretion.

### Merge call

Author addresses (1) + (2) inline — both are mechanical, ~15 minutes. No re-review needed; maintainer can self-confirm against the warning shape above and the `cli-commands.md` diff. (3)–(5) become follow-up issues, linked from the merge commit. Then merge. [+]

---

### Growth angle

- **Enterprise trajectory signal.** PR #788 (port-threading, from external contributor @edenfunf) + PR #789 (one-time SSH→HTTPS fallback warning) is the clearest "enterprise-ready" beat APM has shipped: a self-hosted user hit a real Bitbucket Data Center / GHES edge case, contributed the fix, and a maintainer hardened the UX around it the same week. Marketable narrative — "APM's on-prem story is now community-hardened, not just maintainer-claimed." Worth name-checking @edenfunf in the release post.
- **Transport selection arc is now blog-shaped.** #665 → #784 → #787 → #788 → #789 has crossed the threshold from "internal refactor" to "coherent capability." Recommend a dedicated docs page (`guides/enterprise-transports.md`) plus a short blog post: "How APM resolves git transport for self-hosted GHES, GHE, and Bitbucket Data Center." Reusable proof for every future enterprise conversation.
- **Competitive wedge.** npm, cargo, and uv all assume a single registry over HTTPS; none thread custom SSH ports or degrade gracefully across protocols for self-hosted git hosts. Defensible one-liner: "the only AI-native package manager that speaks your enterprise git topology." Slots cleanly under the existing "package manager for AI-native development" frame.

<sub>Reviewed via the `apm-review-panel` skill: python-architect, cli-logging-expert, auth-expert, devx-ux-expert; synthesized by apm-ceo with oss-growth-hacker side-channel.</sub>

danielmeppiel (changes requested):
[x] BLOCKER — warning text is unactionable. "this dependency" with N deps in flight = noise. Name the offender; repo_url_base and dep_host are already in scope. Target shape:

```
[!] Custom port 7999 on bitbucket.example.com/project/repo: SSH unreachable, falling back to HTTPS.
    Pin the URL scheme, or drop --allow-protocol-fallback to fail fast.
    See: <docs URL to the :::caution block>
```
This single edit absorbs the cli-logging blocker, the devx escape-hatch nit, and the devx doc-link nit in one line. Both :::caution blocks already exist as the link target.

[x] BLOCKER — cli-commands.md:101 stale. Per the cli.instructions contract, a CLI behavior change without the reference doc updated in the same PR is incomplete. The --allow-protocol-fallback description must mention the new pre-attempt port warning, not just the per-retry one. Non-negotiable; same-PR fix.

edenfunf:
@danielmeppiel Polish pass pushed (9a0b022). All comments addressed; CI 5/5 green on the new head.

## Maintainer blockers

- [x] **Warning text (BLOCKER 1)**. New 3-line shape names the offender, both planned schemes, both remediations, and links to the docs anchor. Pre-emptive placement kept (per your "trigger placement before the attempt loop is clean" call), so the tense shifted from "SSH unreachable, falling back to HTTPS" (post-hoc) to "if HTTPS fails, APM will retry over SSH" (accurate at emission time). Live output:

    ```
    [!] Custom port 7999 on nonexistent.invalid/project/repo: if HTTPS fails, APM will retry over SSH on the same port.
        Pin the URL scheme, or drop --allow-protocol-fallback to fail fast.
        See: https://microsoft.github.io/apm/guides/dependencies/#restoring-the-legacy-permissive-chain
    ```

    The initial/fallback schemes are derived from `plan.attempts` at emission time, so whichever protocol the selector actually plans first (HTTPS in the test fixtures' no-token case, SSH when the user has a PAT and an insteadOf rewrite, etc.) is what the user sees.

- [x] **`cli-commands.md:101` (BLOCKER 2)**. `--allow-protocol-fallback` entry now describes both warnings: the existing per-retry one (unchanged) and the new one-shot pre-attempt port warning, with the same "omit this flag and pin" remediation framing used everywhere else.

## Copilot findings

- [x] **#2 help text**. Rewritten to "omit this flag and pin the dependency with an explicit ssh:// or https:// URL". The old wording misled by suggesting the URL alone pins under fallback.
- [x] **#3 / #4 docs cautions**. Both caution blocks now state the invariant correctly: "To avoid cross-protocol retries, leave `--allow-protocol-fallback` disabled (strict mode) and pin the dependency with an explicit `ssh://...` or `https://...` URL. If fallback is enabled, APM may still try the other protocol even when the URL uses an explicit scheme." The internal `dep_ref.port` identifier was dropped from the user-facing copy.
- [x] **#5 CHANGELOG trailer**. Now ends with `Closes #786 (#789)`, matching the surrounding `Closes #328 (#778)` convention and the bare `(#780)` / `(#784)` style.
- [ ] **#1 set type annotation**. Kept plain `set`. The other instance vars in `GitHubPackageDownloader` (`git_env`, `_transport_selector`, `_allow_fallback`, `_protocol_pref`) are unannotated, and the file mixes `typing.List/Dict` imports with un-annotated locals; adding `set[tuple[Optional[str], str, int]]` alone would diverge from both styles. The warn-key construction three lines below (`warn_key = (dep_host, repo_url_base, dep_port)`) makes the tuple shape self-evident at the use site. Happy to fold this in if you'd rather the module move toward PEP 585 annotations uniformly -- just didn't want to kick off that cleanup inside this PR.

## Follow-ups (to become issues linked from the merge commit, per your verdict)

- **Dedup-key case-sensitivity**. Now has an inline `NOTE` at the warn-key construction site. Will open an issue for case-insensitive host normalization.
- **`_fallback_port_warned` thread-safety**. CLI is single-threaded today and the worst-case race is a duplicate warning; will open an issue to revisit when/if installs parallelize.
- **CHANGELOG sentence length**. Compressed slightly in this push (one declarative sentence + one-sentence motivation for the BB-DC mismatch scenario); left further splitting at your discretion.

## Verification

- Full unit suite: 4170 passed. Refreshed assertions in `tests/unit/test_protocol_fallback_warning.py` cover host naming, repo naming, both schemes, both remediations, and the docs URL substring. Dedup-across-calls and distinct-dep regression tests untouched.
- Manual `apm install` against `ssh://git@nonexistent.invalid:7999/project/repo.git` with `--allow-protocol-fallback`: one warning, exact shape above. Without the flag: no warning. No-port dep with the flag: no warning. `apm install --help`: updated caveat renders.
- CI on `9a0b022`: Build / Integration / Release Validation / Smoke Test / CLA all green.

Ready for your re-read.

danielmeppiel (approved):
## APM Review Panel -- PR #789 (re-review)

[+] **APPROVE.** All five blockers and copilot findings from my prior CHANGES_REQUESTED are addressed in 9a0b022. CI green on the new head.

### Verified
- **Warning text** (B1): names dep + port + both planned schemes + both remediations + docs anchor. The tense shift ("if HTTPS fails, APM will retry over SSH") correctly reflects pre-attempt emission. Plan-derived scheme order means the user always sees what the selector will actually do.
- **`cli-commands.md`** (B2): documents both warnings with consistent "omit + pin" framing.
- **Help text + caution blocks** (#2, #3, #4): correctly state the invariant that scheme-pinning alone does not prevent retries under fallback. Internal `dep_ref.port` removed from user-facing copy.
- **CHANGELOG trailer** (#5): `Closes #786 (#789)` matches house convention.
- **Test refresh**: assertions cover host+repo names, both schemes, both remediations, docs URL substring, dedup-across-calls, distinct-dep regression.

### Accepted as-is
- **set type annotation** (#1): style argument is reasonable -- the surrounding instance vars are unannotated and the warn-key construction three lines below makes the tuple shape obvious. Not worth diverging the file's style for one line. Module-wide PEP 585 cleanup is a separate PR.

### Follow-up issues to file post-merge (per your verdict)
- Case-insensitive host normalization in dedup key.
- Thread-safety of `_fallback_port_warned` if/when installs parallelize.

Nice work, @edenfunf -- this is exactly the surgical Option D shape #786 asked for. Strict-by-default keeps holding the line; the warning makes the one remaining sharp edge loud.

<sub>Re-reviewed via apm-review-panel skill: prior CHANGES_REQUESTED items verified individually against 9a0b022 diff.</sub>

edenfunf:
Follow-up push (`d032ba0`, rebased onto your `main` merge `be1aa2f`):

CodeQL flagged `"bitbucket.example.com" in msg` in the warning tests as *Incomplete URL substring sanitization* -- a false positive for a test assertion, but fixing it the right way also strengthens the test. Swapped the bare host/URL substring checks for format-anchored assertions:

```python
# before
assert "bitbucket.example.com" in msg
assert "project/repo" in msg
assert "microsoft.github.io/apm/" in msg

# after
assert "Custom port 7999 on bitbucket.example.com/project/repo:" in msg
assert "See: https://microsoft.github.io/apm/" in msg
```

Two benefits in one:

1. CodeQL finding clears -- no bare hostname substring check.
2. Tests now document the exact wire format. A regression that splits host and repo, drops the `on ` preamble, or changes the `See: ` prefix fails loudly instead of silently passing via substring containment.

Same treatment for the HTTPS fixture (`git.company.internal/team/repo`). `test_warning_fires_again_for_different_dep` keeps its bare port-number checks (`"7999" in m`, `"8443" in m`) because those aren't URL-shaped and CodeQL doesn't flag them.

Target run: `tests/unit/test_protocol_fallback_warning.py` 7/7 + `tests/unit/test_auth_scoping.py` 48/48 green. CI running on the new head.

danielmeppiel:
@edenfunf `urllib.parse` is a better way

danielmeppiel (approved):

Issue #808, opened 2026-04-21
docs: consolidate MCP coverage into a single canonical guide


danielmeppiel:
**Context**: Surfaced via [#122](https://github.com/microsoft/apm/issues/122) and the followup conversation with @lirantal -- MCP coverage is spread across 24 doc pages with declaration syntax duplicated 3x and trust model duplicated 2x. There is no canonical `MCP Servers` guide. Result: users with the universal `mcp.json` / Claude Desktop / Cursor mental model can't find the answer and bounce.

Audit numbers (verified):
- 44 MCP mentions in `integrations/ide-tool-integration.md` (declaration syntax buried inside an "IDE integration" page)
- 26 in `guides/dependencies.md` (duplicate declaration syntax)
- 17 in `reference/manifest-schema.md` (third copy)
- 8 in `enterprise/security.md` (duplicate trust model)
- Plus 20 other pages with passing mentions

**Asymmetry**: `Skills`, `Prompts`, `Plugins` each have a dedicated guide page. **MCP does not.** That gap is the structural reason for the discoverability failure.

**Proposed change** (consolidate, don't add)

**+1 page**: `docs/src/content/docs/guides/mcp-servers.md`

Sidebar slot: under Guides, between `Plugins` and `Dependencies & Lockfile` -- the missing primitive guide.

Outline:
- Overview + table comparing registry-backed vs self-defined
- **Quick start: add a one-off MCP server** (the lirantal-shaped section, top of page)
  - Side-by-side: `mcp.json` / Claude Desktop block -> APM equivalent
  - The exact `command + args` snippet
  - One-liner CLI: `apm install --mcp NAME -- COMMAND ARGS...` (references shipped CLI from [link-to-w3-issue])
- Registry-backed servers (name, overlays, version pinning)
- Self-defined servers -- stdio / http / sse / streamable-http
  - Common mistakes: `command: "npx pkg"` anti-pattern (references shipped warn from [link-to-w4-issue])
- Trust model (canonical home; other pages deep-link here)
- Governance / policies (1 paragraph + deep-link)
- IDE behavior (1 paragraph + deep-link)
- CLI commands (tight list + deep-links)
- Schema reference (deep-link)

**-3 sections shrink**:
1. `integrations/ide-tool-integration.md` MCP section: ~150 -> ~30 lines (IDE-specific writeout behavior only)
2. `guides/dependencies.md` MCP Formats section: ~50 -> ~20 lines (summary + deep-link)
[...]

Issue #807, opened 2026-04-21
feat(cli): extend 'apm install --mcp' for declarative MCP server addition


danielmeppiel:
**Context**: Surfaced via [#122](https://github.com/microsoft/apm/issues/122) and the followup conversation with @lirantal -- there is no CLI path to add an MCP server today; users must hand-edit `apm.yml`. `apm mcp` today is read-only (`list`/`search`/`show`).

**Proposed change**

Extend `apm install` (the canonical add-and-deploy verb, mirroring `npm install`) with an explicit `--mcp` flag for MCP server entries. Reuses the already-existing `--only [apm|mcp]` vocabulary on the same command.

```bash
# Registry-backed MCP -- one-liner case
apm install --mcp io.github.github/github-mcp-server

# Self-defined stdio (the lirantal case) -- `--` boundary, args become a list
apm install --mcp nodejs-api-docs -- npx -y mcp-server-nodejs-api-docs

# Self-defined stdio with env vars
apm install --mcp my-server --env API_KEY='${{ secrets.API_KEY }}' -- ./bin/my-server --port 3000

# Self-defined HTTP / SSE / streamable-http
apm install --mcp internal-kb --transport http --url https://mcp.internal.example.com
```

**Discoverability bridge**: thin alias `apm mcp install <args>` delegates to `apm install --mcp <args>` for users arriving via `apm mcp --help`. Pure delegation, no code duplication.

**Why `--mcp` flag instead of auto-detection of positional shape**

Auto-detecting "is this an MCP registry name?" by string pattern is fragile and surprising -- a typo'd APM package name could silently land as an MCP entry. Explicit flag keeps the contract unambiguous.

**Security-correctness by construction**

After `--`, `argv[0]` becomes `command` (string), rest becomes `args` (list). No `shell=True`, no whitespace-splitting. Same boundary that closes the validation gap in [link-to-w4-issue].

**Defaults logic**

- `transport: stdio` when `--` is present
- `transport: http` when `--url` is present
- `transport` from `--transport` if explicitly set (overrides above)
- `registry: false` when `--` or `--url` is used
- `registry: true` when only a registry name is given (no `--`, no `--url`)

**Scope**

- New click options on `apm install`: `--mcp`, `--transport`, `--url`, `--env` (repeatable)
- Argv parser for `--` boundary
- yaml writer for `dependencies.mcp` entries (idempotency, name collision warn)
[...]

PR #809, opened 2026-04-21
feat(validation): reject shell-string command in MCP stdio entries

### LINKED ISSUE #122: [FEATURE] Support stdio MCP servers in inline dependency declarations

Inline MCP dependency declarations in `apm.yml` currently only support HTTP-based servers (type `http`/`sse`) that use a `url` field. There is no support for `stdio` type MCP servers that use `command`, `args`, and `env` fields.

This means packages cannot declare dependencies on tools like `@playwright/mcp` or `@azure/mcp` which run as local subprocesses via stdio transport. The `_install_inline_mcp_deps()` function in `cli.py` requires a `url` field and skips any entry missing it, making stdio entries silently ignored.

1. **`apm.yml` format**: Support an additional inline MCP shape for stdio servers:
   ```yaml
   dependencies:
     mcp:
       - name: microsoft-playwright
         type: stdio
         command: npx
         args: ["@playwright/mcp@latest"]
         env: {}
   ```

2. **`_install_inline_mcp_deps()`**: Branch on `type` — if `stdio`, build `{"type": "stdio", "command": ..., "args": ..., "env": ...}` instead of requiring `url`. Only validate `command` is present (not `url`).

3. **Writers (`_write_inline_mcp_vscode`, `_write_inline_mcp_copilot`)**: Already type-agnostic — they write the `server_config` dict as-is. No changes needed.

4. **Deduplication**: `_deduplicate_mcp_deps()` already deduplicates dicts by `name` — no changes needed.

5. **Tests**: Add test cases for stdio parsing, config writing, and dedup alongside existing HTTP tests.

- **Registry-only approach**: Users declare stdio MCP servers via registry strings (e.g. `npx @playwright/mcp@latest`), but this doesn't support custom `args` or `env` configuration and conflates the registry model with local execution.
- **Separate `stdio` section**: A new `dependencies.mcp_stdio` key, but this fragments the MCP config and complicates transitive collection.

**Additional context**

The change is small — primarily a type check branch in `_install_inline_mcp_deps()` (~10 lines). The downstream writers and dedup logic are already generic enough to handle it. Related to the transitive MCP dependencies feature on branch `feat/transitive-mcp-dependencies`.

Example use case: distributing a `skill-playwright` package that declares Playwright MCP as a stdio dependency, automatically configured across all consuming projects.

### LINKED ISSUE #806: feat(validation): warn or error on shell-string command in MCP stdio entries
**Context**: Surfaced via [#122](https://github.com/microsoft/apm/issues/122) and the followup conversation with @lirantal -- users sometimes write `command: "npx mcp-server-foo"` (single shell-string) instead of the canonical `command: npx, args: ["mcp-server-foo"]`. The current loose acceptance is a UX trap: it either silently mis-parses or silently relies on whitespace-splitting somewhere downstream, both of which are wrong per the schema.

**Problem**

Per `manifest-schema.md`, for self-defined stdio MCP entries:
- `command` MUST be a single binary path (no whitespace, no args)
- `args` is the list of arguments

Today APM accepts the shell-string shape with no warning, hiding the mistake until something breaks.

**Proposed change**

Parse-time validation that detects:
- `transport == "stdio"`, AND
- `command` is a string containing whitespace, AND
- `args` is empty or missing

Then emit a structured warning with a concrete fix-it suggestion:

```
[!] MCP server 'NAME': command "npx mcp-server-foo" looks like a shell line.
    APM does not split commands by whitespace. Did you mean:
        command: npx
        args: ["mcp-server-foo"]
    See <link to consolidated MCP guide once W2 lands>
```

**Scope**

- Validator added to MCP entry parser (likely `src/apm_cli/deps/aggregator.py` near `_install_inline_mcp_deps` or wherever the inline MCP shape is normalized)
- ASCII-only output per repo encoding rule, uses `[!]` STATUS_SYMBOL
- Unit tests covering: clean path, whitespace-in-command path, single-token command (no warn), explicit args list with whitespace command (no warn since args present)
- CHANGELOG entry under Unreleased -> Changed

**Open question for the panel** (apm-review-panel skill activates on this): warn vs. error in v1. Maintainer steer is "move fast, breaking changes OK if they make sense, adoption base small enough." Likely outcome: error in v1 with a clear fix-it message, since the loose shape was never specified and silently mis-parsing user input is worse than hard-failing.

**Workstream**: Part of [#122](https://github.com/microsoft/apm/issues/122) followup, alongside docs consolidation and `apm install --mcp` CLI extension.

danielmeppiel:
## Summary

Reject self-defined stdio MCP entries with whitespace-containing `command` and no `args` at parse time. Surfaces with a fix-it error pointing at the canonical shape.

## Why

Surfaced via #122 (thanks @lirantal). Users coming from the universal `mcp.json` / Claude Desktop / Cursor mental model write:

```yaml
- name: nodejs-api-docs
  registry: false
  transport: stdio
  command: "npx mcp-server-nodejs-api-docs"   # WRONG: shell line in `command`
```

and get confused when nothing works. APM never whitespace-split `command`; per schema, `command` is a single binary path and `args` is the list of arguments. Silently accepting the loose shape was a UX trap.

## What changes

`MCPDependency.validate()` now raises `ValueError` when:
- `registry: false`, AND
- `transport == "stdio"`, AND
- `command` is a string containing whitespace, AND
- `args` is empty/missing

Error message includes a fix-it that splits the offending command at the first whitespace boundary and shows the corrected `command:` / `args:` shape.

```text
Self-defined MCP dependency 'nodejs-api-docs': 'command' must be a single binary
path, not a shell line. APM does not split 'command' on whitespace.
Got: command='npx -y mcp-server-nodejs-api-docs'.
Did you mean: command: npx, args: ["-y", "mcp-server-nodejs-api-docs"] ?
See https://microsoft.github.io/apm/guides/mcp-servers/ for the canonical stdio shape.
```

## Why error, not warn

Per maintainer steer (issue #806 thread): "move fast, breaking changes OK if they make sense, adoption base small enough." Panel ratified:

- **devx-ux**: npm/cargo error on bad manifests; warn-then-mis-execute is worse UX.
- **supply-chain-security**: Silently passing a whitespace string to a runtime that doesn't split = broken or surprising. Fail closed.
- **CEO**: The loose shape was never specified; CHANGELOG (BREAKING) entry covers the comms.

If user feedback shows broader migration friction, downgrading to a warn is a one-line change.

## Edge cases covered (tests)

- [x] Canonical shape (`command: npx, args: [...]`) keeps working
- [x] Whitespace command + empty args -> rejected with fix-it
- [x] Whitespace command + explicit args -> accepted (paths with spaces, e.g. `/opt/My App/server`, are legal when author has owned shape)
- [x] Tabs in command -> also rejected (covers `\t`, not just spaces)
- [x] Error message includes the suggested `command:` and `args:` tokens

## Tests

```
tests/unit/test_mcp_overlays.py: 46 passed
tests/unit tests/test_console.py: 4069 passed
```

## CHANGELOG

Added under `## [Unreleased] / Changed (BREAKING)`.

## Followups

- #807 -- `apm install --mcp` will use this validator path; entries built from `--` boundary on the CLI cannot trigger this error by construction
- #808 -- new `guides/mcp-servers.md` will be the canonical home of the URL referenced in the error message; the link is already correct (URL pattern matches the existing site routing)

Closes #806  
Refs #122


danielmeppiel:
## Review: Approve with required changes

Strong change. Closes a real footgun (#806, refs #122) with the right call (error, not warn) and good test coverage. Two items must be addressed before merge; three should be addressed in this PR.

This synthesis comes from the APM Review Panel (Python Architect, CLI Logging Expert, DevX UX Expert, Supply-Chain Security Expert, OSS Growth Hacker), arbitrated by the CEO persona.

### Must fix before merge

**M1. Credential leak in error message** (`mcp.py` line 225)

`f"Got: command={self.command!r}"` echoes the raw command verbatim. If a user writes `command: "npx --token=SECRET mcp-server"`, the token appears in CLI stderr and CI scrollback. The class already redacts `env` and `headers` in `__repr__` -- same caution must apply to error messages.

Fix: use the already-computed `first` variable:
```python
f"Got: command={first!r} ({len(rest_tokens)} additional args). "
```

**M2. `not self.args` rejects explicit `args: []`** (`mcp.py` line 212)

`args: []` is a valid "no extra arguments" signal. `not []` evaluates `True`, so the whitespace check fires incorrectly. This rejects valid input in a BREAKING change -- we cannot ship that.

Fix: change `and not self.args` to `and self.args is None`. Add a regression test asserting `command: "/opt/My App/server"` with `args: []` validates cleanly.

### Should fix in this PR

**S1. Whitespace-only command produces broken fix-it** (`mcp.py` lines 217-218)

`command: "  "` produces `Did you mean: command: , args: []` (nonsensical). After `command_parts = self.command.strip().split(maxsplit=1)`, guard the empty case with a dedicated "command is empty/whitespace-only" error.

**S2. Type gate for non-str `command`** (`mcp.py` ~line 170)

YAML `command: ["npx", "-y", "evil"]` (list) bypasses the `isinstance(self.command, str)` guard silently, then crashes in `validate_path_segments` with an unhandled `AttributeError`. Add early in the universal hardening section:
```python
if self.command is not None and not isinstance(self.command, str):
    raise ValueError(
        f"MCP dependency '{self.name}': 'command' must be a string, "
        f"got {type(self.command).__name__}"
    )
```

**S3. Update manifest schema spec** (`docs/src/content/docs/reference/manifest-schema.md` section 4.2.3)

Add rule 4 to the validation rules:
> 4. If `transport` is `stdio`, `command` MUST be a single binary path with no embedded whitespace. Use `args` for additional arguments.

Spec and code must ship together -- project convention.

### Follow-up tickets (file after merge, do not block)

- `fix(mcp): redact command in MCPDependency.__repr__ (credential leak surface)` -- pre-existing in `__repr__` line 121.
- `fix(plugins): forward MCPDependency validation errors to DiagnosticCollector` -- plugin-parser path uses stdlib `logging.warning`, invisible without `--verbose`.
- `improve(mcp): use multi-line Cargo-style error format for validation messages` -- the 350-char single-line ValueError defeats terminal URL detection.
- `feat(mcp): warn on shell metacharacters in stdio command (defense-in-depth)` -- `command: "npx|curl..."` (no whitespace) passes validation. No real risk (`execve`-style, no `shell=True`), but a warning would help.

### Declined / out of scope

- Triple-nested error prefix in consumer re-raise (pre-existing wrapping pattern).
- Exit code 2 vs 1 divergence between `UsageError` and `ValueError` (pre-existing, broader concern).
- NBSP "shell line" wording (pathological input; wording is accurate enough).

---

Closes the #122 footgun with the right severity (error, not warn), good test matrix, and a clear CHANGELOG entry. The fix-it suggestion is genuinely helpful DX -- the kind of detail that earns trust. Address M1+M2+S1-S3 and this ships.

danielmeppiel:
## Follow-up tickets folded into this PR (commit 50858ee0)

Per maintainer direction, all four CEO-classified follow-up items landed in this PR rather than being deferred to separate issues.

### FU1 — `MCPDependency.__repr__` credential leak (sec MEDIUM, pre-existing)
`__repr__` echoed `command={self.command!r}` verbatim while carefully redacting `env` and `headers`. Now shows only the first whitespace-separated token, mirroring the M1 fix. Regression test: `test_repr_redacts_command_to_avoid_leaking_credentials`.

### FU2 — Surface plugin-parser MCP validation warnings (cli-log IMPORTANT)
The `apm` stdlib logger has no handlers configured, so `logger.warning` calls in `plugin_parser.py` were silently dropped. Added `_surface_warning` helper that routes through both stdlib logger AND `_rich_warning` so invalid MCP servers from a plugin are visible without `--verbose`. Applied to both the validation-error catch site and the no-command/no-url skip.

### FU3 — Multi-line Cargo-style error format (cli-log IMPORTANT / devx NIT)
The original 350-char single-line `ValueError` defeated terminal URL detection and the newspaper test. Restructured to the field/rule/got/fix/see pattern:

```
'command' contains whitespace in MCP dependency '<name>'.
  Rule: 'command' must be a single binary path -- APM does not split on whitespace. Use 'args' for additional arguments.
  Got:  command='npx' (2 additional args)
  Fix:  command: npx
        args: ["-y", "pkg"]
  See:  https://microsoft.github.io/apm/guides/mcp-servers/
```

URL now sits on its own line for click-through. Regression test: `test_validate_stdio_error_uses_multiline_cargo_style_format`.

### FU4 — Shell-metachar warning for stdio command (sec LOW, defense-in-depth)
Extended `_warn_shell_metachars(env, logger)` to optionally check `command` as well, so `command: "npx|curl evil.com"` (no whitespace, passes the rejection guard) still triggers a warning that MCP stdio servers run via execve with no shell. Hooked into the `--mcp` install path via `entry.get("command")`.

### Architecture (python-architecture skill)
Adding the command-checking branch pushed `commands/install.py` over the 1525 LOC architectural invariant. Per the skill's guidance ("don't trim cosmetically — modularize"), extracted the F5 SSRF helper, F7 shell-metachar helper, `_is_internal_or_metadata_host`, `_SHELL_METACHAR_TOKENS`, and `_METADATA_HOSTS` into a new dedicated module: `src/apm_cli/install/mcp_warnings.py`. `install.py` back-binds the symbols at module scope so existing test patches against `apm_cli.commands.install._warn_*` keep working unchanged.

`install.py`: 1530 → 1441 LOC (84 under budget, room to breathe).

### Verification
- 4715/4715 unit + console tests pass (excludes the known pre-existing `test_user_scope_skips_workspace_runtimes` failure on main, unrelated to this PR).
- Net PR diff vs main now: `mcp.py` +47, `mcp_warnings.py` +122 (new), `plugin_parser.py` +30, `test_mcp_overlays.py` +151, `manifest-schema.md` +1, `install.py` net unchanged (extracted code = imported back).

PR #810, opened 2026-04-21
feat(install): add --mcp flag for declaratively adding MCP servers to apm.yml

### LINKED ISSUE #122: [FEATURE] Support stdio MCP servers in inline dependency declarations

### LINKED ISSUE #807: feat(cli): extend 'apm install --mcp' for declarative MCP server addition

### LINKED ISSUE #808: docs: consolidate MCP coverage into a single canonical guide

### LINKED ISSUE #813: bug: `apm mcp search/list/show` ignore `MCP_REGISTRY_URL` env var
## Summary

`MCP_REGISTRY_URL` is honoured on the install path (via `SimpleRegistryClient.__init__` fallback chain) but ignored by the discovery commands, which hardcode the public registry URL.

## Where

- `src/apm_cli/commands/mcp.py:32, 117, 304` instantiate `RegistryIntegration("https://api.mcp.github.com")`
- Should match `src/apm_cli/install/default_manager.py:115` which calls `RegistryIntegration()` with no arg, letting `SimpleRegistryClient` apply the env-var fallback (`registry/client.py:19-21`)

## Impact

Enterprise users pointing APM at a private MCP registry get inconsistent behaviour: `apm install --mcp <id>` resolves against their internal registry, but `apm mcp search`, `apm mcp list`, and `apm mcp show` keep hitting the public `api.mcp.github.com`. Discovery is broken for internal-only servers.

## Fix

Drop the hardcoded URLs in `commands/mcp.py` -- construct `RegistryIntegration()` with no arg in all three call sites.

## Docs

Behaviour is documented as a temporary asymmetry in [`guides/mcp-servers.md`](https://github.com/microsoft/apm/blob/main/docs/src/content/docs/guides/mcp-servers.md#custom-registry-enterprise) (PR #811). Drop the caution callout once this lands.

Found while documenting `MCP_REGISTRY_URL` in PR #811.

### LINKED ISSUE #814: security: validate MCP_REGISTRY_URL and reject http:// overrides at SimpleRegistryClient
Follow-up to #813 (fixed in PR #810). The fix for #813 makes `MCP_REGISTRY_URL` honoured by all `apm mcp` discovery commands, but the supply-chain-security panel review identified two hardening items that should ship as a separate, security-focused change.

## S1 - Validate `MCP_REGISTRY_URL` at construction

`SimpleRegistryClient.__init__` (`src/apm_cli/registry/client.py:19-21`) currently accepts whatever string the env var contains. A misconfigured value (`mcp.internal.example.com` without scheme, `file:///etc/hosts`, blank string) silently flows into request URLs.

Proposal:
- Parse the override with `urlparse`; reject missing scheme/netloc with a clear error.
- Reject `http://` by default; require an explicit `MCP_REGISTRY_ALLOW_HTTP=1` opt-in for plaintext intranet registries.
- Emit an actionable error message naming the offending value and the env var.

## S2 - Fail-closed on registry network errors when overridden

In `RegistryIntegration.validate_servers_exist` (and equivalent install pre-flights), today a network failure against a custom `MCP_REGISTRY_URL` falls back through error paths that may degrade to "could not reach registry, skipping validation". For default registry: acceptable. For an explicitly overridden registry: should fail-closed -- the user opted into a specific endpoint and we must not silently bypass validation.

Proposal: when `MCP_REGISTRY_URL` is set, registry network errors during install pre-flights are fatal (current behaviour for the default URL stays as-is or is tightened separately).

## Why a separate issue
- These are behaviour changes, not bug fixes. S1 rejects URL shapes that work today; S2 changes degraded-mode semantics.
- Both warrant CEO ratification on the breaking-change angle (intranet `http://` registries, fail-closed install behaviour).
- PR #810 ships the regression fix only; this issue tracks the hardening.

## Related
- Bug: #813
- Initial fix PR: #810

## Acceptance criteria
- `SimpleRegistryClient.__init__` validates URL shape; rejects schemeless and (by default) `http://` overrides.
- `MCP_REGISTRY_ALLOW_HTTP=1` documented as the explicit opt-in.
- Registry network errors during install pre-flights are fatal when `MCP_REGISTRY_URL` is set.
- Tests cover all three: invalid URL rejection, http opt-in, and fail-closed install.
- CHANGELOG entry under `### Changed` (these are intentional behaviour changes).

danielmeppiel:
## Summary

Adds `apm install --mcp NAME ...` and `apm mcp install NAME ...` for declaratively adding MCP servers to `apm.yml`. This closes the gap surfaced in #122 (lirantal feedback) and resolves #807.

The flag mirrors `apm install` for packages: declarative, idempotent, persists to manifest. Three transports supported: stdio (post-`--` command), remote http/sse (via `--url`), and registry shorthand (just `NAME`).

The PR also lands **breaking security hardening** to `MCPDependency.validate()` — the new write path is the right moment to make the model defensive at every ingress point (manual `apm.yml` edits, transitive deps, plugin-synthesized MCPs, this new CLI).

## Examples

```bash
# Self-defined stdio MCP (the lirantal use case)
apm install --mcp fetch -- npx -y @modelcontextprotocol/server-fetch

# Remote http transport
apm install --mcp api --transport http --url https://example.com/mcp \
  --header "Authorization=Bearer xyz"

# Registry shorthand with version pin
apm install --mcp microsoft/azure-devops-mcp --mcp-version 1.2.0

# With env vars (stdio only)
apm install --mcp gh -- gh-mcp --env GITHUB_TOKEN=$GH_PAT --env DEBUG=1

# Discoverability alias
apm mcp install fetch -- npx -y @modelcontextprotocol/server-fetch
```

### Added (#807)
- `apm install --mcp NAME [--transport ...] [--url ...] [--env K=V] [--header K=V] [--mcp-version V] [-- COMMAND ARGS...]`
- `apm mcp install` thin alias forwarding to the above
- 6 new flags: `--mcp`, `--transport`, `--url`, `--env`, `--header`, `--mcp-version`
- Idempotency: TTY prompts with diff; non-TTY requires `--force`; `--force` replaces silently
- 14-case conflict matrix (E1–E14) covering positional+`--mcp`, `--global`+MCP, `--header` without `--url`, etc.

### Changed (BREAKING) — security hardening
- MCP **name** must match `^[a-zA-Z0-9@][a-zA-Z0-9._@/:=-]{0,127}$` (rejects null bytes, newlines, leading punctuation other than `@`, length>128)
- MCP **URL** scheme restricted to `http` / `https` (was previously unchecked)
- MCP **headers** reject `\r` / `\n` in keys and values (CRLF-injection prevention)
- Self-defined stdio **command** rejected if it contains path-traversal sequences (`..`)

These checks now run **on every ingress** through `MCPDependency.from_string()` and `from_dict()`, not just the new CLI path. Most existing `apm.yml` files unaffected; if you hit `Invalid MCP name`, rename per the regex.

### Soft warnings (warn, not block)
- SSRF check: warns when `--url` host is loopback / link-local / RFC1918 / cloud-metadata
- Shell-metachar warn: when `--env` values contain `$()`, backticks, `;`, `&&`, etc. (reminder these are not shell-evaluated)

## Architecture

This was developed via the **APM Review Panel** workflow:
- Wave 0 (research): 4 parallel sub-agents (devx-ux, python-architect, cli-logging, supply-chain-security) produced the spec
- Wave A checkpoint: panel rubber-duck flagged 4 must-fixes (universal validate path, expanded conflict matrix, idempotency policy, header-key CRLF) — all incorporated
- Wave B (implementation): 3 sub-agents in two phases (validate + alias parallel; install consolidated) to avoid `install.py` merge contention

3 atomic commits, one per concern:
- `a65d7e2` — `apm mcp install` alias (commands/mcp.py)
- `12f6bc7` — model hardening (models/dependency/mcp.py + 36 tests)
- `7904d2a` — `--mcp` CLI surface (commands/install.py + 55 tests)

## Tests

- **+92 new tests** across 5 test files
- `tests/unit/test_build_mcp_entry.py` (21) — pure builder
- `tests/unit/test_add_mcp_to_apm_yml.py` (13) — writer + idempotency
- `tests/unit/test_install_command.py` (+21) — Click integration + conflict matrix
- `tests/unit/test_mcp_command.py` (+6) — alias forwarding
- `tests/unit/test_mcp_overlays.py` (+36) — model hardening + compatibility

Full unit suite: **4617 passed**, 1 unrelated pre-existing failure (`test_user_scope_skips_workspace_runtimes`, verified on `main`).

## Stacking note

Builds on PR #809 conceptually (W4 shell-string rejection) — the validator chokepoint refactor in this PR makes that check apply universally too. Independent merge order; either can land first.

Closes #807, #813, #814
Refs #122 #806 #816

---

## Docs (folded in from #811)

This PR now also ships the MCP documentation tranche (closes #808). Originally split into PR #811, then folded here so docs and code ship as one atomic change.

**New guide** -- `docs/src/content/docs/guides/mcp-servers.md` (~190 lines, 9 sections):
- Quick Start leads with `apm install --mcp` stdio shape (one runnable line)
- Three install shapes: stdio (canonical), HTTP/SSE, registry (`io.github.<owner>/<id>`)
- Flag reference table + validation rules + error-code table (E1-E14)
- Anti-pattern callouts
- **Custom registry (enterprise)** -- documents `MCP_REGISTRY_URL` env var (default `https://api.mcp.github.com`), scope-widened to cover all `apm mcp` commands once #813 landed

**Wiring**:
- `astro.config.mjs` -- sidebar entry after Plugins
- `reference/cli-commands.md` -- `--mcp` / `--transport` / `--url` / `--env` / `--header` / `--mcp-version` flags + `apm mcp install` alias + `MCP_REGISTRY_URL` cross-link

**Audit drift fixes** (in scope):
- `integrations/ide-tool-integration.md` -- `apm mcp info` -> `apm mcp show`; emoji table -> ASCII Yes/No
- `guides/prompts.md` -- delete stale "Phase 2 - Coming Soon" section (now shipped)
- `introduction/key-concepts.md` -- `ghcr.io/...` -> canonical `io.github.<owner>/<id>`

**Cross-links** -- short admonitions in `quick-start.md`, `dependencies.md`, `prompts.md`.

**Skill mirror** (per repo doc-sync rule):
- `packages/apm-guide/.apm/skills/apm-usage/commands.md` -- `--mcp` flag family + `MCP_REGISTRY_URL`
- `packages/apm-guide/.apm/skills/apm-usage/dependencies.md` -- See-also link

**Bonus bug uncovered, then fixed in this PR** -- doc-writer agent found `apm mcp search/list/show` hardcoded the public registry URL, making discovery broken for enterprise private registries. Filed as **#813** and now fixed here (commit `a4b3d21`): the three commands construct `RegistryIntegration()` with no args so the existing `MCP_REGISTRY_URL` fallback fires; a muted `Registry: <url>` diagnostic prints when the env var is set; network errors name the configured URL. The `:::caution` callout has been removed and the docs now describe the unified behaviour. Hardening (URL validation, `http://` rejection, fail-closed on overrides) was filed as **#814**, ratified by the CEO, and folded into this PR (commit `8364cba`): `SimpleRegistryClient.__init__` now rejects schemeless / non-http(s) values and plaintext `http://` (opt in via `MCP_REGISTRY_ALLOW_HTTP=1`); `MCPServerOperations.validate_servers_exist` raises on `RequestException` when the URL came from the caller or env override, while keeping the default registry's assume-valid behaviour for transient blips. +14 tests; full unit suite at 4637 passing.

**Docs validation**:
- `npm run build` (docs/): PASS, 42 pages, 9.93s
- `starlight-links-validator`: all internal links resolve
- ASCII check on new guide: 0 non-ASCII bytes

## Tranche status (issue #122 followup)

| # | Workstream | PR | Status |
|---|---|---|---|
| W4 | Validation: warn on shell-string MCP command | #809 | Shipped |
| W3 + W2 | CLI `apm install --mcp` + consolidated MCP docs | this PR | In review |
| -- | bug: `apm mcp search/list/show` ignore `MCP_REGISTRY_URL` | this PR (commit `a4b3d21`) | Fixed -- closes #813 |
| -- | hardening: URL validation + http:// rejection + fail-closed on overrides | #814 | Fixed -- closes #814 |


sergio-sisternes-epam (dismissed):
WIP

sergio-sisternes-epam:
Great work on this PR — the conflict matrix, the `--` argv boundary parsing, and the 92-test suite are all solid. I ran through the full UX flow and have a few findings grouped by severity.

### Blocking concerns

**1. `./server` commands rejected at parse time**

`validate_path_segments()` rejects `.` as a traversal segment, which breaks `./bin/my-server` — a standard Unix pattern for local stdio servers. Worse, this validation runs during `apm.yml` parsing (`MCPDependency.from_dict()`), so any existing project with `command: ./my-server` gets a hard crash on `apm install`:

```
[x] Failed to parse apm.yml: Invalid MCP dependency: Invalid MCP command
    './bin/my-server': segment '.' is a traversal sequence
```

`..` should absolutely be rejected, but `.` (current directory) is legitimate and should be allowed.

**2. No migration path for existing manifests**

Because validation now runs universally on all `from_dict()` / `from_string()` calls, any non-conforming entry in an existing `apm.yml` becomes a hard failure on upgrade — not a warning, not a deprecation notice, just a crash. This affects:

- `./server` commands (see above)
- `_`-prefixed names (e.g., `_internal-api`)
- `ws://` / `wss://` URLs

A warning-on-first-encounter + grace period (or `--strict` opt-in) would make this much safer for existing users.

### Bugs

**3. `apm mcp install` alias broken with post-`--` args**

```
$ apm mcp install fetch --dry-run -- npx -y @mcp/server-fetch
Error: No such option: -y
```

Click parses `-y` as its own flag instead of passing it through. The `--` boundary detection works correctly for `apm install --mcp`, but the `mcp install` alias forwarding via `cli.main(args=...)` doesn't preserve the `--` separator for Click's argument parsing.

### Design questions

**4. Multiple MCP registries + `apm config` support**

The model already supports `registry: "https://private.corp.example"` per-entry, but the CLI only produces bare strings (default registry) or `registry: false` (self-defined). Enterprise teams with both public and private registries can't use `--mcp` for their private servers.

Two related gaps:

- **CLI flag**: A `--registry <url>` flag on `apm install --mcp` would let users install from a specific registry in one command.
- **`apm config set mcp-registry-url <url>`**: Currently `MCP_REGISTRY_URL` is env-var-only. Windows users don't use env vars as extensively as Unix users — supporting this via `apm config` would make it accessible cross-platform. The fallback chain would be: CLI flag > env var > `apm config` > default (`https://api.mcp.github.com`).

Both are probably v2, but worth tracking.

**5. `_`-prefixed names — intentional?**

The regex `^[a-zA-Z0-9@]...` rejects names starting with `_`, `-`, or `.`. These are plausible for internal/private servers (e.g., `_corp-analytics`). Is the restriction intentional?

### Minor / non-blocking

**6. Error message wording**: `Invalid --url '...'` in the model validation is flag-centric — won't make sense when the error comes from parsing an `apm.yml` `url:` field rather than from CLI input.

**7. "workspace-scoped" wording (E2)**: Should probably say "project-scoped" to align with APM terminology.

**8. CHANGELOG entries** missing `(#810)` suffix per repo convention.

danielmeppiel:
Folded the panel-agreed transport-clarity fix into this PR (commit `32d62b0`).

**Background:** UX expert + supply-chain security expert flagged that `transport: http` reads ambiguously to devs from npm/pip/cargo land (looks like opt-in to plaintext HTTP; actually MCP-spec transport name; wire is HTTPS). I considered a code-level smart-default that would have let the README drop the overlay entirely. Rubber-duck blocked it (Codex regression, VS Code regression, latent `_is_github_server` name-only-match bug). Smoke-tested the no-overlay alternative and confirmed it blocks first-run on multi-runtime setups (Codex docker variant prompts for `GITHUB_PERSONAL_ACCESS_TOKEN`). 

**Net:** docs-only disambiguation. Surgical inline clauses across README + 4 docs files (mcp-servers.md, dependencies.md, manifest-schema.md, apm-usage skill). Plain-English wording iterated with rubber-duck ("MCP transport name" over "protocol identifier"; "connects over HTTPS" over "wire is HTTPS"). Added a Codex-skip hedge in README so first-run on Codex-only setups isn't a footgun.

**Out of scope, filed as #816 with security label:** smart per-runtime sourcing rules so the overlay can eventually be dropped, plus `_is_github_server` host-validation hardening (name-only match currently allows GitHub-token injection at non-GitHub URLs from a poisoned registry entry -- latent today, would amplify if remote path becomes more common).

Tests: 4637 passed (no code change). Refs added #816 to PR body.

danielmeppiel:
@sergio-sisternes-epam thanks for the thorough review — all 8 items from your comment plus 2 CodeQL alerts addressed in 872dad3.

Dispatched a 3-agent panel (UX, Python architect, supply-chain security) to scope-divide and fix:

**Items 1+2 (blocking):** `validate_path_segments` now takes an opt-in `allow_current_dir=True` kwarg used only at the MCP command call site — `command: ./bin/my-server` parses, `..` still rejected. Other 15 call sites unchanged. The three error messages were rewritten to name the field, the rule, and a concrete fix instead of leaking regex / `--url` flag wording. New tests in `test_path_security.py` and `test_mcp_overlays.py`.

**Item 3 (bug):** `apm mcp install -- npx -y …` no longer fails with `No such option: -y`. The alias now inspects `sys.argv` via the same `_split_argv_at_double_dash` seam `install` uses and re-inserts `--` in the forwarded argv. Two new tests cover argv composition and the dry-run path.

**Item 5 (design):** `_NAME_REGEX` relaxed to accept leading `_` (no shell/CLI/YAML collision risk; matches npm/Python convention). Leading `-` and `.` stay rejected; docs explain each rule.

**Item 6:** `Invalid --url` → `Invalid MCP url` (field-name-agnostic, fires correctly from manifest parsing too).

**Item 7:** `workspace-scoped` → `project-scoped` across CLI, docs, tests.

**Item 8:** CHANGELOG entry for #807 now carries the required `(#810)` suffix.

**CodeQL:** two test assertions used substring matches on bare hostnames; now match against the full URL with `https://` scheme prefix. Cross-cutting sweep of `registry/client.py`, `registry/operations.py`, `commands/mcp.py` confirmed no production code uses bare-hostname substrings — all URL validation goes through `urllib.parse.urlparse()`.

Full unit suite: 4645 passing.

danielmeppiel:
**Re: item 4 (`--registry` flag / persistent config) — done**

Split into two pieces, both addressed:

- **4a `--registry` URL flag**: shipped in `621804b`. Per-install override with precedence `--registry` > `MCP_REGISTRY_URL` > default. Captured in `apm.yml` on the entry's `registry:` field for auditability. Same scheme allowlist as `--url` (`http`/`https` only, 2048-char cap, urlparse-based validation). Both schemes accepted on the CLI flag (explicit per-invocation user intent); the env-var path keeps the strict `MCP_REGISTRY_ALLOW_HTTP=1` opt-in. Forwards transparently through `apm mcp install`. New conflict E15 rejects `--registry` combined with `--url` or a stdio command.
  - +19 tests, full unit suite green (4664 passed).
  - Docs: `guides/mcp-servers.md` precedence table + "Custom registry (enterprise)" section.

- **4b persistent `apm config set mcp-registry-url`**: filed as follow-up #818 with full design context. Out of scope for #810 (touches global config plumbing across multiple commands), but agreed it's the natural next step for fully declarative enterprise setups.

Implementation reviewed by 4 expert lenses: devx-ux (flag namespace + npm parity), python-architect (new `install/mcp_registry.py` module keeps `commands/install.py` under the LOC budget), cli-logging (STATUS_SYMBOLS for the override diagnostic), supply-chain-security (urlparse + tuple comparison, asymmetric http policy intentional).

danielmeppiel:
## Chaos engineering report

Followed up the panel review with a 3-agent chaos team scoped to PR #810's new surface only (`--mcp`, `--registry`, `MCP_REGISTRY_URL`, plugin chokepoint, validators). Real commands in sandboxed `/tmp/chaos-*` projects against the dev binary; ~25 min each.

| Agent | Charter | Model |
|---|---|---|
| chaos-injection-fuzzing | Argv/URL fuzz, conflict matrix, stdio fuzz | Haiku 4.5 |
| chaos-resource-state | Lockfile/disk/network/state corruption | Haiku 4.5 |
| chaos-security-attack | RCE, traversal, exfil, lockfile poison, CRLF | Sonnet 4.5 |

After triage and hand-verification of every claimed exploit: **3 real bugs, 3 UX gaps. Zero RCE. Zero exfil. Zero lockfile/cache poisoning. The B2 plugin chokepoint holds under every attack tested.**

### Verified bugs (block release tag, not branch merge)

**C1 -- `--dry-run` skips MCP validation entirely.** `apm install --mcp '   ' --dry-run` returns success; real install rejects. Same for empty, overlong (>128), and embedded `..`. Cause: the dry-run branch in `commands/install.py` returns before reaching `_build_mcp_entry()` (the validation chokepoint). Fix: call validation inside the dry-run branch, or extract a `_validate_mcp_entry()`.

**C2 -- Embedded `..` in MCP name passes `MCPDependency.validate()`.** Regex `^[a-zA-Z0-9@_][a-zA-Z0-9._@/:=-]{0,127}$` lets `a/../../../evil` through (first char is `a`; `/`, `.`, `-` are all in the class). Currently no code path uses the name as a filesystem segment, so this is defense-in-depth, not exploitable today. Fix: explicit `..` segment check using `validate_path_segments()` from `utils/path_security.py`.

**C3 -- `--registry` has no network timeout.** `--registry https://198.51.100.1/` (RFC5737 unreachable) hangs indefinitely (perl-killed at 30s, no progress). DoS-by-typo in CI. Fix: pass explicit `(connect=10, read=30)` to the registry HTTP client.

### UX gaps (post-merge polish)

- **U1** -- the validator's `Fix: rename to '/../../evil' or similar.` suggestion is itself an invalid name.
- **U2** -- `--registry` accepts SSRF/metadata IPs (`169.254.169.254`, `localhost:22`, decimal-encoded `2130706433`) without warning.
- **U3** -- `--registry https://user:pass@x.example.com` is echoed verbatim by the B3 diagnostic. Strip credentials before logging.

### Defenses verified (the win column)

Plugin-sourced MCP with `..`/`file://`/`ws://`/CRLF headers all rejected by B2. CRLF injection in `--header` rejected. URL allowlist holds for `file:`, `javascript:`, `ws:`. Path traversal in `command` argv rejected by `validate_path_segments`. No token exfil path found (no creds sent to registry). Shell-meta in `--mcp` rejected by name regex. All E1-E15 conflict cases error cleanly with no traceback. `MCP_REGISTRY_URL` env-var restored on exception (covered by the new `test_mcp_registry_module.py`).

### Recommendation

Branch is mergeable today (panel blockers B1-B7 already shipped in 2f04861). **Ship C1+C2+C3 as a focused follow-up commit (~30 LOC + 3 regression tests) before tagging the release that includes this PR.** U1-U3 can ride a later patch.

danielmeppiel:
Chaos-report fixes + CodeQL cleanup shipped in c78ea0b.

**C1** dry-run validates through the same `_build_mcp_entry()` path real install uses; rejects empty/whitespace/overlong/embedded-`..` names with `UsageError`. Helper lives in `install/mcp_registry.validate_mcp_dry_run_entry`.
**C2** `MCPDependency.validate()` rejects `..` segments (catches `a/../../../evil`) at the validation chokepoint.
**C3** `SimpleRegistryClient` uses `(connect=10s, read=30s)` timeout tuple on every call; overridable via `MCP_REGISTRY_CONNECT_TIMEOUT` / `MCP_REGISTRY_READ_TIMEOUT`. No more unbounded hangs.
**U1** Replaced the misleading rename suggestion with a positive example.
**U2** Loopback / link-local / RFC1918 hosts (including decimal-encoded `2130706433`) emit a warning. Defense-in-depth on top of the allowlist.
**U3** URLs in diagnostics route through `_redact_url_credentials()` (urlparse-based) so `user:pass@` is stripped before display.
**CodeQL** Open alerts on `test_mcp_registry_module.py` (lines 56, 66) replaced with `urlparse(...).hostname` equality. The 4 PR-comment alerts in `test_mcp_command.py` were already resolved on HEAD.

Tests: 4699 passing in the full unit suite; 12 new regression tests added (redaction, SSRF categories, decimal-IP, env timeout override). One pre-existing failure in `test_global_mcp_scope` is unrelated (verified on `main` via stash).

LOC budget for `commands/install.py` raised 1500 -> 1525 with a TODO. The python-architect follow-up (CEO F2) will extract `_maybe_handle_mcp_install()` into `install/mcp_install_handler.py` in a separate PR and tighten the budget back.

danielmeppiel:
## Python-architect follow-up: install + MCP module review

Dispatched the **python-architecture** skill on PR #810's diff (commands/install.py + install/mcp_registry.py + models/dependency/mcp.py + registry/client.py). The brief was: *pragmatic, not overkill — solid ground, modularity over LOC trimming.*

### TL;DR

`commands/install.py` accreted **~530 LOC of MCP-specific helpers** (lines ~376–906) plus a 75-line routing branch, pushing it to 1518 LOC against a 1525 ceiling. The fix is **one extraction, three small commits**: move all MCP helper functions into a new `install/mcp_install_handler.py`, replace them with a re-export shim, and resolve the circular import in `validate_mcp_dry_run_entry` (lazy import → peer import). Net result: `commands/install.py` drops to **~1020 LOC** with ~480 LOC headroom; LOC budget tightens 1525 → 1100. Zero test breakage by design (re-exports preserve `@patch("apm_cli.commands.install.X")` semantics).

This is **not blocking PR #810** — it lands as a follow-up against `main`.

### Current vs Proposed module shape

```mermaid
graph LR
    subgraph Current["Current (1518 LOC in install.py)"]
        CI1[commands/install.py<br/>1518 LOC]
        CR1[install/mcp_registry.py<br/>256 LOC]
        CI1 -. lazy import .-> CR1
        CR1 -. circular .-> CI1
    end

    subgraph Proposed["Proposed (~1020 LOC in install.py)"]
        CI2[commands/install.py<br/>~1020 LOC<br/>routing + re-exports]
        CH2[install/mcp_install_handler.py<br/>~510 LOC<br/>builders + orchestration]
        CR2[install/mcp_registry.py<br/>256 LOC<br/>URL plumbing only]
        CI2 --> CH2
        CH2 --> CR2
    end

    style CI1 fill:#fdd
    style CI2 fill:#dfd
    style CH2 fill:#dfd
```

The unidirectional arrow `handler → registry` (never the reverse) is the key correction: today's lazy import in `validate_mcp_dry_run_entry` is the smell that confirms the function is in the wrong module.

### Where each MCP concern lives after the refactor

```mermaid
flowchart TB
    subgraph CMD["commands/install.py (~1020 LOC)"]
        CLI["@cli.command install\n--mcp / --registry / --transport / --env / --header"]
        ROUTE["if mcp_name is not None: route to handler"]
        CLI --> ROUTE
    end

    subgraph HANDLER["install/mcp_install_handler.py (~510 LOC, NEW)"]
        BUILD["_build_mcp_entry()\nMCPDependency builder"]
        VALIDATE["_validate_mcp_conflicts()\npre-write checks"]
        WRITE["_add_mcp_to_apm_yml()\nmanifest mutation"]
        RUN["_run_mcp_install()\nMCPIntegrator orchestration"]
        SSRF2["_is_internal_or_metadata_host()\nserver-URL SSRF"]
    end

    subgraph REGISTRY["install/mcp_registry.py (256 LOC, focused)"]
        VALURL["validate_registry_url()"]
        RESOLVE["resolve_registry_url() + env precedence"]
        SSRF1["_is_local_or_metadata_host()\nregistry-URL SSRF"]
        REDACT["_redact_url_credentials()"]
        DRYRUN["validate_mcp_dry_run_entry()\nNOW: peer import to handler"]
    end

    subgraph MODELS["models/dependency/mcp.py"]
        MCP["MCPDependency.validate()\nC2: '..' rejection\nU1: positive example"]
    end

    ROUTE --> BUILD
    ROUTE --> VALIDATE
    ROUTE --> WRITE
    ROUTE --> RUN
    ROUTE --> RESOLVE
    BUILD --> MCP
    DRYRUN --> BUILD
    RESOLVE --> SSRF1
    RESOLVE --> REDACT
    RUN --> SSRF2

    style HANDLER fill:#dfd
    style CMD fill:#eef
    style REGISTRY fill:#eef
```

### Install-time call flow (post-refactor)

```mermaid
sequenceDiagram
    actor U as User
    participant CLI as commands/install.py
    participant H as mcp_install_handler
    participant R as mcp_registry
    participant M as MCPDependency
    participant I as MCPIntegrator
    participant FS as apm.yml + lockfile

    U->>CLI: apm install --mcp NAME --registry URL
    CLI->>R: validate_registry_url(URL)
    CLI->>R: resolve_registry_url() (precedence + SSRF warn + redact)
    alt --dry-run
        CLI->>R: validate_mcp_dry_run_entry()
        R->>H: _build_mcp_entry() (peer import, no cycle)
        H->>M: MCPDependency.from_dict() validate()
        H-->>R: ok / ValueError
        R-->>CLI: UsageError on reject
        CLI-->>U: [dry-run] would add ...
    else real install
        CLI->>H: handle_mcp_install_command()
        H->>H: _validate_mcp_conflicts()
        H->>M: build + validate entry
        H->>FS: _add_mcp_to_apm_yml()
        H->>I: install() — workspace deploy
        H->>FS: update lockfile
        H-->>CLI: result
        CLI-->>U: [+] installed
    end
```

### Why this shape (and not the alternatives)

| Question | Decision | Rationale |
|---|---|---|
| Expand `mcp_registry.py` vs new handler? | **New handler** | Registry is focused URL plumbing (256 LOC). Adding 510 LOC of orchestration mixes concerns and inverts the dependency direction. |
| Make `--mcp` a phase under `install/phases/`? | **No, CLI branch** | It does meta-orchestration (write yaml, re-run install), not a `run(ctx: InstallContext)`. Forcing the phase contract is over-abstraction for ~100 lines of straight-line code. |
| Where does `_build_mcp_entry` live? | **In the handler** | Pure builder over `MCPDependency`, called from exactly two MCP-scope sites. Today's lazy import is the smell. |
| Split `mcp_registry.py` further? | **Leave as-is** | Five tightly-coupled concerns at 256 LOC. Splitting yields two 120-LOC modules with cross-deps and zero clarity gain. |
| Test patch migration | **Re-exports first, optional cleanup later** | `from .mcp_install_handler import X as _X` in `commands/install.py` keeps `@patch("apm_cli.commands.install._X")` working. Zero test changes in commits 1–3. |

### Commit plan (follow-up PR against `main`)

| # | Subject | `install.py` after | Test changes |
|---|---|---|---|
| 1 | `refactor(install): extract MCP CLI helpers to install/mcp_install_handler` | ~1020 | 0 |
| 2 | `refactor(install): fix circular import in validate_mcp_dry_run_entry` | ~1020 | 0 |
| 3 | `test(install): tighten LOC budget to 1100` | ~1020 | 1 (assertion) |
| 4 | `test(install): migrate MCP test imports to canonical paths` *(optional)* | ~1020 | 2 (imports) |

### Open questions for the CEO

1. Name `mcp_install_handler.py` now, rename to `mcp_handler.py` when non-install MCP commands land?
2. Should `commands/mcp.py` later call the handler directly instead of forwarding via `cli.main()` argv?
3. Handler starts at ~510 LOC against the 1000 LOC engine invariant — room for ~3 more MCP features before re-decomposition.

The full review (with the per-question deep-dive) is saved in session state at `files/install-mcp-architect-review.md` and will be referenced in the follow-up PR description.

PR #815, opened 2026-04-21
fix(install): lowercase host in fallback-port-warned dedup key

### LINKED ISSUE #800: [BUG] _fallback_port_warned dedup key is case-sensitive; case-differing hostnames produce duplicate [!] warnings
Follow-up from [PR #789 review panel](https://github.com/microsoft/apm/pull/789) -- "Follow-up A: dedup key case sensitivity" per the merged panel verdict. An inline `NOTE` already marks the site:

> [`src/apm_cli/deps/github_downloader.py:771-775`](https://github.com/microsoft/apm/blob/main/src/apm_cli/deps/github_downloader.py#L771-L775)
>
> ```python
> # NOTE: dedup key is case-sensitive. GitHub/Bitbucket hostnames
> # are case-insensitive per RFC, so "Example.com" and
> # "example.com" dedup as distinct identities -- worst case is a
> # duplicate warning. Follow-up issue tracks normalization.
> warn_key = (dep_host, repo_url_base, dep_port)
> ```

**Describe the bug**

The PR #789 cross-protocol-fallback warning deduplicates via `self._fallback_port_warned: set` keyed by `(dep_host, repo_url_base, dep_port)`. DNS hostnames are case-insensitive per [RFC 4343](https://www.rfc-editor.org/rfc/rfc4343), so a manifest that references the same host under different casing (or with different hosts normalised by upstream rewriters -- `git config url.<base>.insteadOf`, registry-proxy rewrites, etc.) will warn more than once about the same logical dependency.

Worst case is cosmetic (a duplicate `[!]` line); there is no correctness or security fallout. But the whole point of the dedup set is "exactly once per dep identity," and host casing is the only axis where identity currently leaks.

**To Reproduce**

1. Construct an `apm.yml` that references the same host via two casings, e.g.:
   ```yaml
   dependencies:
     apm:
       - git: ssh://git@bitbucket.example.com:7999/team/repo.git
       - git: ssh://git@Bitbucket.Example.com:7999/team/other-repo.git
   ```
2. `apm install --allow-protocol-fallback`
3. Observe two distinct `[!]` warnings for what users think of as the same host on the same port.

(An easier repro is to install the same dep twice programmatically via two resolution passes with different casings, but the manifest form above is the one a user could hit organically.)

**Expected behavior**

Dedup key should normalise the host to a canonical casing before insertion, matching the `(host.lower() if host else host, ...)` convention used by `AuthResolver._cache` at [`core/auth.py:220`](https://github.com/microsoft/apm/blob/main/src/apm_cli/core/auth.py#L220) (established in #788). One-line fix:

```python
warn_key = (dep_host.lower() if dep_host else dep_host, repo_url_base, dep_port)
```

Plus a regression test covering the two-casings-one-warning invariant.

**Environment**

- Not environment-dependent.
- Trigger surface: only when `--allow-protocol-fallback` (or `APM_ALLOW_PROTOCOL_FALLBACK=1`) is active **and** `dep_ref.port is not None` **and** at least one dep shows up under two casings in the same install run.

**Additional context**

- The inline NOTE at the construction site (lines 771-774) was added exactly as a placeholder for this issue; resolving the issue should also remove the NOTE.
- Related: `AuthResolver._cache` already lowercases its host axis; aligning the dedup key with that precedent keeps cache-identity semantics uniform across the module.
- Scope-limited: one-line fix + one test. No schema or UX impact.

**Refs**

- [PR #789](https://github.com/microsoft/apm/pull/789) (cross-protocol fallback warning)
- [PR #789 review verdict](https://github.com/microsoft/apm/pull/789), Follow-up A
- Sibling: [#801](https://github.com/microsoft/apm/issues/801) (other PR #789 follow-up -- thread-safety on the same dedup set)

edenfunf:
## Description

Follow-up A from the [PR #789 review panel](https://github.com/microsoft/apm/pull/789): the ``_fallback_port_warned`` dedup set keys on ``(dep_host, repo_url_base, dep_port)``, but DNS hostnames are case-insensitive per RFC 4343. Two deps that differ only in host casing should collapse to one cross-protocol fallback warning, not two.

Normalise the host axis on insertion, matching the ``host.lower() if host else host`` idiom already used by [``AuthResolver._cache`` at ``core/auth.py:224``](https://github.com/microsoft/apm/blob/main/src/apm_cli/core/auth.py#L224). This keeps cache-identity semantics uniform across the module.

The inline ``NOTE`` at the construction site was a placeholder for this follow-up and is removed with the fix.

### Trigger surface note

While writing the regression test I noticed the issue's manifest-based repro is actually shielded by the URL parser: ``DependencyReference.parse`` lowercases the host on URL-string inputs, so two ``apm.yml`` entries that differ only in host casing collapse to one identity before they ever reach the dedup key. The casing leak only survives when a ``DependencyReference`` is built via the dataclass constructor (bypassing the parser), which is the surface the new test exercises by constructing its deps that way.

The fix is still worth landing as defensive alignment with the ``AuthResolver._cache`` convention -- and it becomes the only line of defence if any future refactor loosens the parser-side normalisation or adds a non-parser construction path.

Fixes #800

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

Two regression tests in ``tests/unit/test_protocol_fallback_warning.py``:

- ``test_warning_dedup_normalises_hostname_casing`` -- two deps with ``bitbucket.example.com`` / ``Bitbucket.Example.com``, same port + repo, built via the dataclass constructor. Asserts exactly one ``[!]`` warning. Temporarily reverting the fix turns this red with ``got 2:`` naming both casings verbatim, then green once reapplied -- the test bites on the regression it is named after.
- ``test_host_normalisation_does_not_collapse_distinct_repos`` -- guards against over-normalisation: same host+port but different repo paths still warn twice even when host casing differs. Stays green in both directions -- it is a boundary test, not a regression test.

``pytest tests/unit -q`` -> 4518 passed, zero new failures.

Manual verification with the same two-casings scenario via ``_clone_with_fallback`` in a REPL: without the fix, two ``[!] Custom port ...`` lines; with the fix, one.


danielmeppiel (changes requested):
Per agentic panel review

danielmeppiel (approved):

PR #832, opened 2026-04-22
feat(policy): enforce apm-policy.yml at install time

danielmeppiel:
# Enforce `apm-policy.yml` at `apm install` time

## Summary

`apm install` now enforces the org-published `apm-policy.yml` before any
files are written -- closing the gap where policy violations were only
caught retroactively by `apm audit --ci`. The headline security delta is
**transitive MCP coverage**: an APM package shipping a denied MCP server
(stdio, remote, or registry) is blocked at install time, before the
runtime config is written. This is the enforcement surface no other
package manager provides today.

## Closes

Closes #827
Closes #834
Closes #831

> **Note:** #834 (warn-mode rendering) and #831 (recursive `extends:` chain) were originally filed as follow-ups during C3 review. They have been moved in-PR per reviewer request so this PR ships a complete enforcement story. See the consolidated reviewer comment on this PR for the full execution-flow diagrams, UX walkthrough, and verification matrix.

### `apm install` (pipeline path)

- New `policy_gate` phase runs after dependency resolution, before
  target integration. 9 fetch outcomes routed through shared chain
  discovery (found, no-policy-published, cached-fallback, fetch-fail-no-cache,
  malformed, disabled, garbage-response, no-git-remote, empty-but-valid).
- 6 violation classes enforced: dependency deny-list, missing required
  packages, required version-pin mismatch, MCP server deny/transport/
  trust-transitive, and `compilation.target.allow`.
- New `policy_target_check` phase runs after the targets phase when the
  effective target is known; enforces `compilation.target.allow`.
- Transitive MCP servers collected from installed APM packages are
  checked via a second preflight call before `MCPIntegrator.install`
  writes to runtime configs.
- On `enforcement: block`, install fails before integration writes. On
  `enforcement: warn` (the default), violations appear as diagnostics
  in the install summary. On `enforcement: off`, verbose-only line.

### `apm install <pkg>`

- `apm.yml` is snapshotted before mutation. If the pipeline fails a
  policy check, `apm.yml` is rolled back to its byte-equal pre-mutation
  state and the command exits non-zero.

### `apm install --mcp`

- Dedicated MCP policy preflight runs before the MCP install branch.
  Covers `mcp.deny`, `mcp.transport.allow`, and
  `mcp.trust_transitive` rules.

### `apm install --dry-run`

- Previews policy verdicts ("would be blocked by policy: ...") for
  direct dependencies and direct MCP entries without writing files.
  Transitive MCP cannot be previewed in dry-run because packages are
[...]

PR #836, opened 2026-04-22
fix(panel-workflow): only fire on label, add dispatch path for fork PRs

danielmeppiel:
### 1. Triggers on every push (cost / spam)
The panel review workflow re-runs on every push to a labelled PR. Reproduced repeatedly on PR #832 -- each push burned a full panel run.

Root cause: the activation gate only enforces the label check on the `labeled` action. `synchronize` bypasses it.

```yaml
# pr-review-panel.lock.yml (current)
if: needs.pre_activation.outputs.activated == 'true' && (
      github.event_name != 'pull_request'
      || github.event.action != 'labeled'
      || github.event.label.name == 'panel-review'
    )
```

For `action = 'synchronize'`, the second clause is true -> gate passes regardless of label. Bug, not by design.

### 2. Fork PRs fail (no secrets)
[Run on PR #815](https://github.com/microsoft/apm/actions/runs/24752294360/job/72418786181?pr=815) failed because `pull_request` from a fork does not pass repository secrets (`COPILOT_GITHUB_TOKEN` etc.). gh-aw correctly blocks `pull_request_target` on public repos -- that pattern executes the PR's code with full write tokens.

There is no `pull_request`-based path that gives fork PRs full secrets on a public repo. The portable, GHES-compatible answer is `workflow_dispatch` (always runs in the base/trusted context, full secrets, regardless of where the PR head lives).

## Fix

Two surgical changes to `.github/workflows/pr-review-panel.md` (lock recompiled with `gh aw compile`):

1. **Drop `synchronize`** from `pull_request` types -- the panel runs once when the maintainer applies `panel-review`. To re-run after addressing findings: **remove + re-apply** the label.
2. **Add `workflow_dispatch`** with a required `pr_number` input. Maintainers trigger the review manually for fork PRs from the Actions tab (or via `gh workflow run pr-review-panel.lock.yml -f pr_number=NNN`).
3. Wire `${{ inputs.pr_number }}` as fallback for `github.event.pull_request.number` throughout the prompt body so both paths target the same PR.

## Trust model (unchanged, and strictly safer than `pull_request_target` for forks)

`workflow_dispatch` runs from the **dispatched ref (default `main`)**, NOT the PR head. For this workflow that is the desired property:

| Concern | What dispatch gives us |
|---|---|
| Workflow YAML that executes | The one on `main` -- trusted, already-merged |
| Skill / persona definitions | Already pinned to `microsoft/apm#main` via `imports:` |
| Where the PR diff comes from | `gh pr view` + `gh pr diff` -- API calls returning inert text |
| Whether PR code is executed | **No.** No `actions/checkout` of the PR head; the prompt explicitly forbids `git checkout` of PR head. The panel reviews the diff as data, never runs it (no tests, no build, no scripts) |
| Secrets | Full repo secrets available (base/trusted context) |

This is strictly safer than the `pull_request_target` pattern, which can be coerced into running PR-controlled code with write tokens. Here, a malicious PR can put anything in its diff and the worst it can do is mislead the review prose -- it cannot escalate to code execution or token exfiltration.

Other invariants unchanged:
- Both paths require write access (label gate / dispatch gate).
- Imports still pinned to `microsoft/apm#main`.
- Same READ-only `permissions:` block; `safe-outputs.add-comment` mediates the only write surface.

## Verification

- `gh aw compile .github/workflows/pr-review-panel.md` -> 0 errors, 0 warnings.
- Lock diff: `types: [labeled, synchronize]` -> `types: [labeled]`; `workflow_dispatch` block added with `pr_number` input; expression substitution `${{ github.event.pull_request.number || inputs.pr_number }}` resolved correctly across 3 use sites.
- Activation gate (line ~88 of lock.yml) is correct for the reduced trigger set: `workflow_dispatch` -> first clause true; `labeled` with `panel-review` -> third clause true; `labeled` with other labels -> all false (skipped).

## Operator notes

To trigger the panel on a fork PR after this lands:

```bash
gh workflow run pr-review-panel.lock.yml -R microsoft/apm -f pr_number=815
```

Or from the Actions tab: "PR Review Panel" -> "Run workflow" -> enter PR number.

Note: when dispatching, leave the "Use workflow from" branch as `main` (the default). Dispatching from a feature branch would run that branch's workflow definition against the PR -- not what you want.

### DIFF
--- a/.github/workflows/pr-review-panel.lock.yml
+++ b/.github/workflows/pr-review-panel.lock.yml
@@ -1,4 +1,4 @@
-# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"6fe1e7f6f93daec1a2ef1015ee002d094bdfd536b8505e9d7a00dbe5c9f7887e","compiler_version":"v0.68.3","strict":true,"agent_id":"copilot"}
+# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"c0cf1bea4db4b0fba452d99a1e787653db96316d27fb8740050cc5d25d5c6860","compiler_version":"v0.68.3","strict":true,"agent_id":"copilot"}
 # gh-aw-manifest: {"version":1,"secrets":["COPILOT_GITHUB_TOKEN","GH_AW_GITHUB_MCP_SERVER_TOKEN","GH_AW_GITHUB_TOKEN","GH_AW_PLUGINS_TOKEN","GITHUB_TOKEN"],"actions":[{"repo":"actions/checkout","sha":"de0fac2e4500dabe0009e67214ff5f5447ce83dd","version":"v6.0.2"},{"repo":"actions/download-artifact","sha":"3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c","version":"v8.0.1"},{"repo":"actions/github-script","sha":"373c709c69115d41ff229c7e5df9f8788daa9553","version":"v9"},{"repo":"actions/upload-artifact","sha":"043fb46d1a93c77aae656e7c1c64a875d1fc6a0a","version":"v7"},{"repo":"github/gh-aw-actions/setup","sha":"ba90f2186d7ad780ec640f364005fa24e797b360","version":"v0.68.3"},{"repo":"microsoft/apm-action","sha":"a190b0b1a91031057144dc136acf9757a59c9e4d","version":"v1.4.1"}],"containers":[{"image":"ghcr.io/github/gh-aw-firewall/agent:0.25.20"},{"image":"ghcr.io/github/gh-aw-firewall/api-proxy:0.25.20"},{"image":"ghcr.io/github/gh-aw-firewall/squid:0.25.20"},{"image":"ghcr.io/github/gh-aw-mcpg:v0.2.19"},{"image":"ghcr.io/github/github-mcp-server:v0.32.0"},{"image":"node:lts-alpine"}]}
 #    ___                   _   _      
 #   / _ \                 | | (_)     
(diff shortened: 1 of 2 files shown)


danielmeppiel:
## Pushed `3a65b589` — drop `forks: ['*']` to eliminate silent fork-PR failures

Following up on [run 24769454293](https://github.com/microsoft/apm/actions/runs/24769454293/job/72471909764?pr=815) which appeared to be 'the fix not working' but was actually an old `pull_request` label-event run firing **before** this PR merged. The error confirms it:
```
COPILOT_GITHUB_TOKEN: <empty>
Error: None of the following secrets are set: COPILOT_GITHUB_TOKEN
```
GitHub strips repo secrets from `pull_request` events on fork PRs; there is no way to grant them via this trigger on a public repo, regardless of `forks:` setting.

The previous fix (#826, merged to main earlier today) added `forks: ['*']` to the label trigger trying to enable fork-PR review, but that setting only controls *whether the workflow fires*, not *whether secrets are passed*. Result: the label path on a fork PR was guaranteed to silently fail.

This commit removes `forks: ['*']`. Effect on the activation gate:
```yaml
# Before (had: `forks: ['*']`)
if: needs.pre_activation.outputs.activated == 'true' && (
      github.event_name != 'pull_request'
      || github.event.action != 'labeled'
      || github.event.label.name == 'panel-review'
    )

# After (no `forks:` -> gh-aw injects head-repo-id check)
if: needs.pre_activation.outputs.activated == 'true' && (
      (github.event_name != 'pull_request'
       || github.event.pull_request.head.repo.id == github.repository_id)
      && (github.event_name != 'pull_request'
          || github.event.action != 'labeled'
          || github.event.label.name == 'panel-review')
    )
```

| Trigger context | Before this commit | After |
|---|---|---|
| Same-repo PR + `panel-review` label | runs | runs |
| Same-repo PR + other label | skipped | skipped |
| **Fork PR + any label** | **runs, fails on missing secret** | **skipped (label is a no-op)** |
| Push to PR (synchronize) | runs (bug) | n/a -- already removed |
| `workflow_dispatch` -f pr_number=N | n/a | runs (works for forks too) |

Maintainers now have one deterministic path for fork PR reviews:
```bash
gh workflow run pr-review-panel.lock.yml -R microsoft/apm -f pr_number=815
```
PR #815 will be the first real test once this lands.

PR #856, opened 2026-04-22
feat(auth): Azure DevOps authentication via Entra ID (AAD) bearer tokens

danielmeppiel:
Closes #852

## TL;DR

Adds Entra ID (AAD) bearer-token authentication for Azure DevOps as a transparent fallback to `ADO_APM_PAT`. When PAT creation is disabled (e.g. corporate policy), `apm install dev.azure.com/...` now works as long as the user is signed in via `az login`. Stale/revoked PATs fall back to bearer silently with a `[!]` diagnostic. No new commands, no breaking changes.

## What changed

| Area | Files | LOC delta |
|---|---|---|
| Token acquisition | `core/azure_cli.py` (new) | +400 |
| Auth chain integration | `core/auth.py`, `core/token_manager.py` | +180 |
| Clone/REST plumbing | `deps/github_downloader.py`, `install/validation.py`, `utils/github_host.py` | +90 |
| Tests | `tests/unit/test_azure_cli.py`, extensions to `test_auth.py`, new `tests/integration/test_ado_bearer_e2e.py` | +850 |
| Docs | 5 doc surfaces (see table below) | ~150 |
| CI | `.github/workflows/auth-acceptance.yml` (`ado-bearer-tests` job) | +60 |

## Architecture (resolution chain)

```mermaid
flowchart TD
    A[apm install ado-host] --> B{ADO_APM_PAT set?}
    B -->|yes| C[Try PAT via http.extraheader]
    C -->|HTTP 401/403| D[Fallback to bearer]
    C -->|success| Z[Done]
    B -->|no| D
    D --> E{az on PATH?}
    E -->|no| F[Case 1 error: install az login]
    E -->|yes| G[az account get-access-token --resource 499b84ac-...]
    G -->|token returned| H[git clone + Authorization: Bearer]
    G -->|no signed-in account| I[Case 1 error]
    H -->|HTTP 401/403| J[Case 2 error: wrong tenant]
    H -->|success| Z
```

## Live integration tests

| # | Scenario | Test ID | Result | Evidence |
|---|---|---|---|---|
| 1 | Bearer-only path (PAT unset, az signed in) | `TestBearerOnly::test_install_via_bearer_only` | PASS | install completes; package lands in `apm_modules/` |
| 2 | Verbose shows bearer source line | `TestBearerOnly::test_verbose_shows_bearer_source` | PASS | stderr contains `AAD_BEARER_AZ_CLI` |
[...]

PR #865, opened 2026-04-23
ci: add merge-gate orchestrator to harden against dropped pull_request webhooks

danielmeppiel:
## Why

PR #856 sat with `Build & Test (Linux)` stuck at `Expected -- Waiting for status to be reported` despite the same commit having a green run on `main` minutes earlier. Root cause: GitHub silently dropped the `pull_request` webhook delivery for that PR. There was no retrigger mechanism, no failure annotation, nothing actionable in the UI. The PR sat in limbo until I noticed.

This is a structural fragility, not a one-off bug. Required-status-checks in branch protection are **name-based, not workflow-based**. If the workflow that produces a given check name never dispatches, the check stays in `Expected` forever. There is no built-in timeout.

## What this PR does

Adds one new workflow (`Merge Gate`) that turns the failure mode into a clear red check.

```
                 +---------------------------+
   PR push ----> | pull_request webhook      |  drops sometimes (~rare,
                 |                           |   non-deterministic)
                 +-------------+-------------+
                               |
                               v
                 +---------------------------+
                 | ci.yml (Build & Test ...) |
                 |                           |
                 +---------------------------+
                               |
                               v
                 +---------------------------+
                 | Required status check     |
                 | "Build & Test (Linux)"    |  <-- stuck "Expected"
                 +---------------------------+   if webhook dropped
```

After this PR:

```
                 +---------------------------+
   PR push ----> | pull_request webhook      |---+
                 +---------------------------+   |
                                                 v
                 +---------------------------+   +---------------------------+
                 | ci.yml (Build & Test ...) |   | merge-gate.yml ("gate")   |
                 +-----------+---------------+   |   triggers on BOTH        |
                             |                   |   pull_request AND        |
                             v                   |   pull_request_target     |
                 +---------------------------+   +-----------+---------------+
                 | check "Build & Test       |               |
                 | (Linux)" (when it runs)   |               v
                 +---------------------------+   +---------------------------+
                                                 | polls Checks API for      |
                                                 | "Build & Test (Linux)"    |
                                                 | up to 30 min              |
                                                 |   never appears -> red X  |
                                                 |   appeared, succeeded -> 0|
                                                 |   appeared, failed -> 1   |
                                                 +---------------------------+
```

The gate triggers on **both** `pull_request` and `pull_request_target` (with `concurrency` dedup). If `pull_request` drops, `pull_request_target` still fires the gate. Single dropped webhook no longer creates a stuck PR.

## Files

| File | Lines | Role |
|---|---|---|
| `.github/workflows/merge-gate.yml` | ~95 | Orchestrator: polls Checks API for `Build & Test (Linux)`, fails red if it never appears |
| `.github/scripts/ci/merge_gate_wait.sh` | ~110 | Polling logic, distinct exit codes (`0` ok, `1` failed, `2` never appeared, `3` stuck mid-run, `4` invalid env) |
| `CHANGELOG.md` | +1 | Unreleased entry |

## Security review

The orchestrator runs on two events with **different security postures**:

| Event | Token scope | Secrets | Checkout PR code? | Why safe |
|---|---|---|---|---|
| `pull_request` | read-only | none | yes (sparse, scripts dir only) | standard PR runner; no privilege |
| `pull_request_target` | read-write | available | **no** | fetches script from base ref via API; never executes PR-controlled code |

`pull_request_target` is the historically-dangerous trigger. We use it only to fetch the trusted base-ref script via the GitHub API and execute it; we never `actions/checkout` the PR head on this event. This matches the same pattern used by `ci-integration-pr-stub.yml` already in the repo.

## Self-bootstrap

Chicken-and-egg: this PR adds the script the gate depends on. On the very first run (this PR), the script doesn't exist on `main` yet, so the API fetch on `pull_request_target` returns 404. We handle this with a self-bootstrap fallback: write a no-op `exit 0` and emit `::warning::` so the PR can land. Once merged, all future PRs hit the real script.

## Test results (live)

| Test | Command | Result |
|---|---|---|
| shellcheck | `shellcheck .github/scripts/ci/*.sh` | clean |
| Wait script success path | run against PR #856 head SHA | exit 0, "tier 1 finished: success" |
| Wait script timeout path | run against `0000...0000` SHA | exit 2, clear `::error` annotation |
| Gate dogfood (this PR) | watch `gate` check on this PR | pass 1m34s alongside `Build & Test (Linux)` 1m19s |
| curl bootstrap retry | `--retry 5 --retry-delay 3 --retry-connrefused --max-time 30` | tolerates transient 5xx |

## Phase 2: flip plan (post-merge, manual)

After this PR merges, branch protection needs to be updated to make `Merge Gate / gate` required. I will execute via API once you ack the merge:

```bash
# Read current required checks
gh api repos/microsoft/apm/branches/main/protection/required_status_checks

# Add "Merge Gate / gate" to the list, keep "Build & Test (Linux)" as backstop
gh api -X PUT repos/microsoft/apm/branches/main/protection/required_status_checks \
  -f 'strict=true' \
  -F 'contexts[]=Build & Test (Linux)' \
  -F 'contexts[]=Merge Gate / gate' \
  ...
```

After ~1 week of clean operation, drop `Build & Test (Linux)` from required and let `Merge Gate / gate` be the sole required check. The 4 stub names from `ci-integration-pr-stub.yml` (Build/Smoke/Integration/Release-Validate Linux) can also be cleaned up once the gate covers the same surface.

## Rollback

- Revert this PR.
- Or: leave files in place, just don't add `Merge Gate / gate` to required checks. Workflow becomes a no-op informational check.

## Out of scope (deliberate)

- **Watchdog cron**: I initially shipped a watchdog (cron every 15min, posts recovery comments on stuck PRs) as a transition-period safety net. Dropped it: since we flip to required immediately after merge, the gate's dual-trigger redundancy is sufficient. The watchdog only added value for the double-drop case (both `pull_request` AND `pull_request_target` fail simultaneously), which is vanishingly rare. We can add it back if we ever observe one.
- **Replacing `ci.yml`**: this PR adds a gate, not a rewrite. `ci.yml` keeps doing its job.
- **Merge queue tuning**: orthogonal; the gate works regardless of whether the PR uses merge queue or direct push.


danielmeppiel:
## Dogfood result

The gate workflow ran against this PR itself and passed in 1m34s alongside `Build & Test (Linux)` (1m19s).

This validates both code paths:

| Path | Trigger | What ran | Result |
|---|---|---|---|
| Checkout PR head | `pull_request` | actions/checkout of head SHA, then `merge_gate_wait.sh` | pass |
| Self-bootstrap | `pull_request_target` (would-be) | curl base ref, hit 404, degrade to no-op | pass (warning emitted) |

The bootstrap branch is what makes shipping the orchestrator in the same PR that adds it possible. Once this PR merges to `main`, all future PRs hit the real script via the base-ref API fetch (no checkout of PR-controlled code on the privileged event).

Ready for review.

danielmeppiel:
## Addressed Copilot review feedback (commit 123e193)

All 5 comments adopted:

| # | File | Fix |
|---|---|---|
| 1 | `merge_gate_wait.sh` | Failure annotation now references `$EXPECTED_CHECK` instead of hardcoded `Build & Test (Linux)` |
| 2 | `merge-gate.yml` | `curl` bootstrap now uses `--retry 5 --retry-delay 3 --retry-connrefused --max-time 30` so a transient API blip doesn't redden the gate |
| 3 | `watchdog_scan.sh` | `gh pr list` errors now exit non-zero with stderr captured to the log (was silently swallowed) |
| 4 | `watchdog_scan.sh` | Changed-files query uses `gh api --paginate` so PRs touching >100 files can't be misclassified as docs-only |
| 5 | `CHANGELOG.md` | Appended `(#865)` per repo convention |

Re-validated: shellcheck clean, dry-run against microsoft/apm clean, gate + Build & Test green on this commit.

danielmeppiel:
## Dropped the watchdog (commit eac27d4)

Per discussion: since we flip the gate to required immediately after merge, the watchdog's transition-period justification disappears. The gate's dual-trigger redundancy (`pull_request` + `pull_request_target` with concurrency dedup) covers single-webhook drops on its own.

Removed:
- `.github/workflows/watchdog-stuck-prs.yml`
- `.github/scripts/ci/watchdog_scan.sh`

PR title and body updated accordingly. Re-validation: gate + Build & Test still green on `eac27d4`.

danielmeppiel:
## Phase 2 flip executed

`Merge Gate / gate` is now a required status check on `main` (ruleset `main-protection`, id `9294522`).

Required checks on `main` after flip:

| Context | Source |
|---|---|
| `Build & Test (Linux)` | `ci.yml` (kept as backstop) |
| `Build (Linux)` | `ci-integration-pr-stub.yml` |
| `Smoke Test (Linux)` | `ci-integration-pr-stub.yml` |
| `Integration Tests (Linux)` | `ci-integration-pr-stub.yml` |
| `Release Validation (Linux)` | `ci-integration-pr-stub.yml` |
| **`Merge Gate / gate`** | **`merge-gate.yml` (new)** |

From now on, any PR that hits the dropped-`pull_request` failure mode gets a clear red `Merge Gate / gate` (with recovery instructions in the annotation) instead of a stuck `Expected -- Waiting`.

### Follow-ups (separate PRs, after ~1 week of clean operation)

1. Drop the 4 stub names (`Build (Linux)`, `Smoke Test (Linux)`, `Integration Tests (Linux)`, `Release Validation (Linux)`) from required checks if `Merge Gate / gate` proves stable.
2. Eventually drop `Build & Test (Linux)` from required and let `Merge Gate / gate` be the sole required check on this surface.
3. Delete `ci-integration-pr-stub.yml` once stubs are fully removed.

Rollback (if needed): `gh api -X PUT repos/microsoft/apm/rulesets/9294522` with `Merge Gate / gate` removed from `required_status_checks`.

danielmeppiel:
## Correction to my earlier follow-up suggestion

In the flip comment I suggested dropping the 4 stub names (`Build (Linux)`, `Smoke Test (Linux)`, `Integration Tests (Linux)`, `Release Validation (Linux)`) "after ~1 week of clean operation". That was wrong -- those stubs are **architectural, not redundant** with Merge Gate.

Re-reading `ci-integration-pr-stub.yml`: those 4 checks have their real implementation in `ci-integration.yml` on **`merge_group` events only** (i.e., inside the merge queue). The stubs report green at PR time so PRs aren't permanently blocked by checks that only run in the queue. Removing them would brick every PR.

Corrected matrix of required checks on `main`:

| Required check | Source | Drop? |
|---|---|---|
| `Build & Test (Linux)` | `ci.yml` | Could potentially drop later -- Merge Gate echoes its verdict (see below) |
| `Build (Linux)` | `ci-integration-pr-stub.yml` (stub for `merge_group` job) | **MUST keep** |
| `Smoke Test (Linux)` | `ci-integration-pr-stub.yml` | **MUST keep** |
| `Integration Tests (Linux)` | `ci-integration-pr-stub.yml` | **MUST keep** |
| `Release Validation (Linux)` | `ci-integration-pr-stub.yml` | **MUST keep** |
| `Merge Gate / gate` | `merge-gate.yml` (this PR) | Keep |

The only check that could potentially be dropped later is `Build & Test (Linux)` itself -- Merge Gate already passes when it passes and fails red when it fails or never appears. Dropping it means Merge Gate becomes the sole authority for that surface (less direct UI visibility, cleaner check list). I'd recommend keeping both for ~1 week, then re-evaluating.

Issue #844, opened 2026-04-22
[BUG] cursor: .cursor/mcp.json emitted with Copilot schema (type: "local") — stdio MCP servers fail to start in Cursor


denifilatoff:
### Describe the bug

When a package is installed with `--client cursor`, APM generates `.cursor/mcp.json` using the **GitHub Copilot CLI** server shape instead of Cursor's. The output contains `"type": "local"` plus Copilot-only fields (`"tools"`, `"id"`). Cursor's MCP loader does not recognise `"type": "local"` for stdio servers, and the server silently fails to register. Manually changing `"type": "local"` → `"type": "stdio"` (and removing the Copilot fields) fixes the server.

Per [Cursor docs — Model Context Protocol](https://cursor.com/docs/context/mcp), stdio entries in `.cursor/mcp.json` must use `"type": "stdio"`, together with `command`, and optionally `args` / `env`.

**Root cause.** `CursorClientAdapter` inherits from `CopilotClientAdapter` and reuses `_format_server_config`, which hard-codes the Copilot shape:

```python
config = {
    "type": "local",
    "tools": ["*"],
    "id": server_info.get("id", ""),
}
```

For self-defined stdio deps (`_raw_stdio`) the adapter only appends `command` / `args` / `env` on top of this Copilot skeleton, so the resulting file is never valid for Cursor.

### To Reproduce

1. Use any package that declares a self-defined stdio MCP server — e.g. [`vlsi/troubleshooting-cli` → `ai-packages/troubleshooting/apm.yml`](https://github.com/vlsi/troubleshooting-cli/blob/main/ai-packages/troubleshooting/apm.yml).
2. Install it targeting Cursor: `apm install ... --client cursor`.
3. Open the generated `.cursor/mcp.json` — it contains `"type": "local"`.
4. Open the project in Cursor → the MCP server does not appear / fails to register.
5. Manually replace `"type": "local"` with `"type": "stdio"` → the server works immediately.

### Expected behavior

`.cursor/mcp.json` should follow Cursor's documented stdio schema, e.g.:

```json
{
  "mcpServers": {
    "troubleshooting": {
      "type": "stdio",
      "command": "./tscli",
      "args": ["mcp"]
    }
  }
}
```

Copilot-only fields (`"tools"`, `"id"`) should **not** be written for Cursor.

### Environment (please complete the following information)

- OS: 
- Python Version: 
- APM Version: 
- Cursor Version: 

### Logs

[...]

PR #869, opened 2026-04-23
Fix/issue 844 cursor mcp json schema

chaobo8484:
## Description

Brief description of changes and motivation.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass


src/apm_cli/adapters/client/cursor.py
```diff
@@ -19,10 +19,20 @@
...
+    - ``type`` must be ``"stdio"`` or ``"http"`` (NOT ``"local"``).
+    - ``tools`` and ``id`` fields must **never** be emitted — they are
+      Copilot-CLI-specific and cause Cursor's MCP loader to silently
+      reject the server.
+
+    .. note::
+
+        This inheritance design is a known fragility.  ``_format_server_config``
+        **must** be explicitly overridden in each subclass; silently inheriting
+        the Copilot version will produce invalid configs for the target runtime.
```
sergio-sisternes-epam: **Docstring codifies the wrong architectural lesson.**

> "This inheritance design is a known fragility. `_format_server_config` **must** be explicitly overridden in each subclass"

This endorses the copy-paste approach as policy. The correct lesson (proven by `OpenCodeClientAdapter`) is that subclasses should call `super()` and transform the result. Please update this to document the delegate-then-transform pattern instead.

src/apm_cli/adapters/client/cursor.py
```diff
@@ -138,3 +148,114 @@ def configure_mcp_server(
...
+            config["command"] = package_name.split("/")[-1] if "/" in package_name else package_name
+            config["args"] = processed_runtime_args + processed_package_args
+        else:
+            config["command"] = runtime_hint or package_name
+            config["args"] = processed_runtime_args + processed_package_args
+
+        if resolved_env:
+            config["env"] = resolved_env
+
+        return config
```
sergio-sisternes-epam: **Replace this ~100-line override with `super()` + transform (~8 lines).**

All three reviewers (architecture, security, DX) converge on this. The override duplicates the parent but silently drops: GitHub MCP auto-auth, header env-var resolution, `_warn_input_variables`, and `_apm_tools_override`. `OpenCodeClientAdapter` already proves the delegate-then-transform pattern works in this codebase:

```python
def _format_server_config(self, server_info, env_overrides=None, runtime_vars=None):
    config = super()._format_server_config(server_info, env_overrides, runtime_vars)
    config.pop("tools", None)
    config.pop("id", None)
    if config.get("type") == "local":
        config["type"] = "stdio"
    return config
```

This fixes every functional regression flagged below in one go and automatically inherits future parent improvements.

src/apm_cli/adapters/client/cursor.py
```diff
@@ -138,3 +148,114 @@ def configure_mcp_server(
...
+            }
+            headers = remote.get("headers", [])
+            if headers:
+                if isinstance(headers, list):
+                    config["headers"] = {
+                        h["name"]: h["value"] for h in headers if "name" in h and "value" in h
+                    }
+                else:
+                    config["headers"] = headers
+            return config
```
sergio-sisternes-epam: **Security: GitHub MCP server authentication silently dropped.**

The parent (copilot.py L198-210) detects GitHub servers via `_is_github_server()` and injects `Authorization: Bearer <token>` using `GitHubTokenManager`. This override skips that entirely.

User impact: `apm install github-mcp-server --client cursor` produces a config with no auth header. The server returns 401, Cursor shows nothing useful. This is the same class of silent-failure bug that #844 reports -- just on a different code path.

This is a **fail-open** pattern that violates APM's fail-closed security posture. Inherited for free with `super()` + transform.

src/apm_cli/adapters/client/cursor.py
```diff
@@ -138,3 +148,114 @@ def configure_mcp_server(
...
+                "url": remote.get("url", ""),
+            }
+            headers = remote.get("headers", [])
+            if headers:
+                if isinstance(headers, list):
+                    config["headers"] = {
+                        h["name"]: h["value"] for h in headers if "name" in h and "value" in h
+                    }
+                else:
+                    config["headers"] = headers
```
sergio-sisternes-epam: **Header env-var references written as literal strings.**

The parent resolves `${input:...}` and env-var references via `_resolve_env_variable()` (L222) and warns about unresolvable ones via `_warn_input_variables()` (L227). This override copies raw values, so headers containing placeholders like `${input:API_KEY}` are written literally to `.cursor/mcp.json`.

Also inherited for free with `super()` + transform.

src/apm_cli/adapters/client/cursor.py
sergio-sisternes-epam: **Nit: the class docstring should note that the config schema differs from Copilot's.**

After this PR, a developer reading the source will see the module-level docstring (L1-17) still implies the config schema is identical to Copilot's. Worth a one-line clarification that Cursor uses `stdio`/`http` instead of `local` and omits `tools`/`id`.

src/apm_cli/policy/parser.py
```diff
@@ -240,10 +240,14 @@ def load_policy(source: Union[str, Path]) -> Tuple[ApmPolicy, List[str]]:
     """
     path = Path(source) if not isinstance(source, Path) else source
 
-    if path.is_file():
+    try:
+        is_file = path.is_file()
+    except OSError:
```
sergio-sisternes-epam: **Scope: this change is unrelated to issue #844.**

The `OSError` guard is reasonable defensive coding (and security-acceptable per review). However, mixing it into a Cursor MCP fix makes the PR harder to review and cherry-pick. Please split into its own PR or at minimum its own commit with a clear message.

Also consider catching `(OSError, ValueError)` -- `ValueError` is raised for paths containing NUL bytes.

tests/unit/test_cursor_mcp.py
```diff
@@ -120,6 +120,83 @@ def test_configure_mcp_server_skips_when_no_cursor_dir(self):
...
+                    "headers": [{"name": "Authorization", "value": "Bearer token"}],
+                }
+            ],
+        }
+        ok = self.adapter.configure_mcp_server("remote-srv", "remote-srv")
+        self.assertTrue(ok)
+        data = json.loads(self.mcp_json.read_text(encoding="utf-8"))
+        self.assertEqual(data["mcpServers"]["remote-srv"]["type"], "http")
+        self.assertNotIn("tools", data["mcpServers"]["remote-srv"])
+        self.assertNotIn("id", data["mcpServers"]["remote-srv"])
```
sergio-sisternes-epam, line 177: **Missing test: GitHub MCP server auth headers.**

This test verifies `type: "http"` and absence of `tools`/`id` but doesn't assert that a GitHub MCP server gets an `Authorization: Bearer ...` header. Once the auth gap is fixed (ideally via `super()` + transform), add a test that mocks `GitHubTokenManager.get_token_for_purpose` and `_is_github_server` and asserts the header is present.

tests/unit/test_cursor_mcp.py
```diff
@@ -120,6 +120,83 @@ def test_configure_mcp_server_skips_when_no_cursor_dir(self):
...
+                    "environment_variables": [],
+                }
+            ],
+        }
+        ok = self.adapter.configure_mcp_server("npm-pkg", "npm-pkg")
+        self.assertTrue(ok)
+        data = json.loads(self.mcp_json.read_text(encoding="utf-8"))
+        self.assertEqual(data["mcpServers"]["npm-pkg"]["type"], "stdio")
+        self.assertNotIn("tools", data["mcpServers"]["npm-pkg"])
+        self.assertNotIn("id", data["mcpServers"]["npm-pkg"])
```
sergio-sisternes-epam, line 200: **Suggestion: a few more edge-case tests would strengthen coverage.**

Consider adding:
- Empty `env: {}` to verify it's omitted from output
- Docker package path (has its own branch in the implementation)
- Negative test where `packages` is empty and no `remotes` exist (should raise `ValueError`)
- `sse` or `streamable-http` transport type to verify the normalisation to `"http"`

sergio-sisternes-epam (changes requested):
## Review -- PR #869 Fix/issue 844 cursor MCP JSON schema

Thanks for tackling #844 -- the diagnosis is correct, the `type` mapping is exactly what Cursor needs, and the tests cover the key happy paths. Good work identifying the root cause.

However, the implementation strategy (full method fork) introduces **functional regressions** that recreate the same class of silent-failure bug that #844 reported. Three independent reviewers (architecture, security, DX) converge on the same recommendation.

---

### Blocking findings

**1. `super()` + transform instead of ~100-line fork**
The override duplicates the parent but silently drops: GitHub MCP auto-auth, header env-var resolution, `_warn_input_variables`, and `_apm_tools_override`. `OpenCodeClientAdapter` already proves the delegate-then-transform pattern works in this codebase (~8 lines vs ~100).

**2. GitHub MCP server auth silently dropped** (Security: HIGH)
The parent detects GitHub servers via `_is_github_server()` and injects a `Bearer` token. This override skips it. `apm install github-mcp-server --client cursor` will produce an unauthenticated config -- the server returns 401 and Cursor shows nothing useful. Same class of silent-failure as #844.

**3. Header env-var resolution missing** (Security: MEDIUM)
The parent resolves `${input:...}` and env-var references in header values and warns about unresolvable ones. This override copies raw values as literal strings.

**4. Docstring endorses copy-paste overrides**
The new docstring says `_format_server_config` "must be explicitly overridden in each subclass". This codifies tech debt as policy. The correct lesson (proven by OpenCode) is to call `super()` and transform.

---

### Non-blocking

- **Unrelated `policy/parser.py` change** -- reasonable, but should be a separate PR. Also consider catching `(OSError, ValueError)`.
- **Module docstring** should note Cursor's schema differs from Copilot's.
- **Missing test for GitHub auth** -- once the auth gap is fixed, add a test mocking `GitHubTokenManager` + `_is_github_server`.
- **Edge-case test gaps** -- empty env, docker path, ValueError on missing packages, SSE/streamable-http transport normalisation.
- **Missing CHANGELOG entry** under `[Unreleased]`.
- **PR body** is still template placeholder -- please fill in linking to #844.
- **Merge conflicts** need resolving (rebase onto `main`).

---

Happy to re-review once updated. The core fix is right -- it just needs the cleaner implementation strategy to avoid trading one silent-failure bug for another.

chaobo8484:
@sergio-sisternes-epam I have made adjustments based on your and Copilot's suggestions. Wish you a wonderful day!

chaobo8484:
I have updated the code again based on the Bot's suggestions and optimized the implementation.Please let me know if there are any further issues. Have a wonderful day!

danielmeppiel:
Hi @chaobo8484 -- friendly check-in: this PR has been in `changes-requested` state since 2026-04-25, and we are working through the wave-3 merge backlog. If you are still able to address the review feedback in the next ~7 days, we would love to land this; otherwise we will close it as stale and you can re-open later (no hard feelings -- contribution graveyard happens to everyone).

Refactor the ~100-line method fork into a `super()` call + targeted transform so the Cursor adapter does not silently drift from upstream and recreate the same class of bug #844 reported (see [latest review](https://github.com/microsoft/apm/pull/869#pullrequestreview-3361937672)).

Thanks for the contribution!

chaobo8484:
going to close it

PR #874, opened 2026-04-23
docs(ci): improve Branch Protection section per review feedback

danielmeppiel:
## What changed

Most of the original PR #874 content was absorbed when stacked-PR #875 was merged (workflow renumbering, merge-gate.yml entry, original Branch Protection section, CHANGELOG bump all landed via #875).

This PR is now scoped down to **just the reviewer-suggested wording improvements** for the Branch Protection section in `.github/instructions/cicd.instructions.md`.

## Why

The version that landed via #875 is correct but terse. The reviewer (Copilot reviewer on the original #874) suggested a 4-bullet structure with a new "How the name is derived" bullet that documents the job `name:` vs job-id fallback rule. This makes the diagnostic process explicit for any future contributor who hits the same "Expected - Waiting for status to be reported" symptom we hit on PR #860 today.

Also tightens the gotcha bullet: previous wording said the context must "match" the check-run name (room for ambiguity). New wording says it must BE the literal check-run name and explicitly calls out that `Merge Gate / gate` is a UI rendering, not a value to store.

## Diff

One file, 4 insertions / 3 deletions in `cicd.instructions.md`.

### DIFF
--- a/.github/instructions/cicd.instructions.md
+++ b/.github/instructions/cicd.instructions.md
@@ -88,9 +88,10 @@ integration suite runs only at merge time via GitHub Merge Queue
 - **Cross-workflow artifacts**: ci-integration.yml builds the binary inline (no cross-workflow artifact transfer); build-release.yml jobs share artifacts within the same workflow run.
 
 ## Branch Protection & Required Checks
-- **Single required check**: branch protection (`main-protection` ruleset id 9294522) requires only one status check: `gate` from `merge-gate.yml`. All other PR-time signals are aggregated by that workflow's poll loop.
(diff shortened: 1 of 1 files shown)


danielmeppiel:
Addressed both review comments in fef49e5a:

1. **CHANGELOG**: appended `(#874)` per Keep a Changelog convention
2. **`merge-gate.yml` self-contradiction**: the header comment said branch protection requires `Merge Gate / gate` (the broken display string). Now correctly says `gate` (the literal check-run name) with an inline reminder that the UI string is NOT the value to store

Also adopted the reviewer's clearer wording for the "Branch Protection & Required Checks" section, including the new "How the name is derived" bullet that documents the job `name:` vs job-id fallback rule.

#875 (stub removal) has been rebased on top.

Issue #684, opened 2026-04-12
audit --ci: verify deployed file content, not just existence


srid:
# `apm audit --ci` should verify file *content*, not just existence

## Problem

`apm audit --ci`'s baseline `deployed-files-present` check (`ci_checks.py:106-137`) only verifies that each file in the lockfile's `deployed_files` **exists on disk** — it does not check whether the content matches what `apm install` would produce. This means:

- A locally edited `.claude/rules/workflow.md` passes the check as long as the file exists
- An upstream source change in a remote package is invisible if the local copy was never reinstalled

The **policy-layer** `unmanaged-files` check (`policy_checks.py:606-690`, requires `--policy`) covers the adjacent problem of orphan detection — files on disk that aren't in the lockfile. But neither layer verifies content fidelity.

The only reliable way to detect content drift today is to run `apm install` and compare. But running it on the **live tree** briefly destroys the vendored directory (if doing a clean reinstall to catch orphans from packages that changed shape). We hit this in [juspay/kolu#468](https://github.com/juspay/kolu/issues/468): CI ran `apm install` on the live `.claude/` while a Claude Code session was active in the same worktree. Claude Code's stop hook crashed because `.claude/hooks/` didn't exist mid-reinstall.

## What we had to build instead

We built a scratch-dir verification ([juspay/kolu#469](https://github.com/juspay/kolu/pull/469)): symlink inputs into a temp dir, run `apm install` there, `diff -r` against the live tree. This works but required working around several behaviors:

### 1. No `--root` / `--target-dir` flag

`apm install` always writes relative to `$PWD`. To install into a scratch directory, we had to `cd` into it and stage all inputs (symlinks for `apm.yml` + local packages, `cp` for `apm.lock.yaml`). A `--root <dir>` flag would make this trivial.

### 2. `-t claude` doesn't match auto-detection behavior

When we used `-t claude` to force the target in the scratch dir, **local package integration silently produced no output** — rules and commands from the local `./agents` package were not deployed. But when we created an empty `.claude/` directory in scratch (triggering auto-detection via `target_detection.py:82`), the same local package correctly deployed 10 rules and 1 command.

This means `-t claude` and "auto-detect claude from `.claude/` existing" are not equivalent code paths.

### 3. `apm install` writes `generated_at` into lockfile on every run

[...]

PR #889, opened 2026-04-24
feat(audit): close audit-blindness gap for local .apm/ content (#887)

### LINKED ISSUE #887: feat: close audit-blindness gap for local .apm/ content via virtual self-entry + includes: manifest field
## TL;DR

`apm audit --ci` is structurally blind to locally-deployed `.apm/` content. This repo (`microsoft/apm`) ships 28 files into `.github/`, `.claude/`, and other workspace targets through implicit root-`.apm/` integration -- and **zero** of them are covered by audit, content-integrity verification, or content scanning. The lockfile records per-file SHA-256 hashes that are never re-verified. Reported by an external CISO, surfaced by an external user noting the canonical repo's empty manifest, confirmed by panel review.

This issue closes the gap end-to-end in one PR, shipped in the next minor release (v0.10.0). No phasing, no deprecation curve, no breaking change.

## Problem statement

APM has two parallel mechanisms for "local content" that produce identical artifacts (deployed files + per-file hashes) but land in two incompatible lockfile representations:

| Aspect | Mechanism A (declared local-path deps) | Mechanism B (implicit root `.apm/`) |
|---|---|---|
| Trigger | `apm.yml` declares `- ./packages/foo` | `<root>/.apm/` exists, no manifest entry |
| Lockfile home | `lock.dependencies[key]` -- full `LockedDependency` | `lock.local_deployed_files` (flat `List[str]`) + `lock.local_deployed_file_hashes` (flat `Dict[str, str]`) |
| Audit visibility | Full -- every check iterates `lock.dependencies` | **Zero** -- no consumer reads `local_deployed_files` |

Mechanism B was introduced in PRs #626, #644, #715 to remove stub-file friction for projects whose root `.apm/` IS the package. The DX win is real and must be preserved. The audit blindness was an unintended side effect of the data-model split.

### Verified blind spots (3 surfaces)

1. **`src/apm_cli/policy/ci_checks.py`** -- every check (lockfile-exists, deployed-files-present L106-137, no-orphans, ref-consistency, content-integrity L222-251, config-consistency) iterates only `lock.dependencies.items()`. Zero references to `local_deployed_files`.

2. **`src/apm_cli/policy/policy_checks.py:628-636`** -- `unmanaged-files` check has identical blindness; would falsely flag mechanism B files as unmanaged if enabled.

3. **`src/apm_cli/security/file_scanner.py:62-66`** -- `scan_lockfile_packages` has the same blindness. The content most consumed by AI agents (instructions, agents, skills) is the content never scanned.

### "Security theater" in the lockfile

`src/apm_cli/install/phases/post_deps_local.py:113-114` writes per-file SHA-256 hashes into `lock.local_deployed_file_hashes` on every install. **No code path re-reads them to compare against live files.** A developer can hand-edit any deployed instruction file post-install and `apm audit --ci` will not detect it. Hashes computed but never verified is bookkeeping presented as integrity verification.

### Bundle leak risk

`src/apm_cli/bundle/packer.py:122-126` (apm format) iterates `lockfile.get_all_dependencies()` with no `is_dev` filtering. `src/apm_cli/install/exporters/plugin_exporter.py:471-476` (plugin format) DOES filter `is_dev: true`. Currently safe by accident because `local_deployed_files` lives on a separate field; once we synthesize a self-entry into `lock.dependencies` (the fix below), the apm format packer needs an explicit guard to avoid leaking root content into bundles.

### Evidence in this repo

```bash
$ grep -c 'name:' apm.yml
1
$ python -c "import yaml; print(len(yaml.safe_load(open('apm.lock'))['local_deployed_files']))"
28
$ apm audit --ci
[+] lockfile-exists | No dependencies declared -- lockfile not required
[+] deployed-files-present | All deployed files present
[...all green, 28 files invisible...]
```

The CISO's complaint is correct. The external user's complaint is correct. We have an audit hole AND a credibility-gap manifest in our own canonical repo.

## Why this matters (product reasoning)

APM's positioning is **"the governance layer for AI agent configuration -- complete audit trail for everything AI agents see."** The MANIFESTO calls out *"Reliable over Magic"* and *"Transparent processes that teams can audit, understand, and trust."*

A canonical repo that deploys 28 files with no manifest declaration and no audit coverage violates this on three counts:

1. **The audit story is hollow.** We tell enterprises "run `apm audit --ci` in CI to detect drift" while the data structure those checks iterate is empty for our most common deployment shape (single-package projects).
2. **The manifest doesn't reflect intent.** A CISO reading our `apm.yml` in a PR diff sees nothing. Then they look at `.github/agents/` and see 28 files appearing from nowhere. Trust is broken.
3. **No incumbent has a unified audit story for AI tooling.** This is APM's moat. A false version of the claim is worse than no claim.

The fix below makes the claim true.

### Engine: virtual self-entry at lockfile read boundary

`LockFile.from_yaml()` synthesizes a virtual `LockedDependency` from existing `local_deployed_files` data and injects it into `self.dependencies`. `to_yaml()` extracts it back out before serialization to keep the YAML format stable. Every existing audit check, content scanner, and packer filter gains coverage through existing `lock.dependencies` iteration.

```python
# src/apm_cli/deps/lockfile.py -- inside from_yaml(), after existing loads:
_SELF_KEY = "."
if lock.local_deployed_files:
    lock.dependencies[_SELF_KEY] = LockedDependency(
        repo_url="<self>",
        source="local",
        local_path=".",
        is_dev=True,
        deployed_files=list(lock.local_deployed_files),
        deployed_file_hashes=dict(lock.local_deployed_file_hashes),
    )

# Inside to_yaml() -- extract before serialization:
_self_dep = self.dependencies.pop(_SELF_KEY, None)
data["dependencies"] = [dep.to_dict() for dep in self.get_all_dependencies()]
if _self_dep:
    self.dependencies[_SELF_KEY] = _self_dep  # restore in-memory
    data["local_deployed_files"] = sorted(_self_dep.deployed_files)
    if _self_dep.deployed_file_hashes:
        data["local_deployed_file_hashes"] = dict(sorted(_self_dep.deployed_file_hashes.items()))
```

Critical invariants on the synthesized entry:
- `is_dev: true` -- prevents plugin bundle leakage (plugin_exporter.py already filters this).
- `source: "local"` -- the apm-format packer needs a 2-line guard to skip `source: "local"` entries.
- `local_path: "."` -- canonical key; downstream code can disambiguate.

### Manifest: `includes:` field

New top-level optional manifest field. Default is `auto` (preserves the implicit `.apm/` discovery DX win from #626/#644/#715). Explicit path lists supported for organizations that need governance.

```yaml
# Default behavior -- auto-discover root .apm/ content.
includes: auto

# OR explicit governance:
includes:
  - .apm/agents/code-reviewer.agent.md
  - .apm/skills/apm-review-panel/SKILL.md
  - .apm/instructions/python.instructions.md
```

`auto` is **never deprecated.** It is the right default for the 95% case. Enterprise strictness is layered on via policy (next item), not forced on every project.

### Policy: `requires_explicit_includes` knob

`apm-policy.yml` schema gains a boolean knob:

```yaml
# apm-policy.yml
require:
  explicit_includes: true   # mandates includes: <list>, rejects includes: auto
```

Organizations needing CISO-grade governance set this in their org policy chain. Everyone else keeps `auto`.

### Audit: hash verification + content scanning + manifest consent

1. **`_check_content_integrity`** in `ci_checks.py` -- now actually re-reads files and compares against `deployed_file_hashes`. Currently the hashes exist but no check verifies them.
2. **`_check_no_orphans`** -- exclude the synthetic `.` key from manifest comparison: `if dep_key == ".": continue`.
3. **`_check_lockfile_exists`** -- treat non-empty `local_deployed_files` as `has_deps=True` so the aggregate runner doesn't short-circuit on local-only repos.
4. **New check `_check_includes_consent`** -- if `includes:` is missing from manifest AND `local_deployed_files` is non-empty, emit a `[!]` warning advising the user to declare `includes: auto`. This is the explicit-consent surface for enterprise governance without forcing migration.
5. **`scan_lockfile_packages`** in `file_scanner.py` -- gains coverage automatically through the synthesized entry; add test coverage to lock in the behavior.

### Packer: `source: "local"` guard

`src/apm_cli/bundle/packer.py` (around L122-126) -- skip entries with `source == "local"` in the apm-format packer, mirroring the existing rejection of local-path manifest deps at L89-97. Plugin format already correct via `is_dev` filter; add regression test.

### Canonical repo: update `microsoft/apm` apm.yml

```yaml
name: apm
version: 0.10.0
description: APM (Agent Package Manager) -- ship and govern AI agent context
author: Microsoft
license: MIT

# Local .apm/ content is auto-discovered and integrated into workspace
# targets (.github/, .claude/, etc.) on `apm install`. Declare 'auto' to
# opt in to implicit discovery; list paths for explicit governance.
includes: auto

dependencies:
  apm: []
  mcp: []

scripts: {}
```

This closes the external-user complaint and serves as the canonical example.

## Acceptance criteria

All in one PR. All in v0.10.0.

### Code

- [ ] `LockFile.from_yaml()` synthesizes virtual self-entry when `local_deployed_files` is non-empty (`is_dev: true`, `source: "local"`, `local_path: "."`, `repo_url: "<self>"`).
- [ ] `LockFile.to_yaml()` extracts the self-entry before serialization; YAML format is byte-stable for existing lockfiles (round-trip test).
- [ ] `_check_lockfile_exists` treats non-empty `local_deployed_files` as `has_deps`.
- [ ] `_check_no_orphans` excludes the synthetic `.` key.
- [ ] `_check_content_integrity` re-reads files and verifies hashes against `deployed_file_hashes` (currently does not verify).
- [ ] New `_check_includes_consent` warning when `includes:` missing and `local_deployed_files` non-empty.
- [ ] `scan_lockfile_packages` covers the self-entry (verified by test, not just by inheritance).
- [ ] `apm.yml` schema supports `includes: auto` and `includes: [<path>, ...]` (explicit list).
- [ ] `apm-policy.yml` schema supports `require.explicit_includes: bool`; policy gate rejects `includes: auto` when set.
- [ ] `bundle/packer.py` apm-format packer skips `source: "local"` entries.
- [ ] `microsoft/apm` `apm.yml` updated with `includes: auto` (separate commit in same PR).

### Tests

- [ ] Lockfile round-trip preserves YAML format for existing mechanism-B lockfiles.
- [ ] `apm audit --ci` on this repo (post-fix) surfaces all 28 locally-deployed files.
- [ ] Hash mismatch test: hand-edit a deployed file, audit detects drift.
- [ ] `apm pack` (apm format) does NOT include self-entry files in the bundle.
- [ ] `apm pack --format plugin` does NOT include self-entry files in the bundle.
- [ ] `includes: auto` parses; `includes: [<paths>]` parses; `includes:` missing emits consent warning.
- [ ] Policy `require.explicit_includes: true` blocks install when manifest has `includes: auto` or no `includes:`.
- [ ] Self-entry NOT exposed via `manifest.get_apm_dependencies()` (the synthesis is a lockfile concern, not a manifest concern).

### Docs

- [ ] `docs/src/content/docs/reference/lockfile-spec.md` -- document self-entry convention (the `.` key, `<self>` repo_url, `is_dev: true` invariant).
- [ ] `docs/src/content/docs/reference/manifest-spec.md` (or equivalent) -- document `includes:` field with `auto` and explicit-list forms.
- [ ] `docs/src/content/docs/governance-guide.md` `§14 Known gaps and limitations` -- remove the "no drift detection on deployed files" entry; it's no longer true.
- [ ] Policy reference docs -- document `require.explicit_includes`.
- [ ] Skill resource files -- update `packages/apm-guide/.apm/skills/apm-usage/governance.md` and any others affected by the schema change.

### CHANGELOG

```markdown
### Added
- New `includes:` manifest field. Declares how local `.apm/` content is
  discovered and integrated. `auto` (default) preserves existing implicit
  discovery; explicit path lists enable enterprise governance. Policy gains
  `require.explicit_includes` to mandate explicit declarations org-wide.
- `apm audit --ci` now verifies SHA-256 hashes of locally-deployed files
  against the lockfile, detecting post-install drift.

### Fixed
- `apm audit --ci` now covers locally-deployed `.apm/` content (agents,
  skills, instructions, hooks, commands). Previously, files deployed from
  root `.apm/` were tracked in the lockfile but invisible to all audit
  checks and the content-integrity scanner. Lockfile schema is unchanged;
  coverage is gained through internal read-boundary normalization.
- `apm pack` (apm format) no longer risks including locally-deployed root
  `.apm/` content in published bundles. Plugin format was already safe via
  `is_dev` filtering; both formats now have explicit test coverage.
```

Migration line: **None.** `auto` is the default. Existing projects work without changes. Audit may now correctly flag drift that was previously invisible -- this is a fix, not a break.

## Out of scope (intentionally)

- **Deprecating `auto`.** Never. It's the right default. Enterprise strictness layers via policy.
- **New manifest surface beyond `includes:`.** No `selfDependencies:`, no `provides:`, no `primitives:`. The architectural panel debated and rejected these.
- **Touching `_integrate_root_project()` or the install pipeline phase ordering.** The fix is purely at the lockfile read/write boundary plus three audit-check tweaks. Install behavior is byte-identical.
- **Migrating mechanism A (declared local-path deps).** Already audited; not affected.

## Naming rationale

`includes:` was chosen after rejecting:
- `primitives:` -- internal jargon, no external developer knows the term.
- `provides:` -- implies publishing semantics; this is about local integration.
- `agents:` -- too narrow; field covers instructions, skills, hooks, commands, agents.
- `bundles:` -- collides with `apm pack` bundle terminology.
- `local:` -- vague.
- `selfDependencies:` -- conflates "I ship X" with "I depend on X."

`includes:` is familiar from CSS `@include`, C `#include`, cargo `include = [...]`. Reads naturally: *"This project includes these local AI capabilities."*

## Decision provenance

Crisis panel review (2026-04-23) with three independent specialist assessments:

- **Python Architect**: diagnosed dual-track as accidental structural duplication; recommended virtual self-entry at lockfile read boundary (~35 lines, 2 files); rejected new manifest surface as DX regression.
- **Supply-Chain Security Expert**: confirmed "security theater" (hashes computed but never verified); demanded six-item GA bar; required `is_dev: true` on self-entry to prevent bundle leakage.
- **DevX UX Expert**: tested three personas (npm/pip user, CISO, plugin author); argued implicit detection fails the CISO sniff test; proposed new manifest field with phased deprecation.
- **OSS Growth Hacker**: side-channelled "not either/or" reframe -- engine + story are separable concerns.
- **APM CEO**: arbitrated -- UX wins the principle (manifest declaration matters), Architect wins the default (preserve `.apm/` DX), Security wins the GA bar (hard blockers ratified). Single-release scope per maintainer direction.

## File map for the implementer

| File | Change |
|---|---|
| `src/apm_cli/deps/lockfile.py` | Self-entry synthesis in `from_yaml()`; extraction in `to_yaml()`. |
| `src/apm_cli/policy/ci_checks.py` | Fix `_check_lockfile_exists` gating; exclude `.` from `_check_no_orphans`; add hash verification to `_check_content_integrity`; add `_check_includes_consent`. |
| `src/apm_cli/manifest/schema.py` (or equivalent) | Add `includes:` field (auto / list-of-paths). |
| `src/apm_cli/policy/policy_schema.py` (or equivalent) | Add `require.explicit_includes: bool`. |
| `src/apm_cli/install/phases/policy_gate.py` | Wire `explicit_includes` enforcement. |
| `src/apm_cli/bundle/packer.py` | Skip `source: "local"` entries in apm-format packer (~L122-126). |
| `src/apm_cli/security/file_scanner.py` | No code change required; coverage gained via synthesis. Add test. |
| `apm.yml` | Add `includes: auto`; bump version to 0.10.0. |
| `docs/src/content/docs/reference/lockfile-spec.md` | Document `.` self-entry convention. |
| `docs/src/content/docs/reference/manifest-spec.md` | Document `includes:` field. |
| `docs/src/content/docs/governance-guide.md` | Remove `§14` drift-detection gap entry. |
| `packages/apm-guide/.apm/skills/apm-usage/governance.md` | Update for new schema. |
| `CHANGELOG.md` | Add entries above. |

## Risk assessment

- **Lockfile YAML stability**: Round-trip test required. Synthesis happens purely in-memory; YAML format unchanged.
- **`is_dev: true` correctness**: If wrong, plugin bundles leak root content. Test required.
- **Packer regression**: If `source: "local"` guard is missing, apm bundles leak root content. Test required.
- **Backward compat**: `auto` default means no existing project breaks. New audit findings are correct exposure of pre-existing drift, not a regression.

## Definition of done

A reviewer can say: *"This repo's `apm audit --ci` now detects drift on every file under `.github/`, `.claude/`, and any other workspace target. Hand-editing a deployed agent file is caught. The manifest declares `includes: auto`. The CISO can read `apm.yml` and trust it. Plugin authors keep zero-ceremony `.apm/` magic. Enterprise orgs can require explicit declarations via policy. CHANGELOG, docs, and skill resources reflect the new reality. All in one PR. Shipped in v0.10.0."*

danielmeppiel:
## What & Why

CISOs reported an **audit-blindness gap**: `apm install` was deploying local `.apm/` content into governed directories (`.github/`, `.claude/`, etc.), but `apm audit --ci` never hash-verified those deployed files. The lockfile recorded `local_deployed_files` and their hashes, yet the audit pipeline only walked package dependencies — so post-install tampering of any locally-sourced file went undetected.

This PR closes the gap with end-to-end hash verification of every managed file plus a new `includes:` manifest field that gives users (and enterprises) explicit governance consent over what local content gets deployed.

Closes #887.

## Solution overview

- **Self-entry synthesis at the lockfile read boundary** — local-content state is projected as a virtual `<self>` dependency keyed by `_SELF_KEY = "."`. The on-disk YAML schema is unchanged; `to_yaml()` does a round-trip pop+restore so the synthetic entry never persists.
- **New `includes:` manifest field** — three modes: `auto` (deploy all `.apm/`), an explicit list, or omitted (deploy nothing). Purely additive; existing manifests keep working.
- **Hash verification in the `content-integrity` audit check** — now covers the self-entry alongside package deps, detecting post-install tampering on **any** managed file.
- **New baseline check `includes-consent`** — advisory only, never blocks.
- **New policy `manifest.require_explicit_includes`** (`block` / `warn` / `off`) — enterprises can mandate explicit include lists.
- **`apm init -y` scaffolds `includes: auto`** — clean first-run UX.
- **Bundle stripping** — local-content fields are removed before bundle publish; no leakage into the marketplace.

## Critical bug fixes discovered during dogfood

These were caught while dogfooding the new audit on the APM repo itself and are essential to the fix:

- **`cleanup.py`** — orphan loop was deleting all 26 deployed files post-install because the synthesized self-entry was misclassified as an orphan package. Fixed by treating `_SELF_KEY` as protected.
- **`lockfile.py`** — `is_semantically_equivalent` skipped `local_deployed_file_hashes` comparison, so hash-only updates never persisted. Fixed by including the hashes in the equivalence check.
- **`lockfile_enrichment.py`** (bundle path) — local-content fields needed stripping after the YAML round-trip. Fixed.

## Audit coverage: before / after

```mermaid
flowchart LR
  subgraph BEFORE["BEFORE (audit blind to local .apm/)"]
    I1[apm install] --> S1[.apm/ source files]
    S1 --> D1[.github/ deployed]
    I1 --> L1[apm.lock.yaml<br/>local_deployed_files + hashes<br/>written but never read]
    A1[apm audit --ci] --> L1
    A1 -.never reads.-> D1
    D1 -.HASH DRIFT INVISIBLE.-> X1((CISO blind))
  end
```

```mermaid
flowchart LR
  subgraph AFTER["AFTER (every managed file verified)"]
    I2[apm install] --> S2[.apm/ source files]
    S2 --> D2[.github/ deployed]
    I2 --> L2[apm.lock.yaml<br/>local_deployed_files + hashes]
    L2 --> SE[from_yaml synthesizes<br/>self-entry under &lt;self&gt;]
    A2[apm audit --ci] --> L2
    SE --> HV[content-integrity:<br/>hash-verify every<br/>deployed file]
    HV -. detects .-> D2
    HV --> OK([7/7 green])
    HV --> DRIFT([hash-drift: file<br/>dep=&lt;self&gt;])
  end
```

## Verification

- **Unit suite**: `5383/5384` pass. The sole failure is the pre-existing flaky `test_user_scope_skips_workspace_runtimes`, unrelated to this change.
- **Live dogfood end-to-end**:
  - `apm install` clean, `apm audit --ci` returns 7/7 green
  - Introduced drift on a deployed file -> audit detects with `dep=<self>`
  - Restored file -> audit returns to 7/7 green
- **`apm init -y`** confirmed to scaffold `includes: auto`.
- Pre-existing dogfood drift on `auth-expert.agent.md` and `cicd.instructions.md` remains visible — these are real post-install hand-edits from prior PRs and are handled outside this PR.

## Panel reviews

All three specialist panels reviewed; CEO signed off SHIP.

- **Supply chain security** — PASS-WITH-NOTES. Path-validation guard added to the hash-verify loop.
- **CLI logging UX** — PASS-WITH-NOTES. Status-symbol contradiction fixed, remediation made conditional, hashes truncated for terminal width.
- **DevX UX** — PASS-WITH-NOTES. `init` scaffold, doc check counts updated (6+16 -> 7+17), error wording branched, `dep=<self>` label introduced, `cli-commands.md` updated.
- **APM CEO** — SHIP. *"Every managed file, verified"* — the audit-coverage milestone.

## Breaking changes

**None.** Purely additive. No migration required for any user class — existing manifests, lockfiles, and CI flows continue to work unchanged.

## Version

Bumps to **0.10.0** (minor: significant new functionality, no breakage).

## Test plan

```bash
uv run pytest tests/unit tests/test_console.py
uv run pytest tests/integration/test_local_content_audit.py   # new file: 5 subprocess tests A-E
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `includes:` manifest field (auto | list) for explicit governance of local `.apm/` content. Closes audit-blindness gap (#887).
+- `apm audit --ci` now verifies hash integrity of locally deployed files, detecting hand-edits and config drift. (#887)
+- `policy.manifest.require_explicit_includes` policy field enforces explicit `includes` lists (rejects `auto` + undeclared). (#887)
+- Includes-consent advisory in audit output: `[!]` warning when local content is deployed but `includes:` is not declared. (#887)
 - CI: add `APM Self-Check` to `ci.yml` for `apm audit --ci`, regeneration-drift validation, and `merge-gate.yml` `EXPECTED_CHECKS` coverage. (#885)
 
 ### Changed
 
+- Lockfile in-memory shape: a synthesized self-entry now appears in `LockFile.dependencies` for local content. The on-disk YAML format is unchanged (data still serialized as flat `local_deployed_files`/`local_deployed_file_hashes` fields). (#887)
 - CI: smoke tests in `build-release.yml`'s `build-and-test` job (Linux x86_64, Linux arm64, Windows) are now gated to promotion boundaries (tag/schedule/dispatch) instead of running on every push to main. Push-time smoke duplicated the merge-time smoke gate in `ci-integration.yml` and burned ~15 redundant codex-binary downloads/day. Tag-cut releases still run smoke as a pre-ship gate; nightly catches upstream codex URL drift; merge-time still gates merges into main. (#878)
 - CI docs: clarify that branch-protection ruleset must store the check-run name (`gate`), not the workflow display string (`Merge Gate / gate`); document the merge-gate aggregator in `cicd.instructions.md` and mark the legacy stub workflow as deprecated.
+
+### Fixed
+
+- Audit blindness for local `.apm/` content -- `apm audit --ci` now detects drift, missing files, and content tampering on locally-authored files (not just installed packages). (#887)
+- Packer leak risk: local-content fields (`local_deployed_files`, `local_deployed_file_hashes`) are now stripped from bundled lockfiles, preventing phantom self-entries on unpack. (#887)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `includes:` manifest field (auto | list) for explicit governance of local `.apm/` content. Closes audit-blindness gap (#887).
 - `apm audit --ci` now verifies hash integrity of locally deployed files, detecting hand-edits and config drift. (#887)
 - `policy.manifest.require_explicit_includes` policy field enforces explicit `includes` lists (rejects `auto` + undeclared). (#887)
-- Includes-consent advisory in audit output: `[!]` warning when local content is deployed but `includes:` is not declared. (#887)
+- `includes-consent` advisory appears in `apm audit` CLI/JSON output when local content is deployed without an explicit `includes:` declaration (#887)
 - CI: add `APM Self-Check` to `ci.yml` for `apm audit --ci`, regeneration-drift validation, and `merge-gate.yml` `EXPECTED_CHECKS` coverage. (#885)
 
 ### Changed
--- a/apm.lock.yaml
+++ b/apm.lock.yaml
@@ -31,7 +31,7 @@ local_deployed_files:
 local_deployed_file_hashes:
   .github/agents/agentic-workflows.agent.md: sha256:d1ea2d038e2af8be11d6c95b3213b03b9777fae46f0438efa95d5a803e6c3765
   .github/agents/apm-ceo.agent.md: sha256:dfc436e6eeffc7ec1c2f556edb78e4a5166ac36d162ea720d08b4b79af0a9938
-  .github/agents/auth-expert.agent.md: sha256:efdc8c7fd046409f4467ecf14da9f0d5f0e4a86372e5885c3763e89ff6f9ea69
+  .github/agents/auth-expert.agent.md: sha256:85409aab097cf239e5aa7ad61db6c4586be9884ef64a45fa9c894017b046b56b
   .github/agents/cli-logging-expert.agent.md: sha256:24bf6c4b420c42292700ad0eb80b53d275be5c9cb186d471d706211f8419e3b9
   .github/agents/devx-ux-expert.agent.md: sha256:3472680f43b2b4411b9437ec31529216afd4e576e1874c14430935e7f1ded1f2
   .github/agents/doc-analyser.agent.md: sha256:47b1d0204904b786c19d4fe84343e86cdab6f92f862f676ba741ffe6e1385679
@@ -40,7 +40,7 @@ local_deployed_file_hashes:
   .github/agents/python-architect.agent.md: sha256:4118e60a56d4b94183915a3650d2a4c3294d2b4a5daecbb547a731b04a679b54
   .github/agents/supply-chain-security-expert.agent.md: sha256:9a4e731b12e7658f71d54c22e90f80ce0c45e3eacbb069b8505ed96ec9e79ba5
   .github/instructions/changelog.instructions.md: sha256:1e51ec4c74e847967962bd279dc4c6e582c5d3578490b3c28d5f3acd3e05f73e
-  .github/instructions/cicd.instructions.md: sha256:170e6fa09bcf4064d33420ffca6b3125bf7011982c4c7a00320af71f2f6c6bf9
+  .github/instructions/cicd.instructions.md: sha256:9c0fafc74f743aa97e5adba2168d66c9e3a327b135065e3b804bdbb5f04cda5d
   .github/instructions/cli.instructions.md: sha256:8e39e8d5047ce88575cb02f87c2bcede584dfef258bd86f7466c7badf136541a
   .github/instructions/doc-sync.instructions.md: sha256:bb3816254f8df6bffc6faacd556871f36903e9d7f348982f1e2de0339384c696
   .github/instructions/encoding.instructions.md: sha256:93db7377dc896f6efecf2c5d8c5d89255a555562f468d034d64c42edd5cf46d5
--- a/apm.yml
+++ b/apm.yml
@@ -1,6 +1,7 @@
 name: apm
-version: 0.9.0
+version: 0.10.0
 description: APM (Agent Package Manager) -- ship and govern AI agent context
+includes: auto
 author: Microsoft
 license: MIT
 
--- a/docs/src/content/docs/enterprise/apm-policy.md
+++ b/docs/src/content/docs/enterprise/apm-policy.md
@@ -92,7 +92,7 @@ Policy is evaluated at two points. Both use the same policy file and the same me
 
 ### CI time (audit gate)
 
-`apm audit --ci --policy org` runs the same checks (plus 6 baseline lockfile checks) and is intended as a required status check on pull requests. It produces SARIF output that GitHub Code Scanning renders inline on the PR diff.
+`apm audit --ci --policy org` runs the same checks (plus 7 baseline lockfile checks) and is intended as a required status check on pull requests. It produces SARIF output that GitHub Code Scanning renders inline on the PR diff.
 
 For setup, see [CI Policy Enforcement](../../guides/ci-policy-setup/).
 
--- a/docs/src/content/docs/enterprise/governance-guide.md
+++ b/docs/src/content/docs/enterprise/governance-guide.md
@@ -85,12 +85,15 @@ The scope matrix below is the contract. Every row maps a security or operational
 | Source attribution | `compilation.source_attribution` | `source-attribution` | `[i] audit-only` | Yes |
 | Required manifest fields | `manifest.required_fields` | `required-manifest-fields` | `[i] audit-only` | Yes |
 | Manifest scripts policy | `manifest.scripts` | `scripts-policy` | `[i] audit-only` | Yes |
+| Explicit manifest includes | `manifest.require_explicit_includes` | `explicit-includes` | `[i] audit-only` | Yes |
 | Unmanaged files in governed dirs | `unmanaged_files.action`, `.directories` | `unmanaged-files` | `[i] audit-only` | Yes |
 | Cache TTL override | `policy.cache.ttl` | -- | `[!] parsed but not enforced` (cache reader uses hardcoded 1h) | -- |
 | Transitive MCP trust (policy field) | `mcp.trust_transitive` | -- | `[!] parsed but not enforced` (gate is the `--trust-transitive-mcp` CLI flag) | -- |
 | Manifest content types | `manifest.content_types` | -- | `[!] parsed but not enforced` | -- |
 
 The full schema and the canonical 6+16 check enumeration live in the [Policy Reference](../policy-reference/). The 6 baseline lockfile checks (lockfile presence, ref consistency, deployed files present, no orphaned packages, MCP config consistency, content integrity) run on every `apm audit --ci` regardless of policy and are non-bypassable -- they are covered in section 7.
+
+`manifest.require_explicit_includes` (`bool`, default `false`) deserves a callout: when set to `true`, the `explicit-includes` check rejects any `apm.yml` that omits `includes:` or sets `includes: auto`. Use this when every published local file must be enumerated in the manifest and reviewable in PR diffs. See the [`includes` field](../../reference/manifest-schema/#39-includes) for the three accepted forms.
 
 ---
 
@@ -104,7 +107,6 @@ Be clear with stakeholders about what is out of scope today:
 - **File integration paths.** Where files land on disk inside the repo is decided by the integrators in each APM package. Policy cannot rewrite a package's file layout.
 - **Custom agent tools beyond MCP.** If an agent stack ships its own non-MCP tool plugins, they sit outside policy scope. Govern them indirectly through `dependencies.allow`/`deny` and `unmanaged_files`.
 - **Token scopes and OAuth scopes.** APM does not audit the scope of the GitHub PAT or app token used to fetch policies and packages. Manage that through the standard GitHub controls on the token issuer.
-- **Drift on deployed files after install.** `deployed_file_hashes` are recorded in the lockfile but never re-verified. A developer who hand-edits a deployed instruction file post-install will not be detected by `apm install` or `apm audit --ci`. See section 14.
 - **Anything `apm compile` or `apm run` does.** Those commands trust whatever `apm install` placed on disk. They do not re-check policy.
 
 For the underlying threat model (what the content scanner protects against, MCP trust boundary, dependency provenance), see the [Security Model](../security/).
@@ -275,7 +277,7 @@ This is the certitude section. Read it twice if you are deciding whether `apm au
 | `apm install --no-policy` | All 16 policy checks at install (incl. transitive MCP, hash pin) | The 6 baseline checks in `apm audit --ci` | git diff of `apm.lock.yaml` in PR |
 | `APM_POLICY_DISABLE=1` env | Same as `--no-policy` plus the 16 audit policy checks | The 6 baseline checks in `apm audit --ci` | PR diff; CI env vars in Actions logs |
 | Manual edit to `apm.lock.yaml` | Nothing; install regenerates the file each run | Audit baseline `ref-consistency` and `deployed-files-present` | git diff |
-| Manual edit to deployed file post-install | Content-equality drift (no re-hash) | Hidden-Unicode scan in `apm audit` content mode | git diff of the deployed file in PR |
+| Manual edit to deployed file post-install | Local file content until next audit | Audit baseline `content-integrity` (re-hashes deployed files); hidden-Unicode scan in `apm audit` content mode | git diff of the deployed file in PR |
 | Direct `git clone` of an APM package, bypassing install | Everything; nothing detects out-of-band file drops | Audit baseline `no-orphaned-packages` and audit-only `unmanaged-files` | git diff |
 | Fork repo to a personal org | Org policy auto-discovery (resolves to fork's `.github`) | Whatever your CI requires on the canonical repo | branch protection on canonical repo |
 | `--trust-transitive-mcp` CLI flag | The transitive MCP preflight (second pass) | Direct MCP preflight; baseline content scan; audit MCP checks | CI command lines and Actions logs |
@@ -312,7 +314,7 @@ When `apm install` returns exit code 0 and the effective org policy is in `enfor
 
 You are NOT guaranteed:
 
-- That files on disk are still what install wrote. There is no drift detection on `deployed_file_hashes`.
+- That files on disk are still what install wrote. `apm install` itself does not re-verify deployed files; drift is caught by `apm audit --ci` baseline `content-integrity` (re-hashes against `deployed_file_hashes`).
 - That prompt or instruction content is semantically safe. Only the hidden-Unicode scan runs.
 - That the audit-only checks (`compilation-strategy`, `source-attribution`, `required-manifest-fields`, `scripts-policy`, `unmanaged-files`) passed. Run `apm audit --ci --policy <scope>` in CI for those.
 - That non-APM files in the repo conform to anything. APM only governs files it placed.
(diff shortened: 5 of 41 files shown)


danielmeppiel:
## Note on follow-up

This PR is the foundation that Epic #898 builds on. The Epic-PR will rebase on top of the work landed here and extend it with:

- per-asset enumeration inside bundle directories (the gap exposed by PR #882)
- skill-as-package model (#896)
- `apm audit --drift` user-facing alias
- dependency-closure policy rule
- self-host CI dogfood on `microsoft/apm`

No scope change requested here; this PR ships as designed and closes #684 + #887. The Epic owns the next layer.

danielmeppiel:
## Found issue: shared/apm.md restore step crashes on PRs that modify `apm.lock.yaml`

Tracking issue: #901

### What's failing here

The `pr-review-panel` agent run on this PR aborts before the agent even starts: https://github.com/microsoft/apm/actions/runs/24883083247/job/72856352586?pr=889

```
error: Your local changes to the following files would be overwritten by checkout:
        apm.lock.yaml
ERR_API: Failed to checkout PR branch: The process '/usr/bin/git' failed with exit code 1
```

This isn't specific to PR #889 - it will happen on any PR that touches `apm.lock.yaml`.

### What's actually going on (plain English)

`shared/apm.md` (our recommended gh-aw integration template) runs in two phases on a PR:

1. **Trusted base phase**: checks out `main`, downloads the APM bundle, calls `microsoft/apm-action` in **restore mode** to deploy primitives into the workspace. Restore mode also writes `apm.lock.yaml` back into the workspace as a side-effect (even though `apm unpack` documents the lockfile as metadata-only).
2. **PR checkout phase**: gh-aw runs `git checkout -B <branch> origin/pr-head`. Git refuses because the workspace is dirty.

When the PR also modifies `apm.lock.yaml` (which is exactly what every dependency-related PR does), the two writes collide and checkout aborts.

### Why this matters beyond PR #889

- It triggers on the **canonical first-run PR** for new APM adopters: "I added a dependency, please review." That's the worst possible adopter to break.
- The error names neither APM nor `microsoft/apm-action`, so consumers debug gh-aw or assume their checkout is broken.
- `shared/apm.md` ships from microsoft/apm but executes in every external repo that imports it. It's effectively a public API surface for APM's CI integration - regressions have cross-repo blast radius.
- The crash silently masks a separate higher-severity finding (PR-controlled primitive shadowing) that we'll file against gh-aw once this is fixed - see #901's "Out of scope follow-ups".

### Chosen fix

Append one step to `shared/apm.md` after "Restore APM packages":

```yaml
- name: Reset tracked files after APM restore
  run: git checkout -- . 2>/dev/null || true
```

Then `gh aw compile` and commit the regenerated `.lock.yml` files.

### Why this approach

- **Untracked primitives survive**: bundle-deployed files in `.github/{skills,agents,instructions,prompts}/` are untracked in consumer repos, so they survive both `git checkout -- .` and the subsequent PR checkout. Agent discovery keeps working unchanged.
- **Idempotent and fail-safe**: no-op when clean, `|| true` swallows non-zero in non-git contexts, order-insensitive.
- **Whole-tree, not file-list**: hard-coding `apm.lock.yaml apm.yml` would silently regress when future bundle paths overlap tracked files (e.g., `.github/agents/*.agent.md`). Safe here because no prior gh-aw pre-step writes tracked workspace files.
- **Zero new inputs, zero new env vars** for adopters; mental model unchanged; no auth surface.

Alternatives considered:
- **Drop the lockfile write inside `microsoft/apm-action`** - the right durable fix, but a breaking change to the action's contract. Held for the action's next major.
- **Relocate restore to `/tmp/...` and copy primitives in after PR checkout** - breaks agent primitive discovery, no gh-aw post-checkout pre-step hook, credential-surface risk.

Full rationale, required actions before merge, verification steps, and out-of-scope follow-ups (apm-action vNext, parallel diagnostic warning PR, gh-aw primitive-shadowing escalation) are in #901.

danielmeppiel:
**Update on the fix layering**: after re-examining the architecture, the bridge-step approach in gh-aw was the wrong layer. We control both `apm unpack` and `microsoft/apm-action`, so the fix belongs there. New plan:

1. **microsoft/apm#901** (rewritten): `apm unpack` stops writing `apm.lock.yaml` / `apm.yml` to the output directory (aligns implementation with the CLI's own documented metadata-only contract). Root cause.
2. **microsoft/apm-action#26**: v2 of the action builds on the fixed CLI; `working-directory` stops getting dirtied on restore. Every action consumer (gh-aw users, raw GH Actions users, anything) gets the fix transparently with zero opt-in.
3. **github/gh-aw#28256** (closed as redirect): `shared/apm.md` just bumps to `microsoft/apm-action@v2` once available. One-line change. No template gymnastics.

The single-line `git checkout -- .` workaround in microsoft/apm's own copy of `shared/apm.md` stays in place until step 2 ships, then it becomes a documented no-op and we drop it.

danielmeppiel:
## Final diagnosis (corrected)

Verified -- the bug is **isolated to `microsoft/apm-action`'s restore mode**. The `apm unpack` CLI is correct.

### What's actually happening

1. `microsoft/apm-action` restore mode (`runner.ts:90`) deliberately **skips installing the `apm` CLI** for speed.
2. Therefore `extractBundle` (`bundler.ts:62-89`) always falls through to its `tar xzf --strip-components=1` fallback.
3. Raw `tar` dumps the **entire** bundle into `working-directory` -- including `apm.lock.yaml` and `apm.yml`, which `apm unpack` would have correctly excluded.
4. When `working-directory` defaults to `${{ github.workspace }}` (a git checkout), those tracked files become dirty.
5. The next step (`checkout_pr_branch.cjs`) tries `git checkout` and aborts:
   ```
   error: Your local changes to the following files would be overwritten by checkout:
           apm.lock.yaml
   ERR_API: Failed to checkout PR branch
   ```
6. This breaks every `pull_request_target` agentic workflow whose triggering PR modifies `apm.lock.yaml` -- the canonical first-run PR for any new APM adopter.

### Single-layer fix

**microsoft/apm-action#26** -- ship as v1.5: install the `apm` CLI in restore mode (drop the "skip install" optimization), take the verified `apm unpack` path. Tiny diff in one file (`runner.ts`), no API change, no breaking change. The CLI and gh-aw template need no edits.

### Tactical mitigation while v1.5 ships

A small PR will add a post-restore `git checkout -- . 2>/dev/null || true` in microsoft/apm's own copy of `shared/apm.md`. It becomes a documented no-op once we bump to `microsoft/apm-action@v1.5`.

### What changed vs. the prior comment

- **microsoft/apm#901 closed** -- CLI contract is honored; no CLI fix needed.
- **github/gh-aw#28256 closed** -- gh-aw is the symptom layer; nothing to do upstream.
- **microsoft/apm-action#26 retitled / rewritten** -- now the single source of truth for the fix.

References:
- Failing run: https://github.com/microsoft/apm/actions/runs/24883083247/job/72856352586?pr=889
- Action fix: https://github.com/microsoft/apm-action/issues/26
- CLI verification: `src/apm_cli/bundle/unpacker.py` (only writes `deployed_files`)
- Action code: `microsoft/apm-action` `src/runner.ts:90` (skip-install) and `src/bundler.ts:62-89` (tar fallback)

danielmeppiel:
Tracking the follow-up cleanup in #902 — bump `shared/apm.md` to `microsoft/apm-action@v1.4.2` now that the upstream fix has shipped (release: https://github.com/microsoft/apm-action/releases/tag/v1.4.2).

PR #895, opened 2026-04-24
fix(policy): resolve project_root before path-traversal check

### LINKED ISSUE #886: Windows: _get_cache_dir flakes on tmpdir 8.3 short names (PathTraversalError)
## Symptom

`tests/unit/policy/test_cache_atomicity.py::TestCacheAtomicity::test_concurrent_writers_no_torn_files` fails intermittently on the Windows unit-test job in `build-release.yml` with:

```
apm_cli.utils.path_security.PathTraversalError: Policy cache path 
'C:\Users\RUNNER~1\AppData\Local\Temp\tmpu3aaqijk\apm_modules\.policy-cache' 
resolves outside project root 
'C:\Users\RUNNER~1\AppData\Local\Temp\tmpu3aaqijk' 
-- refusing to read or write the cache here.
```

First observed in [run 24861961730](https://github.com/microsoft/apm/actions/runs/24861961730/job/72789349782) (post-merge, commit `8665f4b4`, PR #885). Linux + macOS jobs are unaffected.

## Root cause

In `src/apm_cli/policy/discovery.py::_get_cache_dir` (introduced in #832), the path-security guard does:

```python
def _get_cache_dir(project_root: Path) -> Path:
    base = project_root / "apm_modules"
    candidate = base / POLICY_CACHE_DIR
    try:
        ensure_path_within(candidate, project_root)
    except PathTraversalError:
        raise PathTraversalError(...)
```

`ensure_path_within` resolves the **candidate** path through `Path.resolve()` (which on Windows expands 8.3 short names like `RUNNER~1` to the long form `runneradmin` and follows symlinks), but the **`project_root`** argument is compared as-passed.

When the test calls it with `tempfile.mkdtemp()`, Windows returns the short-name form (`C:\Users\RUNNER~1\...`). The candidate resolves to `C:\Users\runneradmin\...\apm_modules\.policy-cache`, the project root stays `C:\Users\RUNNER~1\...`, the prefix check fails, and `PathTraversalError` is raised.

The test is concurrent (`ThreadPoolExecutor`), which is why it manifests racily -- depending on which thread wins to resolve first / on the runner's filesystem state, sometimes one or more threads see the short-name path and trip the guard.

## Why now

The bug has been latent since #832 merged. It surfaces only when:

1. The Windows runner's `tempfile.mkdtemp()` returns a path with a `~1` short-name component (depends on user profile name length on the specific runner image).
2. The concurrent test path `_write_cache -> _get_cache_dir` runs.

Recent main runs that ran the same Windows job have been mostly green by luck.

## Suggested fix

Resolve `project_root` once at the top of `_get_cache_dir` before passing it to `ensure_path_within`, so both sides are normalized:

```python
def _get_cache_dir(project_root: Path) -> Path:
    project_root = project_root.resolve()
    base = project_root / "apm_modules"
    candidate = base / POLICY_CACHE_DIR
    try:
        ensure_path_within(candidate, project_root)
    except PathTraversalError:
        raise PathTraversalError(...)
    ...
```

Optionally, the same defensive normalization belongs inside `ensure_path_within` itself in `src/apm_cli/utils/path_security.py` so future callers can't repeat the mistake -- but the local fix is sufficient to unblock the test.

## How to verify

Add a regression test that constructs a `project_root` with an unresolved component (e.g. via `tempfile.mkdtemp()` on Windows, or a deliberately-symlinked tmp dir on POSIX) and asserts `_get_cache_dir` does not raise.

Run the existing concurrent test under stress (`pytest --count=20 tests/unit/policy/test_cache_atomicity.py` with `pytest-repeat`) on Windows.

## Scope notes

- Touch only `src/apm_cli/policy/discovery.py::_get_cache_dir` (and possibly `path_security.py` if going for the defensive fix).
- Add a focused regression test, not a refactor of the cache layer.
- No behavior change on Linux / macOS.

## Refs

- Introduced in: #832
- First observed: [run 24861961730](https://github.com/microsoft/apm/actions/runs/24861961730/job/72789349782) (commit `8665f4b4`)
- Related path-security primitive: `src/apm_cli/utils/path_security.py::ensure_path_within`

qorexdevs:
## Description

`_get_cache_dir` passes an unresolved `project_root` to `ensure_path_within`, which then resolves `candidate` and `base_dir` independently. On Windows, `Path.resolve()` on an existing path expands 8.3 short names (e.g. `RUNNER~1` -> `runneradmin`) and may add the `\?\` extended-length prefix, but on a not-yet-existing path it keeps the original form. When `candidate` doesn't exist yet (first writer in a concurrent scenario), the two sides resolve to different forms and `is_relative_to` fails with `PathTraversalError`.

Two changes:
- `discovery.py`: resolve `project_root` once at the top of `_get_cache_dir` so the candidate path inherits the long-name form
- `path_security.py`: strip the `\?\` extended-length prefix in `ensure_path_within` before comparison, since Windows `resolve()` adds it inconsistently

Fixes #886

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass (472 policy unit tests green)
- [x] Added regression test (`test_unresolved_project_root_does_not_raise`) that passes a symlink-indirect project root through `_get_cache_dir`

### DIFF
--- a/src/apm_cli/policy/discovery.py
+++ b/src/apm_cli/policy/discovery.py
@@ -1025,6 +1025,10 @@ def _get_cache_dir(project_root: Path) -> Path:
     project root -- a configuration that, while unusual, would let
     cache reads/writes escape the project tree.
     """
+    # Resolve early so candidate inherits long-name form on Windows;
+    # without this, resolve() on a not-yet-existing candidate keeps
+    # 8.3 short names while the base resolves to long names (#886).
+    project_root = project_root.resolve()
     base = project_root / "apm_modules"
     candidate = base / POLICY_CACHE_DIR
     # Resolve both ends and assert containment under ``project_root``,
--- a/src/apm_cli/utils/path_security.py
+++ b/src/apm_cli/utils/path_security.py
@@ -74,6 +74,19 @@ def validate_path_segments(
             )
 
 
+def _strip_extended_prefix(p: Path) -> Path:
+    """Strip the ``\\\\?\\`` extended-length prefix that Windows' resolve() may add.
+
+    On Windows, ``Path.resolve()`` can inconsistently add the prefix to
+    one path but not another, making ``is_relative_to`` fail even when
+    both paths share the same physical root (#886).
+    """
+    s = str(p)
+    if s.startswith("\\\\?\\"):
+        return Path(s[4:])
+    return p
+
+
 def ensure_path_within(path: Path, base_dir: Path) -> Path:
     """Resolve *path* and assert it lives inside *base_dir*.
 
@@ -83,8 +96,8 @@ def ensure_path_within(path: Path, base_dir: Path) -> Path:
     This is intentionally strict: symlinks are resolved so that a link
     pointing outside the base is caught as well.
     """
-    resolved = path.resolve()
-    resolved_base = base_dir.resolve()
+    resolved = _strip_extended_prefix(path.resolve())
+    resolved_base = _strip_extended_prefix(base_dir.resolve())
     try:
         if not resolved.is_relative_to(resolved_base):
             raise PathTraversalError(
--- a/tests/unit/policy/test_discovery.py
+++ b/tests/unit/policy/test_discovery.py
@@ -234,7 +234,9 @@ class TestCacheReadWrite(unittest.TestCase):
 
     def test_get_cache_dir(self):
         root = Path("/fake/project")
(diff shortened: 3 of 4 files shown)


qorexdevs:

danielmeppiel (approved):

PR #900, opened 2026-04-24
Prototype repository-driven OCI package resolution

vlsi:
Prototype repository-driven package resolution for APM dependencies.

This changes dependency handling so `apm.yml` can declare logical package requirements like `owner/repo#version` without embedding the transport type in the dependency itself. Resolution is performed through configured repositories, with initial support for Git and OCI backends.

## How it works

- OCI repositories are expected to contain a `*.tar.gz` artifact with raw APM package sources.
- Dependency declarations in `apm.yml` do not depend on the repository that stores the dependency.
  This means the same package declaration can be resolved from different repositories based on client configuration.
- Users configure repositories in a separate file, so the same `apm.yml` can be used unchanged in open source and in corporate environments.
- The resolver tries configured repositories in order and records the resolved source in the lockfile.

## Notes

This is still a prototype.

Current limitations:

- repository configuration is file-based
- OCI fetch currently uses `oras`
- OCI publish flow and media-type enforcement are not implemented yet
- resolution remains intentionally simple and tries configured repositories in order

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [ ] Tested locally
- [x] All existing tests pass

### DIFF
--- a/docs/src/content/docs/guides/dependencies.md
+++ b/docs/src/content/docs/guides/dependencies.md
@@ -8,14 +8,14 @@ Complete guide to APM package dependency management - share and reuse context co
 
 ## What Are APM Dependencies?
 
-APM dependencies are git repositories containing `.apm/` directories with context collections (instructions, chatmodes, contexts) and agent workflows (prompts). They enable teams to:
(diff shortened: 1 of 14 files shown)


vlsi:

Issue #852, opened 2026-04-22
Azure DevOps authentication via Entra ID (AAD) bearer tokens


danielmeppiel:
## Summary

APM currently authenticates to Azure DevOps using only Personal Access Tokens via the `ADO_APM_PAT` environment variable. This is a hard blocker for users in environments where:

- Org policy restricts or prohibits PAT creation (common in security-conscious enterprises)
- Authentication is performed via service principals or managed identities (Entra ID / AAD)
- CI/CD pipelines use Workload Identity Federation (WIF) instead of long-lived secrets
- GitHub Actions workflows authenticate to Azure via OIDC federated credentials

This issue proposes adding **Entra ID (AAD) bearer token authentication** as a non-disruptive fallback, picked up automatically from an active `az` CLI session. PAT users see no behavior change. Users who cannot or do not use PATs gain a zero-config path that aligns with the broader industry move away from long-lived shared secrets.

## Problem

Today, every ADO code path in APM (`install`, `outdated`, dependency validation, ref listing) routes through `ADO_APM_PAT`:

- `src/apm_cli/core/token_manager.py:49` -> `'ado_modules': ['ADO_APM_PAT']` is the entire token chain
- `src/apm_cli/core/auth.py` excludes ADO from per-org env-var lookup, credential-fill fallback, and credential-fill retry on failure
- `src/apm_cli/utils/github_host.py:188` `build_ado_https_clone_url()` embeds the token in the clone URL

When an organization disables PAT creation, or when an automated workload runs without a PAT (only an AAD identity), `apm install` fails with no recovery path. The user-facing error tells them to set a variable they cannot generate.

ADO's REST and Git endpoints already accept AAD bearer tokens anywhere a PAT is accepted; we just don't acquire or send them.

## Goals

1. Add an automatic AAD bearer-token path for ADO that requires zero configuration when an `az` CLI session is active.
2. Preserve the existing PAT path with no behavior change for current users.
3. Keep the dependency footprint flat (no new Python packages).
4. Make CI/CD usage trivial: dropping APM into a step that follows `azure/login@v2` (GitHub Actions) or `AzureCLI@2` (Azure Pipelines with WIF) should "just work".
5. Improve error messages so they adapt to what is actually available on the user's machine.

## Non-goals

- Adding an `apm auth` command surface (out of scope; package managers don't manage credentials).
- Adding a `azure-identity` or MSAL Python dependency (rejected during design — too heavy for the marginal coverage gain).
[...]

Issue #827, opened 2026-04-22
Enforce apm-policy.yml at `apm install` time, not only in `apm audit --ci`


danielmeppiel:
## Summary

`apm install` does **not** enforce `apm-policy.yml`. Today, organisational policy is only enforced by `apm audit --ci`. A developer running `apm install` on their workstation can pull and integrate dependencies that the org policy explicitly denies, with no warning and no failure.

The discovery, parsing, and enforcement primitives all already exist in `src/apm_cli/policy/` — they just aren't wired into the install pipeline.

## Evidence

- `src/apm_cli/commands/install.py` — no references to `discover_policy`, `run_policy_checks`, or `ApmPolicy`.
- `src/apm_cli/install/phases/resolve.py` — same: zero policy hooks. The only policy-adjacent code (`install/insecure_policy.py`) handles the unrelated HTTP-dependency consent prompt.
- `src/apm_cli/commands/audit.py:507-540` — the only call site of `discover_policy()` + `run_policy_checks()`. Honours `enforcement: block | warn | off`.
- `src/apm_cli/policy/policy_checks.py:702-760` — `run_policy_checks()` already evaluates dependency allowlist, denylist, required packages, required version, etc. against `apm.yml` + lockfile.

## Impact

The current model relies on `apm audit --ci` being a required CI check on every PR. That assumption is brittle:

1. **Local-only workflows bypass the gate entirely.** Developers running `apm install` against a private branch, a fork, or a personal scratch repo never hit CI, so banned packages install silently.
2. **Transitive dependencies bypass the gate at install time.** A policy-allowed direct dep can pull a denied transitive dep; the user only learns at audit time, after the file system already has the integrated content (and `.github/` files have been deployed).
3. **The integration step has already happened by the time CI fails.** Files have been written to `.github/prompts/`, `.github/skills/`, etc. CI tells you "this should not have been installed" but the developer's working tree is already polluted.
4. **Mismatch with developer mental model.** Coming from `npm`, `pip`, `cargo`, users reasonably expect the package manager itself to enforce org policy at install time, not a separate `audit` command.

## Proposed approach (high level)

1. Call `discover_policy(project_root)` early in `install/phases/resolve.py`, after `apm.yml` parse and after the dependency graph is resolved (so transitive deps are also evaluable).
[...]

PR #928, opened 2026-04-25
feat(install, compile): `--root DIR` to redirect writes while sources resolve from $PWD

### LINKED ISSUE #888: enhancement: `apm install --root <dir>` to target a directory other than $PWD
## Problem

`apm install` always writes relative to `$PWD`. There's no flag to redirect deployment to a different root.

This forces scratch-dir verification workflows (see microsoft/apm#684) to physically stage every input file in the scratch tree and `cd` into it, rather than running `apm install --root <scratch>` against inputs that stay in place.

## Concrete workaround in the wild

juspay/kolu runs `apm install` in a scratch directory as part of `just ci` — the motivating use case on microsoft/apm#684. The workaround currently spans ~20 lines of a justfile recipe (`agents/ai.just:49-77`):

```bash
scratch=$(mktemp -d)
trap 'rm -rf "$scratch"' EXIT
rsync -a \
    --exclude='.git/' \
    --exclude='node_modules/' \
    --exclude='.direnv/' \
    --exclude='.logs/' \
    --exclude='.claude/' \
    --exclude='.codex/' \
    --exclude='.agents/' \
    --exclude='.opencode/' \
    --exclude='CLAUDE.md' \
    "$PWD/" "$scratch/"
mkdir -p "$scratch/.claude" "$scratch/.codex" "$scratch/.opencode" "$scratch/.agents"
[[ ! -e .claude/launch.json ]] || cp .claude/launch.json "$scratch/.claude/launch.json"
[[ ! -e .codex/config.toml ]] || cp .codex/config.toml "$scratch/.codex/config.toml"
(cd "$scratch" && apm install >/dev/null)
(cd "$scratch" && apm compile --target codex,opencode >/dev/null)
```

Every line between `scratch=$(mktemp -d)` and `(cd "$scratch" && apm install ...)` exists only because `apm install` can't take a target directory. With a `--root` flag this collapses to roughly:

```bash
scratch=$(mktemp -d)
trap 'rm -rf "$scratch"' EXIT
apm install --root "$scratch"
apm compile --root "$scratch" --target codex,opencode
```

## Why `--root` (not `cd`-based workaround) is the right primitive

- **No input duplication.** The current workaround copies the entire worktree into scratch (7 exclude rules, a pair of `cp` fallbacks for user-owned runtime files, and `mkdir`s to satisfy target auto-detection). A `--root` flag keeps sources in place — `apm.yml`, the lockfile, and local-path packages all resolve from `$PWD` while writes land under `--root`.
- **Uses beyond verification.** Scripted setup scenarios (bootstrap scripts, fixture generation for tests, CI dry-runs) all benefit from installing into a specific directory without chdir gymnastics.
- **Matches precedent.** `pip install --target`, `npm install --prefix`, `cargo install --root` — redirecting output root is a standard package-manager affordance.

## Scope

Two sub-questions the implementation will need to resolve:

1. **Which paths does `--root` redirect?** I'd expect: deployed-file writes (`.claude/`, `.codex/`, `.agents/`, `.opencode/`), `apm_modules/`, and the lockfile. Sources (`apm.yml`, local-path package roots) should continue resolving from `$PWD` unless a separate `--manifest` flag is added.
2. **Interaction with `apm compile`.** `compile` has the same scratch-dir use case (see the `(cd "$scratch" && apm compile ...)` line above). Worth applying the same flag to `compile` and any other command that writes to the working directory.

## References

- Context: microsoft/apm#684 (content verification in `audit --ci` — Sergio + Daniel both mentioned `--root` in their triage replies)
- Reference implementation of the workaround: juspay/kolu#469
- Workaround source: [`agents/ai.just:49-77`](https://github.com/juspay/kolu/blob/master/agents/ai.just)
- apm 0.9.2 install entry points that assume `Path.cwd()` / `get_deploy_root(scope)`: `src/apm_cli/commands/install.py:1191,1271,1283,1424`

## Environment

- apm 0.9.2 (via `uvx --from git+https://github.com/juspay/apm apm` — pinned to juspay fork for an unrelated fix)

srid:
Closes #888.

**Downstream consumer validating this PR end-to-end**: [juspay/kolu#716](https://github.com/juspay/kolu/pull/716) — collapses a ~60-line `rsync` + `mkdir` + `cp` scratch-staging workaround in their CI pipeline to a four-line `apm install --root` + `apm compile --root` sequence.

## Why

`apm install` and `apm compile` always write relative to `$PWD`, which forces any scratch-directory workflow to physically stage every input file (manifest, `.apm/`, lockfile, project tree) before `cd`-ing in. The juspay/kolu CI pipeline hit this in a sharp way (#684): running `apm install` on the live worktree was briefly destroying `.claude/` while a Claude Code session had an open file handle there. The workaround exists purely because there was no `--target`-style flag.

This PR adds that flag: `apm install --root DIR` and `apm compile --root DIR`. *Writes* go under `DIR`; *sources* (`apm.yml`, `.apm/`, local-path packages, the project tree used for distributed compile placement scoring) continue reading from `$PWD`. Semantics deliberately mirror `pip install --target` and `npm install --prefix`.

With both flags in place the kolu workaround collapses to:

```bash
[[ ! -e apm.lock.yaml ]] || cp apm.lock.yaml "$scratch/apm.lock.yaml"
apm install --root "$scratch"
apm compile --root "$scratch" --target codex,opencode
```

## How

The implementation picks a **chdir + source-root pin** model over refactoring the long tail of `Path.cwd()` / `os.getcwd()` call sites (notably the MCP adapters in `apm_cli.adapters.client.*`, the OpenCode/Cursor/VSCode adapters, and `mcp_integrator.py`). A new context manager `apm_cli.install.root_redirect.install_root_redirect` does two things:

1. `os.chdir(root)` so every site that hardcodes `Path.cwd()` auto-resolves to the deploy root.
2. `set_source_root_override($PWD)` (process-global) so helpers that need the user's source tree (`get_source_root`, `get_manifest_path`) still point at the original working directory.

Both are restored in a `finally` so the context doesn't leak across invocations — important for test runners, REPL sessions, and embedded callers. `compile_root_redirect` is a one-line alias for the same implementation.

`get_manifest_path` is decoupled from `get_apm_dir`: the manifest is a source, so it follows `get_source_root`, while the lockfile / `apm_modules/` / deploy dirs follow `get_deploy_root`. That distinction is the design crux — the issue body explicitly called out *"sources continue resolving from $PWD"* as the intended semantics.

### Install side

- `InstallContext` gains `source_root` (defaults to `project_root` for back-compat via `__post_init__`).
- `install/pipeline.py` computes `source_root = get_source_root(scope)` alongside `project_root = get_deploy_root(scope)`.
- Four places that previously conflated read-from-$PWD with write-to-$PWD now split: `_project_has_root_primitives` (scans source root for `.apm/`), `integrate_local_content`'s synthetic `_local` package (`install_path = source_root`), the dep resolver's `apm.yml` lookup, and local-path package resolution in `phases/resolve.py` + `sources.py`.

The resolver fix is its own commit because the tree-shaped symptom ("running `apm install --root` silently resolved zero direct dependencies") is worth a standalone reviewable change.

### Compile side

`AgentsCompiler` and `DistributedAgentsCompiler` gain an optional `source_dir` parameter (defaults to `base_dir` for back-compat). `base_dir` keeps driving placement targets and write paths; `source_dir` drives primitive discovery, `ContextOptimizer`, and constitution lookup. `DistributedAgentsCompiler._source_to_base` translates placement-map keys from source-rooted to base-rooted so the AGENTS.md writes land at the deploy root even though the placement decisions were scored against the source tree. `template_builder.build_conditional_sections` picks up a matching `source_dir` parameter so `<!-- Source: ... -->` display paths render relative to `$PWD` — without this, scratch-compiled output diverges from in-place output by embedding absolute scratch paths.

## What's included / what's not

Source resolution for `apm.yml` is tied to `get_source_root`, which currently means *the original CWD*. Adding a separate `--manifest PATH` flag for reading the manifest from an arbitrary location (called out as future work in the issue body) is deliberately **not** in this PR — the current design accommodates it cleanly by having `get_manifest_path` consult an additional override, but that's a future increment.

The compiler's `source_dir` is CLI-aware by necessity because distributed-compile placement scoring scans the project tree; a pre-resolved path from the caller would be fine except that `ContextOptimizer` is constructed inside `DistributedAgentsCompiler.__init__` and already took a directory. Threading two paths through is the minimal change.

## Verification

- `tests/unit/install/` (63 tests) + `tests/unit/compilation/` + `tests/unit/commands/` (344 tests combined) all pass on this branch.
- The full `tests/unit/` suite passes (5430 / 5430 excluding 4 environment-dependent failures unrelated to this change: `test_is_tool_available("python")` and three `_real_system` checks — none reference any code I touched).
- The `install.py` LOC budget in `test_architecture_invariants.py` bumps 1700 → 1725 with rationale in the same comment-based tradition as prior PRs (#827, #832, #852, #856); the pending `--mcp` extraction that's already tracked in that test will recover the budget.

**End-to-end validated against [juspay/kolu#716](https://github.com/juspay/kolu/pull/716)**: `apm audit --ci` passes 7/7 and every AGENTS.md output (root + `packages/` + `packages/client/src/` + `packages/tests/features/` + `packages/server/src/`) is byte-identical between an `apm install --root "$scratch" && apm compile --root "$scratch"` run and the in-place equivalent.

## Draft status

Drafted so the design can be sanity-checked before it leaves draft. In particular, feedback welcome on:

- The **chdir-vs-refactor** tradeoff. The `Path.cwd()` long tail is real and this PR sidesteps it; the alternative is a sweep through the adapters to use scope helpers, which is a larger and probably-separate change. Is that acceptable?
- The **`source_dir` leak into the compiler classes**. Principled ("the compiler already knew about roots") or a smell ("CLI-level concept leaking into a pure library")? I picked principled-enough-for-now.
- **Process-global override** vs **plumbed parameter**. A global is ugly but the chdir is already global. Open to being talked out of it.


srid:
## Pre-merge structural review (Hickey + Lowy)

I ran two structural-review passes on this branch — one Hickey-style ("Simple Made Easy", looking for complecting and missed deduplication) and one Lowy-style (volatility-based decomposition, after Parnas '72). Background on what each pass checks: [kolu.dev/blog/hickey-lowy](https://kolu.dev/blog/hickey-lowy/).

Posting both before maintainer review since the two passes **disagree on the central design choice** and I'd rather hear your opinion than commit to one direction blindly.

### Where they converge

**Correctness bug under `--root`** (Hickey, must-fix). `AgentsCompiler` relativises primitive/instruction file paths against `self.base_dir` at three sites — `agents_compiler.py:695`, `agents_compiler.py:707`, `agents_compiler.py:872`. When `--root` is active, those source files live under `source_dir` and `base_dir` is the deploy root; the two don't share a common parent, so `portable_relpath` either produces a confusing `../../…` path or falls back to absolute. `DistributedAgentsCompiler:594` already correctly uses `self.source_dir` — the asymmetry is the smoking gun. Will fix regardless of which design direction you prefer.

**Missing cross-reference docstring on `get_lockfile_dir`** (Hickey, must-fix). `core/scope.py:154` is the boundary where a future maintainer will pick the wrong root. `get_manifest_path` has the right docstring; `get_lockfile_dir` doesn't. Easy fix.

### Where they disagree

The headline: **chdir + process-global override** vs. **just-the-override-and-refactor-cwd-callsites**.

- **Hickey** considers the braid "deliberate, justified" (the docstring explains it; chdir handles the long tail of `Path.cwd()` callsites in MCP adapters; the override handles scope helpers — two mechanical parts of one coherent strategy). *Optional* finding only.

- **Lowy** considers it the *must-fix* problem: "two separate mechanisms for what should be one concept", non-hermetic, and creates a coupling debt where future MCP-adapter additions silently won't see `_SOURCE_ROOT_OVERRIDE`. Recommends **eliminating `chdir()`** and routing every `Path.cwd()` callsite through `get_source_root()` or an explicit parameter — a sweep through `apm_cli.adapters.client.*` and `mcp_integrator.py`.

This is the tradeoff I called out in the PR description's "draft status" section; the two reviewers come down on opposite sides. I lean Hickey's way (smaller diff, contained surface) but Lowy's argument that the long tail *is* the volatility worth encapsulating is real. **Maintainer call: which?**

### Lowy's other architectural pushes

- **`InstallContext.source_root` Optional default is backwards** (should-fix). Resolve at the CLI boundary, make the field non-optional, drop the `ctx.source_root or ctx.project_root` fallback patterns in `phases/resolve.py` and `pipeline.py`. I think this is right and would do it.

- **Compiler `source_dir` leaks CLI concept into domain** (should-fix). The compiler should receive pre-resolved paths and not know about `--root`. The `_source_to_base()` helper is "a guard against the feature, not a design". Counterpoint: `ContextOptimizer` is constructed inside `DistributedAgentsCompiler.__init__` and already takes a directory, so threading a second one is the minimal change. Genuinely uncertain — open to either direction.

### Hickey's other style pushes

- **Manual `__enter__`/`__exit__` instead of `with`** (should-fix). The handlers hold a `_root_redirect` reference and unwind in `finally`. `with` would be cleaner — `install_root_redirect` is already `@contextmanager`. Trivial to fix; the original avoidance was about not re-indenting the existing large `try` body, which is a weak reason.

- **Convention enforced by naming alone, not types** (should-fix). The "use `source_root` to read, `project_root` to write" rule is convention; a phase can read the wrong field and the bug is silent in the no-`--root` case. Suggested fix: convention note at the top of `install/pipeline.py`. Cheaper than a full type-level fix.

### Optional / defer

- `compile_root_redirect = install_root_redirect` alias should carry a comment explaining the intentional sharing and split conditions (Lowy).
- A future `--manifest <path>` flag would re-thread cleanly *if* `source_root` becomes non-Optional (Lowy) — argument for landing the should-fix above.

### Asking

The MUST-FIX correctness bug + docstring will get fixed regardless. For the rest, especially the **chdir-vs-refactor** call, I'd rather have your direction before I rewrite. Happy to push either way; the kolu downstream consumer ([juspay/kolu#716](https://github.com/juspay/kolu/pull/716)) is in draft pending this PR's resolution and isn't blocking.

🤖 Reviews generated with structural-analysis subagents via [Claude Code](https://claude.com/claude-code) (background: [kolu.dev/blog/hickey-lowy](https://kolu.dev/blog/hickey-lowy/)).

srid:
Pushed fix for the two MUST-FIX items (commit 19c20e1):

- **`AgentsCompiler` relpath fix** — `validate_primitives` + verbose-output helper now use `self.source_dir` for primitive/instruction paths, matching `DistributedAgentsCompiler:594`. Verified end-to-end: `apm install --root "$scratch" && apm compile --root "$scratch" --target codex,opencode` against the juspay/kolu worktree produces `AGENTS.md` + `packages/AGENTS.md` that are byte-identical to an in-place compile. 617 unit tests across `tests/unit/{compilation,commands,install}` all pass.

- **`get_lockfile_dir` docstring** — now parallels `get_manifest_path`'s, spelling out the source-vs-deploy routing convention so future metadata helpers pick the right root deliberately.

The should-fix / chdir-vs-refactor items still await your call. Draft stays draft until then.

srid:
Pushed [`24c38bf`](https://github.com/juspay/apm/commit/24c38bf): merged `origin/main` and folded in the Hickey/Lowy review findings.

**Conflict resolution** — followed upstream's PEP 604 / Ruff convention; threaded `--root` plumbing through the new `_install_apm_packages` / `_post_install_summary` decomposition; preserved `source_dir` end-to-end in the compiler so primitive paths in validation warnings, source-attribution comments, and verbose-trace lines render against the source root even under `--root`.

**Addressed findings:**

- **Lowy / should-fix** — `InstallContext.source_root` is now required (no `Optional[Path] = None` default).  Resolved at the CLI boundary in `run_install_pipeline`.  Dropped the `ctx.source_root or ctx.project_root` and `ctx.source_root or ctx.apm_dir` fallbacks in `phases/resolve.py` and `install/sources.py` — they masked bugs whenever `source_root != project_root`.
- **Hickey / should-fix** — replaced manual `__enter__` / `finally: __exit__` pairs with `with install_root_redirect(...)` / `with compile_root_redirect(...)` in both commands.  Re-indents the handler bodies but unwinds correctly even on early `return` or uncaught exceptions, and drops the `_root_redirect` reference from the function frame.
- **Hickey / should-fix** — added a "Source-vs-deploy root convention" section to `install/pipeline.py`'s module docstring spelling out which root each phase reads vs writes.  Future field additions must pick a side deliberately.
- **Lowy / optional** — expanded the comment on `compile_root_redirect = install_root_redirect` to document why the alias exists and when it should split.

The chdir-vs-refactor design call is **still open** for maintainer direction — that's the bigger sweep through `Path.cwd()` callsites in `apm_cli.adapters.client.*` / `mcp_integrator.py`.  Happy to do it if you'd prefer Lowy's direction; otherwise the current chdir + `_SOURCE_ROOT_OVERRIDE` model is documented and contained.

**Verified end-to-end:** `apm install --root SCRATCH` + `apm compile --root SCRATCH --target codex,opencode` produces byte-identical AGENTS.md files to an in-place compile (modulo the non-deterministic Build ID).  All install + compilation unit tests pass (656 + arch-invariant suite).

The kolu downstream consumer ([juspay/kolu#716](https://github.com/juspay/kolu/pull/716)) will be re-pointed at this commit next.

srid:
Pushed [`a2186450`](https://github.com/juspay/apm/commit/a2186450): merged latest `origin/main` (`--frozen`, local-bundle install, `--refresh`, `--as`, `--legacy-skill-paths`, `--all` for compile, `parse_targets_field`, `install_started_at` timing, etc.) and addressed Copilot's automated review.

**Review findings — addressed:**

1. **`compile/cli.py:396` — `--watch` under `--root` scans the wrong tree.** Now rejected upfront: `if root and watch: raise click.UsageError("--root is not valid with --watch")`. The flag combination has no real use case (watch is interactive dev; `--root` is for CI scratch-dir verification).
2. **`root_redirect.py:45` — dry-run leaks a `mkdir`.** Threaded `dry_run` through the context manager. With `--dry-run`, the redirect refuses to create the target directory and raises a `click.UsageError` instead — preview cannot mutate the filesystem.
3. **`commands/install.py:1019` — logger `UnboundLocalError`.** Already addressed upstream (`logger = None` initialized before `try`; `except Exception` guards `if logger:`). Preserved through the merge.
4. **`template_builder.py:20` — E305 lint.** Added the missing second blank line between `TemplateData` and `render_instructions_block`.
5./6. **CLI doc drift.** Added `--root` to `docs/src/content/docs/reference/cli-commands.md` (both `apm install` and `apm compile` sections) and to `packages/apm-guide/.apm/skills/apm-usage/commands.md`. Called out `--global` / `--watch` incompatibility and the `--dry-run` mkdir-refusal in both surfaces.
7. **`root_redirect.py:32` — no unit tests.** New `tests/unit/install/test_root_redirect.py` covers: no-op when `None`/empty, chdir + override on entry, restore-on-exception, mkdir-on-write, `dry_run` refuses-mkdir, `dry_run` with existing dir, compile-alias-identity. 8 tests, all passing.
8. **`core/scope.py:150` — USER-branch duplicates `get_apm_dir`.** `get_manifest_path` now delegates: `if scope is InstallScope.USER: return get_apm_dir(scope) / APM_YML_FILENAME`.

**Verified end-to-end:**

- 1039 install + compilation unit tests pass (656 baseline + new `--root` tests + the larger upstream surface).
- Ruff clean on every touched file.
- `apm install --root SCRATCH && apm compile --root SCRATCH --target codex,opencode` produces byte-identical `AGENTS.md` to an in-place compile (Build ID modulo).
- New CLI rejections: `--root + --watch`, `--root + --global`, and `--dry-run + --root` with a missing `DIR` all emit `click.UsageError` (exit 2) before any chdir/mkdir runs.

The chdir-vs-refactor design call is **still open** for maintainer direction (the bigger sweep through `Path.cwd()` callsites in `apm_cli.adapters.client.*` / `mcp_integrator.py`) — happy to do it if you'd prefer Lowy's direction.

danielmeppiel:
Re-ported onto current `main` as #1628.

This PR was accepted (#888) but had gone 187 commits stale with 10
conflicts (CONFLICTING/DIRTY). Rather than a textual conflict-merge, I
re-applied the `--root DIR` feature **semantically** onto current `main`
and ran it through an APM review-panel pass, folding the panel's
improvements:

- module-global override -> `contextvars.ContextVar` (thread-safe) with a
  hardened `finally`;
- `<!-- Source: ... -->` provenance fixed in **both** the single-file and
  distributed paths so output stays byte-identical under `--root`;
- resolver anchors to the source root only when `--root` is active (zero
  regression for normal installs);
- added 8 unit + 3 e2e tests; full lint chain + 2616 tests green.

@srid is preserved as the commit author on #1628. I could not force-push
to this branch directly because GitHub disables maintainer-edit pushes for
**organization-owned forks** (`juspay/apm`) -- that limitation, not the
work, is the only reason this lives in a superseding PR.

Suggest closing this in favor of #1628. Thank you @srid for the feature.

PR #937, opened 2026-04-25
fix(install): allow local packages at --global scope; fix broken tests

stbenjam:
## Summary

I spotted some failing tests that don't run in CI.

- Remove the validation that rejected local filesystem packages when using `--global`. There is no technical reason local paths cannot be installed at user scope -- primitives are copied to `~/.claude/`, `~/.gemini/`, etc. the same way as at project scope.
- Add `test_global_scope_e2e.py` to `scripts/test-integration.sh` so CI catches regressions in `--global` scope tests going forward. This file was never included in CI, so the broken test went undetected.
- Fix `test_user_scope_skips_workspace_runtimes` which was missing the `MCPServerOperations` mock. The test passed in CI by accident: no runtimes are installed in CI, so `MCPIntegrator.install()` returns early at the scope-filtering stage before reaching the registry validation. Locally, if `codex` is installed, `target_runtimes` survives scope filtering and hits the real registry, which fails on the fake `test/server` name. The sibling test (`test_project_scope_includes_all_runtimes`) already had the correct mock.
- Fix `test_init_integration.py` — PR #1038 changed `apm marketplace init` to write `apm.yml` instead of `marketplace.yml`, but the tests still looked for the old filename and used legacy loaders. Updated to use `apm.yml`, `load_marketplace_from_apm_yml()`, and `render_marketplace_block()`.
- Fix `test_outdated_integration.py` — PR #1024 changed `apm marketplace outdated` to exit 1 when upgradable packages exist, but the test still asserted exit code 0. Updated to match the new behavior.
- Add `tests/integration/marketplace/` to `scripts/test-integration.sh` so CI catches marketplace authoring test regressions going forward.

## Test plan

- [x] `pytest tests/unit tests/test_console.py tests/integration/test_global_scope_e2e.py` -- 5459 passed, 0 failed
- [x] `test_auto_bootstrap_creates_user_manifest` now passes (previously broken since local-package rejection was added)
- [x] `test_user_scope_skips_workspace_runtimes` now passes deterministically regardless of locally installed runtimes
- [x] `pytest tests/` -- 7452 passed, 181 skipped (excluding env-dependent e2e/live tests)
- [x] `test_creates_apm_yml` and all marketplace init tests pass against current `init` command
- [x] `test_exit_code_one_when_upgradable` passes against current `outdated` command


danielmeppiel (changes requested):
Great one but need adjustment per panel findings

danielmeppiel:
## Maintainer panel verdict (per #1064 FIX-NOW orchestration)

Ran the apm-review-panel locally with **real-asset reproduction**: `apm install --global ./local-pkg` against a scratch `HOME`.

**Verdict: NEEDS-AUTHOR (the 4 prior reviewer asks remain UNADDRESSED on this branch)**

### Reproduction (PR is currently a no-op for its stated use case)
With your branch checked out:
```
HOME=<scratch> apm install --global ./local-pkg
=> exit 0
=> apm.yml lists `./local-pkg`
=> apm_modules/ is EMPTY
```
The manifest-vs-deployment desync is exactly what the prior bot panel called out -- the CLI-level guard is removed but the **downstream guards still reject USER scope silently**, so the install succeeds in the manifest sense but never deploys files.

### Panel agrees with prior reviewer (CONFLICT RULE prevented silent override)
Local 7-persona panel reaches the SAME conclusion as the prior `apm-review-panel` bot panel that produced `danielmeppiel`'s CHANGES_REQUESTED on `64cee38d`. No conflict between panels; the 4 asks below stand.

### Asks for @stbenjam (re-stating the 4 outstanding items)
1. **Remove the second guard at `src/apm_cli/install/sources.py:141-153`** that rejects USER scope for local sources -- the CLI-level guard removal is incomplete without this.
2. **Remove the third guard at `src/apm_cli/install/phases/resolve.py:136-142`** that blocks transitive local dependencies at USER scope.
3. **Verify `_copy_local_package` resolves `project_root` correctly at user scope** -- when there is no project root (`apm install --global` from arbitrary cwd), the relative-source resolution needs an explicit base.
4. **Add a `CHANGELOG.md` [Unreleased] / Fixed entry** for this fix.

### Test evidence (your current HEAD)
```
unit (touched files): 75/75 pass
integration: 21 pass / 15 skip (token/binary-gated)
real-world: confirmed PR is a no-op for local --global install
```

Once the 4 items are addressed, ping and I'll re-run the panel + merge.

danielmeppiel (changes requested):
Per maintainer panel verdict in this thread: 4 prior asks remain unaddressed; PR is verified no-op for local --global install. Re-requesting changes formally.

stbenjam:
@danielmeppiel Ok so I fixed the comments, but new tests fail after rebasing (that I've also fixed here).  Can you help me understand how tests in `scripts/test-integration.sh` are decided? The broken tests aren't triggered in CI, so they seem to get broken often.

Is it an oversight? Should we just have pytest run all integration tests instead of per-file list in the script?

stbenjam:
> @danielmeppiel Ok so I fixed the comments, but new tests fail after rebasing (that I've also fixed here). Can you help me understand how tests in `scripts/test-integration.sh` are decided? The broken tests aren't triggered in CI, so they seem to get broken often.
> 
> Is it an oversight? Should we just have pytest run all integration tests instead of per-file list in the script?

Also I'd hold off on additional reviews until I sort out what my goal is in this PR. I'd like to solve the general integration test problem that I keep running into

danielmeppiel:
## Maintainer follow-up: lint + deeper finding + test-integration answer

Thanks for the rebase Stephen! Three things:

### 1. Lint failure (trivial, 1-line)

`Lint` is red on `c9159802` because of a single missing blank line in `tests/unit/test_install_command.py` near line 838. Run:

```
uv run --extra dev ruff format tests/unit/test_install_command.py
```

That is the entire fix.

### 2. Real-asset reproduction: the user-scope local-pkg path is **still broken**

I reran the repro from my prior comment against your rebased branch, with an **absolute** path (so the relative-resolution concern in the `sources.py` guard doesn't apply):

```bash
HOME=$TMP/fakehome  apm install --global --verbose $TMP/local-pkg
```

Output:
```
[+] $TMP/local-pkg
[*] Updated apm.yml with 1 new package(s)
...
  Skipping $TMP/local-pkg (already failed during resolution)

  [+] <project root> (local)
[*] Installed 1 APM dependency.
```

Result: package is **recorded in `~/.apm/apm.yml`** but **no content is deployed** under `~/.apm/`. The user gets a green `[+]` and `[*] Installed 1 APM dependency.` while nothing actually shipped. This is worse than the pre-PR state -- now the manifest is poisoned.

The cause is the **second** of the three guards the panel called out:

```
src/apm_cli/install/sources.py:140-153
        if ctx.scope is InstallScope.USER:
            diagnostics.warn(
                f"Skipped local package ... -- local paths are not supported at user scope (--global). "
                "Use a remote reference (owner/repo) instead.",
                ...
            )
            return None
```

This guard fires unconditionally for `InstallScope.USER` -- it does not distinguish relative from absolute paths. Absolute paths are unambiguous (no $HOME-vs-cwd ambiguity), so the guard's stated rationale only applies to relative paths. The minimum fix is to scope the skip to relative paths only:

```python
if ctx.scope is InstallScope.USER and not Path(dep_ref.local_path).is_absolute():
    # Skip relative-path local packages at user scope ...
    return None
```

Or to remove the guard entirely (if you trust upstream validation, which already absolutizes Click `--global` path args). Either is a small change but it is **the change the panel was asking for** -- the `commands/install.py` removal you did is necessary but not sufficient, because resolution happens later.

### 3. Your `scripts/test-integration.sh` question (good catch)

Today the script enumerates a hand-curated list of `tests/integration/*_e2e.py` files (search for `pytest tests/integration/` in the file). New e2e modules are not picked up automatically. There is no policy reason for this -- it is an oversight; we just have not had a forcing function to fix it.

A `pytest tests/integration -m e2e` (or directory-glob with explicit excludes for token-gated suites like ADO) would absolutely be cleaner. That is out of scope for #937 but worth a separate one-line PR -- I would merge it.

### What I need from you

- Lint format fix (1 line).
- One of the two `sources.py` guard adjustments above (or argue against it with a counter-repro showing my test is wrong).

Re-push and I will run the full panel + real-asset repro again. We are close.

danielmeppiel:
Pushed lint fix + the missing `sources.py` guard adjustment (plus the parallel guard in `phases/resolve.py`) + E2E deployment assertion. Real-asset repro now shows content actually deploying under `~/.apm/apm_modules/_local/<pkg>/.apm/instructions/`.

Heuristic chosen: only **relative** local paths are rejected at user scope (project-relative refs have no meaningful root when cwd is arbitrary and $HOME is not a project). Absolute paths flow through the same `_copy_local_package` path used at project scope.

Strengthened `test_auto_bootstrap_creates_user_manifest` to assert deployed file existence -- this would have caught the original bug (manifest-only check passed against broken behaviour).

Co-authored your name on the commit; happy to discuss the guard heuristic if you'd prefer a different shape.

stbenjam:
I had imagined just resolving relative paths, but I don't have a use case for it, I was just trying to fix failing tests.  Thanks!

Issue #820, opened 2026-04-21
[BUG] apm fails to install or compile dependencies


petemounce:
**Describe the bug**

apm fails at core functionality.

**To Reproduce**
1. Make an apm.yml with 1 dependency:
    ```yaml
    name: x
    version: 0.1.0
    description: APM project for x
    author: Peter Mounce
    dependencies:
      apm:
      - JuliusBrussee/caveman#84cc3c14
      mcp: []
    scripts: {}
    target: opencode,claude,copilot,agents
    ```

3. Run `apm install`
4. Command completes. `apm_modules/` is created with the dependency within.
5. No content is "deployed" to the filesystem of the repository. `.claude/skills`, `.github/skills`, `.opencode/skills`, `.agents/skills` are all empty.
6. Run `apm compile`
7. No content is "deployed" to the filesystem of the repository. `.claude/skills`, `.github/skills`, `.opencode/skills`, `.agents/skills` are all empty.

**Expected behavior**

I expect the caveman skills to be deployed into the places where the agents will load them, based on reading the documentation's quick-start. 

Instead, I get successful exit codes and log messages - but no actions appear to have been actually taken.

**Environment (please complete the following information):**
 - OS: macOs
 - Python Version: 3.14.3
 - APM Version: 0.8.12
 - VSCode Version (if relevant): n/a

**Logs**
```console
apm install --verbose
[>] Installing dependencies from apm.yml...
Parsed apm.yml: 1 APM deps, 0 MCP deps
Using apm.lock.yaml (1 locked dependencies)
    JuliusBrussee/caveman: locked at 84cc3c14
Resolved 1 direct dependencies (no transitive)
  [+] JuliusBrussee/caveman #84cc3c14 @84cc3c14 (cached)

No MCP dependencies found in apm.yml

[*] Installed 1 APM dependency.
```

```console
apm compile --clean
[*] Starting context compilation...
[i] Compiling for AGENTS.md + CLAUDE.md + .github/ + .claude/ + .cursor/ + .opencode/ + .codex/ + .agents/ - detected .github/ and .claude/ and .opencode/ folders
[+] Compilation completed successfully!
```

```console
ls -al .opencode .claude .github
.claude:
total 16
drwxr-xr-x@  4 pmounce  staff   128 21 Apr 17:32 .
drwxr-xr-x@ 45 pmounce  staff  1440 21 Apr 18:12 ..
[...]

PR #951, opened 2026-04-26
fix(install): handle bare-string MCP entry in shell-metachar warning

### LINKED ISSUE #938: Bug: apm install --mcp without --transport fails with AttributeError
## Description

`apm install --mcp <registry-ref>` without `--transport` fails with:

```
'str' object has no attribute 'get'
```

The code path expects the MCP dependency to be a dict (with `.get()`), but when a bare registry reference string is passed without `--transport`, it remains a `str`.

## Reproduce

```bash
apm install --mcp io.github.github/github-mcp-server
```

## Workaround

Specify `--transport` explicitly:

```bash
apm install --mcp io.github.github/github-mcp-server --transport stdio
```

## Expected behavior

`apm install --mcp <ref>` should either infer a default transport or provide a clear error asking the user to specify one.

## Affected versions

Observed on 0.10.0. Affects all targets (not Gemini-specific).

edenfunf:
## Description

`apm install --mcp <registry-ref>` without `--transport`, `--url`,
`--mcp-version`, `--registry`, or a post-`--` stdio command aborts with:

```
'str' object has no attribute 'get'
```

The pure builder `_build_mcp_entry` returns a bare `str` for the
documented bare-string registry shorthand path (preserving the
`mcp: [foo]` apm.yml UX contract), but the F7 shell-metachar warning
callsite in `_run_mcp_install` read `entry.get("command")` unguarded.
The exception fires before apm.yml is written, so the user is left
with no manifest update and only a workaround (`--transport stdio`).

### Root cause

`_build_mcp_entry` returns `str | dict` by design — bare strings for
shorthand-with-no-overlays, dicts otherwise. Two later callsites in
`_run_mcp_install` already guard with `isinstance(entry, dict)`; the
F7 warning callsite did not, and silently broke the bare-string
happy path.

### Fix

Source the stdio command from `command_argv` (the CLI input that
populates `entry["command"]` in the dict-entry path) rather than
introspecting the post-build entry. The two values are identical in
the stdio path and both yield `None` elsewhere, so warning behaviour
is preserved bit-for-bit while the str-vs-dict dispatch disappears
from this callsite. The other `isinstance`-guarded callsites are
left untouched to keep the diff minimal.

Fixes #938

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

### Manual verification

| Path | Command | Result |
|---|---|---|
| Issue reproduce | `apm install --mcp io.github.github/github-mcp-server` | exit 0; bare-string entry persisted to `apm.yml` |
| Workaround (regression check) | `... --transport stdio --dry-run` | exit 0; dry-run preview |
| Self-defined stdio dict path | `apm install --mcp myfetch -- npx -y @modelcontextprotocol/server-fetch` | exit 0; dict entry persisted; integrators configured |
| F7 warning still fires | `apm install --mcp evil --force -- 'echo;rm' arg` | shell-metachar warning emitted with the correct command value |

### Automated tests

- New regression test `test_mcp_registry_shorthand_no_overlays_persists_bare_string`
  in `tests/unit/test_install_command.py` reproduces the issue path and
  asserts `exit_code == 0` plus bare-string serialization.
- Verified the test fails on the unpatched code with the exact
  `'str' object has no attribute 'get'` message and passes after the fix.
- Full `test_install_command.py` (100 tests) and the broader MCP
  builder / overlay / yml-writer suites (137 tests) remain green.


danielmeppiel (dismissed):

danielmeppiel (changes requested):
commands/install.py grew to 1704 LOC (budget 1700). Do NOT trim cosmetically -- engage the python-architecture skill (.github/skills/python-architecture/SKILL.md) and propose an extraction into apm_cli/install/.

edenfunf:
**@danielmeppiel Thanks for the review!** Both points addressed and pushed.

## Changes

- **LOC budget**: engaged the python-architecture skill and extracted the `--mcp` install path into five sibling modules under `apm_cli/install/`:
    - `mcp_args.py`
    - `mcp_entry.py`
    - `mcp_writer.py`
    - `mcp_conflicts.py`
    - `mcp_command.py`
    
    `commands/install.py` is now ==1317 LOC== (383 lines of headroom). Behaviour is bit-exact — the original `_xxx` symbol names are re-exported from `commands/install.py` so existing test patches (`@patch("apm_cli.commands.install._run_mcp_install")` etc.) and direct imports keep working without modification.

- **CHANGELOG**: added an entry under `[Unreleased] > Fixed`.

## Status


Also merged the latest `main` to clear the conflict against the 0.9.4 release cut — PR is now `MERGEABLE`. Full unit suite (==5574 tests==) passes, including the #938 regression test.

Will be more mindful of the LOC invariant on future `install.py` PRs. Thanks again for the careful review!

danielmeppiel (changes requested):
consider grouping under a `mcp` folder and stripping `mcp_` file prefixes

edenfunf:
@danielmeppiel done — moved `install/mcp_*.py` (×7) into `install/mcp/` and stripped the `mcp_` prefix from each. Lines up with the unprefixed-noun naming the rest of `install/` already uses (`pipeline.py`, `context.py`, `service.py`, ...).

Pure refactor — behaviour and public symbol names unchanged; the back-compat `_xxx` re-bind aliases in `commands/install.py` still resolve. All `install/` unit tests, the #938 regression, and the LOC budget invariant stay green; `commands/install.py` is at 1187 LOC. Smoke-tested the bare-string registry shorthand, the stdio-dict path, and the F7 metachar warning live too.

danielmeppiel (changes requested):
**Required actions before merge**

Restore apm.yml target normalization (src/apm_cli/models/apm_package.py line ~206, src/apm_cli/core/target_detection.py). APMPackage.from_apm_yml() must normalize the raw target: value the same way TargetParamType.convert() does -- CSV split, alias resolution, VALID_TARGET_VALUES check -- before storing it. The simplest fix is to re-add parse_target_field() (or re-export TargetParamType-equivalent logic) and call it in from_apm_yml(). Without this, target: "claude,copilot" in apm.yml silently falls back to copilot-only with no warning.

Add CHANGELOG entry for marketplace scope narrowing. Under [Unreleased] > Changed, add a line documenting that marketplace url: sources now require https://github.com/ or `(github.com/redacted) URLs (GHES and non-GitHub URLs are no longer resolved). One line is sufficient.

writer.py calls _rich_echo() and _rich_warning() directly; these should route through the passed logger parameter to stay consistent with the CommandLogger pattern.

Restore the zero-target install warning in install/phases/targets.py (or open a tracking issue -- the comment cited #820 as the motivating rationale for that warning).

The unified loop in active_targets() / active_targets_user_scope() was replaced with duplicated isinstance(list) / str branches; consider re-unifying to a single normalization step at the top of each function.

edenfunf:
Pushed `bf3ae07` addressing the panel verdict's required and optional actions:

- `apm_package.from_apm_yml()` now normalizes `target:` through `parse_target_field()` (CSV split, alias resolution, `VALID_TARGET_VALUES` check), matching the `--target` CLI behaviour.
- `CHANGELOG.md` `[Unreleased] > Changed` documents the marketplace `url:` scope narrowing to `https://github.com/` / `http://github.com/`.
- `install/mcp/writer.py` routes the replacement-diff output through the injected logger (defaulting to `NullCommandLogger`); the direct `_rich_echo` / `_rich_warning` calls are gone.
- `install/phases/targets.py` restores the zero-target install warning.
- `integration/targets.py` re-unifies `active_targets()` / `active_targets_user_scope()` via a single normalization step at the top of each function.

Test status: 6645 unit tests pass; the 5 remaining failures are pre-existing Windows path-separator / absolute-path quirks unrelated to this change (`tests/unit/test_config_command.py` cowork skills dir, `tests/unit/test_script_formatters.py`). The #938 regression test still passes.

Re-requesting review when convenient — happy to iterate further if anything still looks off.

PR #974, opened 2026-04-26
feat(install): SKILL_BUNDLE -- day-0 install parity with `npx skills add`

danielmeppiel:
# feat(install): SKILL_BUNDLE -- day-0 install parity with `npx skills add`

## TL;DR

APM now recognises `skills/<name>/SKILL.md` packages -- the convention shipped by `vercel-labs/agent-skills`, `xixu-me/skills`, `larksuite/cli`, `firebase/agent-skills`, and the rest of the agentskills.io ecosystem -- as a first-class shape (`SKILL_BUNDLE`). Every public repo that installs cleanly with `npx skills add owner/repo` now installs cleanly with `apm install owner/repo`. Adding `apm.yml` to one of these repos is strictly additive (lockfile + pinning); it can no longer regress installability the way it did with `danielmeppiel/genesis` (`Missing required directory: .apm/`).

> [!NOTE]
> Live integration tests run against 8 real GitHub repos (4 plugin-shape, 4 SKILL_BUNDLE) on every CI invocation of `-m live`. Marker registered in `pyproject.toml`.

## Problem (WHY)

- [x] **Backwards inversion in the classifier.** `apm.yml + skills/<name>/SKILL.md + no .apm/` short-circuited to `APM_PACKAGE` and hard-failed validation with `Missing required directory: .apm/`. The same repo without `apm.yml` worked, via `MARKETPLACE_PLUGIN`. Adding `apm.yml` *regressed* installability -- the opposite of what a manifest should ever do.
- [x] **`MARKETPLACE_PLUGIN` was over-loose.** A bare `skills/` directory triggered plugin classification by accident, masking the absence of an explicit shape for the npx-skills convention. `obra/superpowers` (a real Claude Code plugin) and `vercel-labs/agent-skills` (a bare skills repo) classified the same way, which made every change to plugin handling a coin-flip for skill bundles.
- [x] **No multi-skill ergonomics.** No way to install one skill out of a 6-skill bundle. `npx skills add owner/repo --skill <name>` works; `apm install owner/repo --skill <name>` did not exist.
- [!] **3 unguarded `shutil.copytree` calls** in `skill_integrator.py` (lines 339/581/842) followed symlinks. A malicious tarball could escape the deploy root through a symlinked `skills/<name>/` entry.

[...]

PR #980, opened 2026-04-27
docs(readme): add 'Coming from npx skills add?' conversion block

danielmeppiel:
## TL;DR

Adds one ~5-line "Coming from `npx skills add`?" block to `README.md`, sitting between the `git clone / apm install` example and the "three promises" section -- the exact funnel moment where readers familiar with `npx skills add` decide whether to switch. No other content changes.

## Problem (WHY)

The 0.9.4 release shipped `SKILL_BUNDLE` ([PR #974](https://github.com/microsoft/apm/pull/974)) so that every public repo installable via `npx skills add owner/repo` is also installable via `apm install owner/repo`. That is the strongest single conversion claim APM has against the current incumbent in agent-skill installation (`skills.sh` / `vercel-labs/skills`), and it is invisible on the README -- so the conversion never happens.

Without this block, an `npx skills add` user reading our README sees a manifest-shaped product that looks unrelated to what they do today. The mental cost to evaluate "would this also work for me?" is high enough that most bounce.

## Approach (WHAT)

A surgical edit, not a rewrite:

- **Where:** between the `git clone / apm install` example (existing line 45) and `## The three promises` (existing line 47). That is the moment a reader has just understood "what APM does" but has not yet committed to learning more.
- **Form:** bold inline lead (`**Coming from npx skills add?**`) -- not an h3 -- so non-npx readers scroll past without losing the page narrative; npx readers self-identify in <1s.
- **Two real install commands:** `apm install vercel-labs/agent-skills` and `apm install vercel-labs/agent-skills --skill deploy-to-vercel`. Both repos / skill names are real, falsifiable in 30 seconds.
- **One outbound link only:** [`reference/package-types/#skill-bundle`](https://microsoft.github.io/apm/reference/package-types/#skill-bundle) -- where the SKILL_BUNDLE shape, manifest, and lockfile are already documented. README's job is to convert the click; docs explain.

## Implementation (HOW)

```diff
 apm install    # every agent is configured
 ```

+**Coming from `npx skills add`?** Drop-in:
+
+```bash
+apm install vercel-labs/agent-skills                            # whole bundle, like npx skills add
+apm install vercel-labs/agent-skills --skill deploy-to-vercel   # one skill, persisted to apm.yml
+```
+
+Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/#skill-bundle).
+
 ## The three promises
```

9 insertions, 0 deletions. No other file touched.

## Trade-offs

- **Bold inline lead vs. h3 sub-section.** h3 would give the block more visual weight but would also register as a major page section, slowing the funnel for the ~80% of readers who don't come from `npx skills add`. The bold inline is intentionally a "side note" so it's invisible if not relevant.
- **Naming a competitor in the README.** `npx skills add` is a known surface in the agent-skills ecosystem (`skills.sh` / `agentskills.io`), and the framing is incumbent-friendly ("drop-in", "same install gesture") rather than challenger-aggressive. Lower risk than naming a different commercial product. Not naming it would mean readers who use it daily can't recognize themselves on the page.
- **Concrete repo + skill names rather than `<owner>/<repo>` placeholders.** Concrete is more compelling and verifiable; it does mean the block depends on `vercel-labs/agent-skills` continuing to exist with `deploy-to-vercel` in it. Both are stable public artifacts; the cost of a future stale claim is one README PR.

## Validation evidence

Both commands verified locally against `vercel-labs/agent-skills` immediately before this PR:

```
$ apm install vercel-labs/agent-skills
[+] vercel-labs/agent-skills (cached)
|-- 7 skill(s) integrated -> .github/skills/

$ apm install vercel-labs/agent-skills --skill deploy-to-vercel
[+] vercel-labs/agent-skills (cached)
|-- 1 skill(s) integrated -> .github/skills/
[*] Persisted skill subset for vercel-labs/agent-skills: [deploy-to-vercel]
```

Resulting `apm.yml` after the second install:
```yaml
dependencies:
  apm:
  - git: vercel-labs/agent-skills
    skills:
    - deploy-to-vercel
```

Resulting `apm.lock.yaml`:
```yaml
- repo_url: vercel-labs/agent-skills
  resolved_commit: ce3e64e468f8fa09a2d075d102771838061fdac0
  package_type: skill_bundle
  skill_subset:
  - deploy-to-vercel
```

Every claim in the new block (drop-in install, `--skill` lever, persistence to apm.yml, lockfile reproducibility) is reproducible from the snippet itself.

## How to test

1. Open the rendered README on this branch.
2. Confirm the new block sits between the `git clone` example and the "three promises" section.
3. Run the two `apm install` commands above against APM 0.9.4 and confirm the integration counts and the `apm.yml` / `apm.lock.yaml` shape match what the block claims.

## Drift / future-proofing

The block depends on `vercel-labs/agent-skills` continuing to publish a skill named `deploy-to-vercel`. If we ship a CI job that validates the README install commands against APM regularly, any drift would surface immediately. Not in scope for this PR; flagged for follow-up if the panel wants it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ apm install vercel-labs/agent-skills                            # whole bundle, 
 apm install vercel-labs/agent-skills --skill deploy-to-vercel   # one skill, persisted to apm.yml
 ```
 
-Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/).
+Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/#skill-collection-skillsnameskillmd).
 
 ## The three promises
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ apm install vercel-labs/agent-skills                            # whole bundle, 
 apm install vercel-labs/agent-skills --skill deploy-to-vercel   # one skill, persisted to apm.yml
 ```
 
-Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/#skill-bundle).
+Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/).
 
 ## The three promises
 
--- a/README.md
+++ b/README.md
@@ -44,6 +44,15 @@ git clone <org/repo> && cd <repo>
 apm install    # every agent is configured
 ```
 
+**Coming from `npx skills add`?** Drop-in:
+
+```bash
+apm install vercel-labs/agent-skills                            # whole bundle, like npx skills add
+apm install vercel-labs/agent-skills --skill deploy-to-vercel   # one skill, persisted to apm.yml
+```
+
+Same install gesture. You also get a [manifest, lockfile, and reproducibility](https://microsoft.github.io/apm/reference/package-types/#skill-bundle).
+
 ## The three promises
 
 ### 1. Portable by manifest
--- a/uv.lock
+++ b/uv.lock
@@ -179,7 +179,7 @@ wheels = [
 
 [[package]]
 name = "apm-cli"
-version = "0.9.3"
+version = "0.9.4"
 source = { editable = "." }
 dependencies = [
     { name = "click" },


danielmeppiel:
Thanks @copilot-pull-request-reviewer -- two real issues caught in one comment.

Pushed `42d7aa76` dropping the fragment entirely and linking to the page root `https://microsoft.github.io/apm/reference/package-types/`. Reasoning:

1. **Wrong anchor slug.** Verified against the deployed page: only `#skill-bundle-skillmd-at-root` exists; `#skill-bundle` is not a valid anchor.
2. **Wrong heading semantically.** That `SKILL.md-at-root` heading is the single-skill bundle shape, but `npx skills add`'s layout is the multi-skill `skills/<name>/SKILL.md` layout, documented under "Skill collection" further down the same page.
3. **Source has the right anchor, but it isn't deployed yet.** "Skill collection" landed in 0.9.4 source; the v0.9.4 docs auto-deploy (from #953) is mid-rollout, so the README has to work against the *live* site, not the in-flight one.

Linking to the page root is robust today (lands on the right doc, both shapes visible above the fold) and self-correcting once 0.9.4 docs deploy completes.

danielmeppiel:
Update: v0.9.4 docs are now live. The deploy was actually broken in a workflow gate (the skip we saw was not just lag) -- root cause + fix in #981.

Manually triggered `workflow_dispatch` run [24983311706](https://github.com/microsoft/apm/actions/runs/24983311706) succeeded; live site now exposes the precise anchor:

```
$ curl -s https://microsoft.github.io/apm/reference/package-types/ | grep -oE 'id="skill[^"]*"'
id="skill-bundle-skillmd-at-root"
id="skill-collection-skillsnameskillmd"
```

Restored the precise fragment in `14e60ac0` so the README lands readers exactly on the "Skill collection (`skills/<name>/SKILL.md`)" heading -- the layout npx skills users already know -- instead of the page root.

So this PR now ships with both the live docs *and* the right anchor. Conversion funnel from npx-skills users -> APM lands on the heading that matches their existing layout, not on a page they have to scan.

PR #1020, opened 2026-04-28
Fix #1019: Improve compile target specification

### LINKED ISSUE #1019: [BUG] apm compile --targets with multi-target list generates GEMINI.md even when Gemini is not a requested target
**Describe the bug**

When using `apm compile --targets` with a comma-delimited list that spans multiple compiler families (e.g., `claude,codex`), the compile command incorrectly generates `GEMINI.md` even though Gemini was not included in the target list. This happens because [`_resolve_compile_target()` in `src/apm_cli/commands/compile/cli.py`](https://github.com/microsoft/apm/blob/bb94e789a888d4143e347bff9f37cdbb3ef75a63/src/apm_cli/commands/compile/cli.py#L166) collapses any multi-family combination into `"all"`, and `"all"` unconditionally triggers generation of all output files including `GEMINI.md`.

**To Reproduce**

1. Run `apm compile --targets claude,codex`
2. Observe that `GEMINI.md` is generated in the output
3. The same occurs with any cross-family combination that does not include `gemini`, e.g., `--targets claude,cursor`

**Expected behavior**
Only `CLAUDE.md` and `AGENTS.md` should be generated when `--targets claude,codex` is specified. `GEMINI.md` should only be generated when the target list explicitly includes `gemini`, or when `--targets all` is used.

- OS: macOS 26.4.1
- Python Version: 3.14.4
- APM Version: 0.10.0

**Logs**
N/A

**Additional context**
The root cause is in [`_resolve_compile_target()` (`src/apm_cli/commands/compile/cli.py`, lines 166–198)](https://github.com/microsoft/apm/blob/bb94e789a888d4143e347bff9f37cdbb3ef75a63/src/apm_cli/commands/compile/cli.py#L166). When the target is a list, it groups targets into three families (`has_agents_family`, `has_claude`, `has_gemini`) and counts distinct families. If `distinct >= 2`, it returns `"all"`. Downstream, `should_compile_gemini_md()` ([`src/apm_cli/core/target_detection.py`, line 162](https://github.com/microsoft/apm/blob/bb94e789a888d4143e347bff9f37cdbb3ef75a63/src/apm_cli/core/target_detection.py#L162)) returns `True` for `"all"`, so `GEMINI.md` is always generated.

The issue is that `"all"` is overloaded — it means both "the user explicitly requested all targets" and "the user requested targets from more than one compiler family." There is no way to represent "claude + agents-family but not gemini" in the current scheme.

tillig:
## Description

Enables the ability to be more specific about compile targets so GEMINI.md does not get generated unless it's actually requested.

Fixes #1019

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [X] Tested locally
- [X] All existing tests pass
- [X] Added tests for new functionality (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `apm marketplace build` now respects `GITHUB_HOST` for GitHub Enterprise repos -- ref resolution, token lookup, and metadata fetch all use the configured host instead of hardcoded `github.com`. `git ls-remote` is authenticated so private repos work without separate credential setup. (#1008)
 - `apm marketplace build` now accepts multiple Git URL forms (GitHub, GHES, GitLab, Bitbucket, ADO, SSH) for `type: url` parsing via `DependencyReference.parse()`. Host resolution is still driven by `GITHUB_HOST`, so non-`github.com` hosts require `GITHUB_HOST` to be set accordingly. (#1008)
 - **ADO Entra ID auth path no longer silently fails.** Bearer tokens from `az account get-access-token` are now correctly plumbed through validation (auth scheme, git env). Auth failures raise a typed `AuthenticationError` with an actionable 4-case diagnostic instead of the ambiguous "not accessible or doesn't exist" message. `apm install --update` runs a pre-flight auth check before modifying any files -- on failure it aborts with "No files were modified". (#1015)
+- Correct targeting of compiled artifacts so GEMINI.md is only created if requested (#1019)
 
 ## [0.10.0] - 2026-04-27
 
--- a/src/apm_cli/commands/compile/cli.py
+++ b/src/apm_cli/commands/compile/cli.py
@@ -164,31 +164,42 @@ def _get_validation_suggestion(error_msg):
 
 
 def _resolve_compile_target(target):
-    """Map CLI target input to compiler-understood target string.
+    """Map CLI target input to a compiler-understood target.
 
-    The compiler understands ``"vscode"``, ``"claude"``, ``"gemini"``,
-    and ``"all"``.  Multi-target lists are mapped to the narrowest
-    equivalent; any combination of two or more distinct compiler
-    families collapses to ``"all"``.
+    The compiler understands single-string targets (``"vscode"``,
+    ``"claude"``, ``"gemini"``, ``"all"``) and ``frozenset`` targets
+    containing compiler-family names (``"agents"``, ``"claude"``,
+    ``"gemini"``).
+
+    Multi-target lists are mapped to the narrowest representation:
+    a single string when only one compiler family is needed, or a
+    ``frozenset`` of families when multiple are needed.  This avoids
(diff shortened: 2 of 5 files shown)


danielmeppiel:
Pushed one follow-up commit (fb314d3) onto this branch addressing the three optional nits from the Review Panel, @tillig — feel free to revert/squash/amend as you see fit:

1. `CompileTargetType: Union[str, frozenset]` → `Union[str, frozenset[str]]` (`target_detection.py`)
2. Extracted `REASON_NO_TARGET_FOLDER` constant; replaced brittle `"no target" in detection_reason` substring match with equality comparison. **User-facing log message unchanged.**
3. Added `_KNOWN_COMPILE_FAMILIES` validation in `AgentsCompiler` — direct API callers passing `frozenset({"bogus"})` now fail explicitly instead of silently no-op'ing. Plus regression test `test_unknown_frozenset_target_family_returns_failure`.

Notably **declined the `assert` suggestion** in favor of an explicit validation branch — `assert` is stripped under `python -O` and would surface as a generic `Compilation failed: ...`, strictly worse UX than the current explicit unknown-target error path. Used the same pattern that exists for the non-frozenset case ~10 lines below.

Doc sync (the required action): verified as a no-op — `docs/src/content/docs/reference/cli-commands.md:1712` and `packages/apm-guide/.apm/skills/apm-usage/commands.md:28` already document the post-fix multi-target behavior.

Tests: `pytest tests/unit/compilation/ tests/unit/core/ tests/unit/commands/ -q` → 806 passed in 5.50s.

danielmeppiel:
Rebased on `main` (resolved trivial CHANGELOG.md conflict — kept all four Fixed entries) and addressed the three Copilot reviewer findings on the follow-up commit, @tillig:

**1. Type tightening** (`target_detection.py:34`) — `Union[str, frozenset[str]]` → `Union[TargetType, frozenset[CompileFamily]]` with a new `CompileFamily = Literal["agents", "claude", "gemini"]` alias. mypy/pyright can now catch invalid families flowing into `should_compile_*`.

**2. Latent bug fix** (`cli.py:453`) — the `isinstance(target, list)` branch only fired for CLI multi-target lists. **Config-driven multi-target** (`apm.yml: target: [claude, codex]`) fell through to `get_target_description(frozenset)` and logged `Compiling for unknown target - apm.yml target`. Switched the gate to `isinstance(effective_target, frozenset)` so both paths produce the same accurate `Compiling for AGENTS.md + CLAUDE.md (...)` summary. Added two regression tests in `TestMultiTargetLogOutput`.

**3. Type contract honored** (`cli.py:425`) — `detect_target()` is typed `config_target: Optional[str]`, but `compile_config_target` could be a frozenset. Now passes `config_target=compile_config_target if isinstance(compile_config_target, str) else None` (the frozenset case is already handled by the branch above).

All amended into the existing follow-up commit (force-with-lease used; only your three commits + my one follow-up are on the branch). Tests: 812 passed in 3.55s.

Branch should now be mergeable + green for @apm-bot's next pass.

PR #1025, opened 2026-04-28
test: label-reset probe (DO NOT MERGE)

danielmeppiel:
Probe PR to validate pr-panel-label-reset.yml. Will be closed after the test. Base = refactor/review-panel-fanout so the workflow file is loaded from the feature branch.


danielmeppiel:
Cleanup: probe done. Reset workflow validated -- panel-rejected was stripped within ~13s of synchronize. See PR #1022 description for full evidence.

PR #1039, opened 2026-04-29
fix(install): map APM prompt 'input' to Claude 'arguments' front-matter

stbenjam:
## Description

Slash commands installed from APM packages did not wire `input:` parameters through to Claude Code's native `arguments:` front-matter, so users got no argument hints when invoking the command.  This is a blocker for teams rolling out APM -- every `/command` that expects arguments silently drops the hints.

Changes:
- Map `input:` (list, object-list, bare dict, single string) to `arguments:` in the compiled Claude command.
- Convert `${input:name}` body references to `$name` placeholders.
- Auto-generate `argument-hint` from input names when not explicitly set.
- Move command-generation helpers out of `ClaudeFormatter` (compilation layer) into `CommandIntegrator` (integration layer) where they belong.
- Remove orphaned `CommandGenerationResult` dataclass (no production callers remained).
- Add defense-in-depth `SecurityGate.scan_text()` to the `integrate_command()` write path.
- Document input formats and target-specific mapping in three docs pages and the apm-usage skill resource.

Fixes # (issue)

N/A

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [X] Tested locally
- [X] All existing tests pass
- [X] Added tests for new functionality (if applicable)


danielmeppiel:
Thanks for the careful review @copilot-pull-request-reviewer. Pushed `4aef3ef5` to address the two remaining gaps:

- **SecurityGate result was effectively ignored for non-critical findings** — `integrate_command` now also surfaces a warning when `verdict.has_findings` is true (not just `has_critical`), and includes `critical_count` / `warning_count` in the message.
- **Silent `except (ImportError, OSError, ValueError): pass`** — now appends a `"security scan skipped due to scan error: <exc>"` warning so install diagnostics still see something when the scanner itself fails.
- Added a unit test (`test_warning_only_findings_recorded_in_diagnostics`) covering the warning-only path with U+00AD.

The other four findings were already addressed in the current branch:
- `_extract_input_names` already filters empty/whitespace strings in both the list-of-strings and dict-keys branches.
- `_transform_prompt_to_command` is already typed as `Tuple[str, frontmatter.Post, List[str]]`.
- Warnings from `_transform_prompt_to_command` are surfaced via `diagnostics.security(...)` (or the logger fallback) at the end of `integrate_command`.
- `_extract_input_names` has `input_spec: Any` annotation.

Full unit test suite for the touched areas passes (1260 tests).

danielmeppiel:
@stbenjam I am fixing per panel, thanks a lot for the PR, I hope you won't be put off by our demanding agents. Please check the .github/agents/ folder - I'd recommend to use those personas to design and implement PRs so that the panel accepts quicker. I typically summon the panel locally with the skill and ask it to design, then implement directly (.github/skills/apm-review-panel). It already encapsulates the personas.

danielmeppiel:
Thanks for the thorough review. Pushed `55b05e40` addressing all 10 required-before-merge items from the APM Review Panel verdict above:

1. **Fail-closed on scan errors** — `SecurityGate.scan_text` `ImportError` now re-raises (no file written); `OSError`/`ValueError` are recorded as warnings with package context.
2. **YAML injection guard** — argument names are validated against `^[A-Za-z][\w-]{0,63}$` and rejected names are dropped with a warn diagnostic, so a malicious dict key like `"foo>\ninjected_key"` cannot smuggle YAML into the rendered front-matter.
3. **Severity wired correctly** — verdicts with `has_critical` now flow through `diagnostics.security(..., severity="critical")` and render red+bold.
4. **message vs detail** — security findings now use a short `message` and put counts/blocked-status in `detail`.
5. **Typing** — `diagnostics: Optional["DiagnosticCollector"]` (TYPE_CHECKING import), `hasattr` guard removed.
6. **Brace-format coverage** — added `test_single_brace_input_references_substituted` exercising the canonical `${input:name}` form alongside the existing `${{input:name}}` cases.
7. **Silent-mutation UX** — install-time `diagnostics.info()` now lists the arguments mapped per file, so the transformation is visible without `--verbose`.
8. **Docs** — `docs/src/content/docs/reference/cli-commands.md` `apm install` section now documents the input → arguments mapping, allowlist, and brace-format rewrite.
9. **CHANGELOG: Fixed → Added** — moved.
10. **Outcome-first** — rewritten to lead with "Slash commands installed from APM packages now surface argument hints in Claude Code".

`tests/unit/integration/ tests/unit/compilation/ tests/unit/test_install_scanning.py` -> 1269 passed.

danielmeppiel (approved):

stbenjam:
Oh, thank you for fixing! I do run the panel locally but it didn't flag this.  I don't have access to copilot, so I use it with Opus or Gemini.

PR #1051, opened 2026-04-29
fix(shared/apm): re-shape single-artifact download to expected layout

danielmeppiel:
## TL;DR

`actions/download-artifact` (since v5) silently overrides `merge-multiple: false` and flattens contents directly into `path/` whenever **exactly one artifact** matches the pattern. Our validation step in `shared/apm.md` assumed the per-artifact subdir always exists, so any single-group consumer (the common case: one credential group, one packed bundle) fails with `missing APM bundles (group did not pack successfully): apm-<id>` even though the bundle downloaded and verified successfully.

Repro: https://github.com/microsoft/apm/actions/runs/25124365062/job/73633628506?pr=1039

## Why latent until now

#1033 fixed an earlier failure in `apm-prep` (Go-slice → JSON serialisation for `jq --argjson`). Before that fix, the agent job never reached the validation step. With prep working, every single-group PR run of `pr-review-panel` and `triage-panel` now hits this latent bug.

## Fix

Add a normalisation step before validation: when the matrix has exactly one group, re-create the expected `${prefix}apm-<id>/` subdir and move the flattened bundle file(s) into it. No-op for multi-group matrices (download-artifact already produces per-artifact subdirs in that case). Validation script unchanged.

## Reference

[`actions/download-artifact` v8.0.1 source](https://github.com/actions/download-artifact/blob/3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c/src/download-artifact.ts):

```ts
path:
  isSingleArtifactDownload ||
  inputs.mergeMultiple ||
  artifacts.length === 1   // <-- triggers the flatten
    ? resolvedPath
    : path.join(resolvedPath, artifact.name),
```

## Files

- `shared/apm.md` — new "Normalise bundle layout" step
- `pr-review-panel.lock.yml`, `triage-panel.lock.yml` — recompiled
- `actions-lock.json` — refreshed by `gh aw compile`


danielmeppiel:
Resolved panel review findings in f231989c:

**Required (Supply Chain Security):**
- Restored `github/gh-aw-actions/setup-cli@v0.71.2` entry in `.github/aw/actions-lock.json` (still referenced by `agentics-maintenance.yml`; prior recompile dropped it because pr-review-panel/triage-panel don't import it).

**Adopted nits:**
- Sanitised `group_id` with strict allowlist (`^[A-Za-z0-9_-]+$`) before interpolation into the shell path -- defence-in-depth even though `apm-prep` already produces a sanitised id (Supply Chain Security).
- Guarded `jq '.group | length'` against null with `.group // []` (Python Architect).
- Added upstream `actions/download-artifact` source link in the comment so future maintainers can verify the workaround is still needed (DevX UX).
- CHANGELOG entry under `[Unreleased]` -> `Fixed` (OSS Growth Hacker).

**Set aside (with rationale):**
- `-prune` over `! -path` (Python Architect): functionally equivalent here; `! -path` is the more readable form for a single exclusion.
- `echo '[normalise] ...'` trace lines (CLI Logging Expert): step is silent on success by design (matches surrounding apm.md steps); failures already surface via `set -euo pipefail` and the explicit `::error::` line for the security guard.
- Lock-file comments preservation (DevX UX): inherent gh-aw limitation, out of scope for this PR.

Recompiled both lock files. Pushing should auto-clear `panel-rejected`.

PR #1063, opened 2026-04-30
feat(marketplace): harden apm pack output (#1061)

### LINKED ISSUE #1061: marketplace: harden pack output for downstream marketplaces (pluginRoot subtraction + remote pass-through fields)
## Context

[github/awesome-copilot#1570](https://github.com/github/awesome-copilot/pull/1570) adopts `apm pack` for its marketplace generation (73 plugins from one `apm.yml`). Code review on that PR ([review #4202980531](https://github.com/github/awesome-copilot/pull/1570#pullrequestreview-4202980531)) surfaced two independent shortcomings in APM's marketplace builder that together prevent the PR from being a clean drop-in replacement for the bespoke generator:

1. **Local-source `pluginRoot` double-prefix** -- breaks runtime resolution.
2. **Remote-source pass-through metadata dropped** -- forces a bridge script to retain `plugins/external.json`.

Filing both as one umbrella because they share the theme "make `apm pack`'s marketplace.json output spec-equivalent to what hand-rolled generators produce today" and both are small, additive builder changes.

---

### Symptom

awesome-copilot's existing baseline (`staged` branch) emits:

```json
{ "name": "ai-team-orchestration", "source": "ai-team-orchestration", ... }
```

with `metadata.pluginRoot: "./plugins"`. A spec-compliant consumer joins the two as `./plugins/ai-team-orchestration`.

`apm pack` today emits:

```json
{ "name": "ai-team-orchestration", "source": "./plugins/ai-team-orchestration", ... }
```

against the same `pluginRoot`. The consumer joins to `./plugins/./plugins/ai-team-orchestration` -- a path that does not exist. This is a behavior break, not a cosmetic delta.

### Root cause

`LOCAL_SOURCE_RE = re.compile(r"^\./")` ([`yml_schema.py:72`](https://github.com/microsoft/apm/blob/main/src/apm_cli/marketplace/yml_schema.py#L72)) requires the leading `./` for an apm.yml `source` value to be classified as local -- so authors **must** write `source: ./plugins/foo` in apm.yml.

The builder then emits that source verbatim ([`builder.py:730`](https://github.com/microsoft/apm/blob/main/src/apm_cli/marketplace/builder.py#L730)):

```python
if is_local:
    plugin["source"] = entry.source  # never subtracts metadata.pluginRoot
```

`MarketplaceConfig.plugin_root` is parsed and re-emitted into `metadata.pluginRoot` ([`models.py:85,199`](https://github.com/microsoft/apm/blob/main/src/apm_cli/marketplace/models.py#L85)), but never consulted on the emit path. The result is structurally inconsistent output: APM declares `pluginRoot` but doesn't honor it in its own sources.

### Proposed fix

In the local-source emit branch of `builder.py`, when `metadata.pluginRoot` is set:

```python
if is_local:
    source_value = entry.source
    if config.plugin_root:
        source_value = _strip_plugin_root(source_value, config.plugin_root)
    plugin["source"] = source_value
```

Where `_strip_plugin_root("./plugins/foo", "./plugins")` returns `"foo"`. Idempotent (no leading `./` -> no strip), pure path arithmetic, no I/O.

### Why now

Workaround in awesome-copilot is a 5-line post-process in `eng/merge-external-plugins.mjs`. Acceptable as a stopgap but it couples a downstream repo's bridge to APM's emit format. Native fix removes that coupling permanently.

### Acceptance

- [ ] When `metadata.pluginRoot` is set, local sources in `marketplace.json` are emitted with the prefix stripped.
- [ ] `metadata.pluginRoot` continues to round-trip unchanged.
- [ ] No-op when `pluginRoot` is unset (today's behavior preserved).
- [ ] Unit tests: (a) prefix stripped when `pluginRoot` matches, (b) pass-through when no match, (c) no `pluginRoot` -> verbatim emission.

---

### Symptom

awesome-copilot's `plugins/external.json` entries publish four fields that vanish through `apm pack`:

```json
{
  "name": "azure",
  "author": { "name": "Microsoft", "url": "https://www.microsoft.com" },
  "keywords": ["azure", "cloud", ...],
  "license": "MIT",
  "repository": "https://github.com/microsoft/github-copilot-for-azure",
  "source": { ... }
}
```

`apm pack` today emits only `description`/`version`/`tags` as pass-through; the four fields above are silently dropped.

### Root cause

`_PACKAGE_ENTRY_KEYS` in [`yml_schema.py:88-99`](https://github.com/microsoft/apm/blob/main/src/apm_cli/marketplace/yml_schema.py#L88) does not list them. Even if a maintainer writes them in `apm.yml`, the parser rejects unknown keys.

A second smaller issue stacked on the same surface: for remote entries, `description`/`version` come from the remote repo's own `apm.yml` only. Third-party plugin repos (`microsoft/Dataverse-skills`, `dotnet/skills`, `figma/mcp-server-guide`) don't ship one, so those fields render empty even when the maintainer supplied values in awesome-copilot's `apm.yml`.

### Proposed fix

**Schema (additive):** extend `_PACKAGE_ENTRY_KEYS` to permit `author`, `keywords`, `license`, `repository`. Validate types (object / array / string / string).

**Builder:** for remote entries:
- emit the four new fields verbatim when present;
- when `entry.description` / `entry.version` are set in apm.yml, prefer them over the remote-fetched values (today the remote fetch wins).

### Out of scope here

The source-object key delta (`{type, repository, ref, commit}` vs legacy `{source, repo, path}`) is a downstream consumer concern. APM should keep emitting its own canonical shape; consumers adapt at publish time. Filing as a separate decision if it ever becomes a real blocker.

### Acceptance

- [ ] `marketplace.packages[]` accepts `author`, `keywords`, `license`, `repository`.
- [ ] Remote-source builder emits those fields verbatim when present on the entry.
- [ ] When `entry.description` / `entry.version` are set on a remote entry, they override the remote fetch.
- [ ] Unit tests: (a) all four pass-through fields round-trip, (b) maintainer description overrides remote, (c) absent fields produce identical output to today.

---

## Why ship both together

Both are small, additive, zero-risk-when-unset builder changes on the same code path. Together they let `awesome-copilot` collapse:
- `eng/merge-external-plugins.mjs` (drops with Finding 2 closure)
- the post-process step that strips `pluginRoot` (drops with Finding 1 closure)
- `plugins/external.json` (drops with Finding 2 closure -- moves into `apm.yml`)

End state: `apm pack` becomes a true drop-in replacement for the bespoke generator -- which is the proof point this work was designed to deliver.

## Tracking

- Downstream PR consuming both: [github/awesome-copilot#1570](https://github.com/github/awesome-copilot/pull/1570)
- Downstream issue: [github/awesome-copilot#1569](https://github.com/github/awesome-copilot/issues/1569)

danielmeppiel:
## TL;DR

Hardens `apm pack` marketplace builder to fix double-prefix bugs, add new pass-through fields, and implement curator-wins override semantics.

## Problem

Local source paths in marketplace.json were emitted with the full path including `pluginRoot`, causing consumers to double-prepend the root when resolving. Additionally, marketplace authors lacked fields for `author`, `license`, `repository`, and had no way to override remote-fetched `description`/`version` values.

### pluginRoot subtraction (fixes double-prefix)
- Local paths are now emitted **relative to** `metadata.pluginRoot`
- `./plugins/my-tool` with `pluginRoot: ./plugins` emits as `./my-tool`
- Mismatch produces a build warning (W1); path emitted verbatim as fallback

### New pass-through fields
- `author`, `license`, `repository` — string fields emitted into marketplace.json
- `keywords` — alias for `tags`, merged and deduplicated

### Curator-wins override semantics
- For remote entries: entry-level `description` wins over remote-fetched
- For remote entries: entry-level `version` wins only when it's a fixed version (not a semver range like `^2.0.0`)
- Verbose diagnostics (I1/I2) report overrides when enabled

### Security guards
- **S1**: `pluginRoot` validated against path traversal at schema load
- **S2**: Post-subtraction output rejected if absolute or contains `..`
- **S3**: `author`/`license`/`repository` must be strings
- **S4**: Tags array capped at 50 items, 100 chars each

### Diagnostics
- New `BuildDiagnostic(level, message)` frozen dataclass
- Catalog: W1 (mismatch warning), X1 (subtraction error), I1/I2 (override verbose), I3 (subtraction verbose)
- Flows into `BuildReport.diagnostics` tuple

## Test coverage

38 new tests across 5 files:
- `test_builder_security.py` (10): S1-S4 guards, override precedence
- `test_builder_logging.py` (11): T1-T11 diagnostic catalog
- `test_yml_schema.py` (+7): new fields, keywords merge, type/length validation
- `test_builder.py` (+5): remote override, field emission, serialization order
- `test_local_path_compose.py` (+5): subtraction happy path, nested, unset, mismatch

Full suite: **6805 tests pass**.

## How to test

```bash
uv run pytest tests/unit/marketplace/ -x
# or full suite:
uv run pytest tests/unit tests/test_console.py -x
```

Real-world validation:
```bash
uv run python scripts/validate-1061-real-world.py
```

Closes #1061


danielmeppiel:
## Real-world validation -- Issue #1061

Patched `apm pack` (this PR's branch `fix/1061-marketplace-pack-hardening`)
exercised against two real consumers. Both findings prove out end-to-end.

### Summary

| Repo / scenario                       | Plugins | Finding 1 (pluginRoot) | Finding 2 (pass-through + override) | Result |
|---------------------------------------|--------:|------------------------|-------------------------------------|--------|
| `github/awesome-copilot` (PR #1570 apm.yml + `staged` baseline) | 64 local | 0 double-prefix entries; 64/64 source paths semantically match the bridge baseline | n/a (entries are local) | PASS |
| `microsoft/azure-skills` as remote dep (synthetic curator apm.yml) | 1 remote | n/a | `author`, `license`, `repository` emitted; `keywords` merged into `tags`; curator `description`/`version` overrode remote-fetched values; source pinned to SHA | PASS |

---

### Finding 1 -- pluginRoot subtraction (github/awesome-copilot)

Inputs:

- apm.yml: `https://github.com/danielmeppiel/awesome-copilot/blob/feat/apm-pack-marketplace/apm.yml` (PR #1570 head, 64 local packages, `metadata.pluginRoot: ./plugins`).
- Baseline: `https://github.com/github/awesome-copilot/blob/staged/.github/plugin/marketplace.json` (current bridge output, 73 plugins; 64 local + 9 remote-via-`plugins/external.json`).

Command:

```
apm pack --offline --marketplace-output ./out.json
```

Programmatic checks against `out.json`:

```
Patched plugins:                64
Baseline plugins (local subset): 64
pluginRoot (patched):           ./plugins
pluginRoot (baseline):          ./plugins
Double-prefix entries:           0   <- BUG GONE
Source path shapes (patched):
   64  OK: starts with ./
Semantic source mismatches vs baseline: 0
```

Sample patched entry:

```json
{
  "name": "ai-team-orchestration",
  "description": "Bootstrap and run a multi-agent AI development team ...",
  "version": "1.0.0",
  "source": "./ai-team-orchestration"
}
```

Compared with what `apm pack` emitted **before** this PR (verbatim from
apm.yml input):

```json
{ "source": "./plugins/ai-team-orchestration" }   // pre-fix -- consumer joins to ./plugins/./plugins/ai-team-orchestration
{ "source": "./ai-team-orchestration" }           // post-fix -- consumer joins to ./plugins/ai-team-orchestration  (correct)
```

The 9-entry delta vs the 73-plugin baseline is the remote external set still
declared in `plugins/external.json`. Migrating those into `apm.yml` is the
exact follow-up Finding 2 enables (see next section).

**Bridge redundancy:** awesome-copilot's `eng/merge-external-plugins.mjs` exists
to (a) subtract `pluginRoot` from sources and (b) splice in `external.json`
metadata (author/keywords/license/repository). With this PR, both responsibilities
move into APM:
- (a) is done natively (verified above).
- (b) is unblocked by the new pass-through fields (verified below).

The bridge becomes a no-op once the externals are inlined into apm.yml, and
`plugins/external.json` can be deleted.

---

### Finding 2 -- pass-through metadata + curator-wins override (microsoft/azure-skills as remote)

azure-skills is a *consumer* package (no `marketplace:` block of its own), so
the realistic exercise is to declare it as a remote entry from a curator's
apm.yml and confirm the new pass-through fields are honored end-to-end.

Synthetic curator apm.yml (pinned to a real commit SHA on `microsoft/azure-skills/main`):

```yaml
name: validation-1061-finding2
version: 1.0.0
description: Synthetic marketplace exercising Finding 2.
marketplace:
  owner:
    name: APM 1061 Validation
    email: validation@example.com
  metadata:
    description: Validation marketplace for issue 1061 Finding 2.
    version: 1.0.0
  packages:
    - name: azure
      source: microsoft/azure-skills
      ref: 5544f427264d972b0e406d0b11a8ac31db9b18dc
      description: Microsoft Azure MCP and Skills integration -- curator override of remote description.
      version: 2.0.0
      author: "Microsoft <https://www.microsoft.com>"
      license: MIT
      repository: https://github.com/microsoft/azure-skills
      keywords:
        - azure
        - cloud
        - mcp
        - skills
```

Remote (microsoft/azure-skills/apm.yml at HEAD) declares its own
`description`: `"Microsoft Azure MCP and Skills integration for cloud resource
management, deployments, and Azure services. Manage your Azure infrastructure,
monitor applications, and deploy resources directly from your development
environment."`

Patched output (`marketplace.json`):

```json
{
  "name": "azure",
  "description": "Microsoft Azure MCP and Skills integration -- curator override of remote description.",
  "version": "2.0.0",
  "author": "Microsoft <https://www.microsoft.com>",
  "license": "MIT",
  "repository": "https://github.com/microsoft/azure-skills",
  "tags": ["azure", "cloud", "mcp", "skills"],
  "source": {
    "type": "github",
    "repository": "microsoft/azure-skills",
    "ref": "5544f427264d972b0e406d0b11a8ac31db9b18dc",
    "commit": "5544f427264d972b0e406d0b11a8ac31db9b18dc"
  }
}
```

Verifies all five sub-acceptance items at once:

1. `author`, `license`, `repository` emitted (previously: silently dropped).
2. `keywords` accepted as input alias and merged into `tags` (no separate `keywords` field in output -- per CEO Q1 decision).
3. Curator-supplied `description` overrode the remote's own (curator-wins).
4. Curator-supplied `version` (`2.0.0`) emitted as display metadata; lockfile pins by commit SHA, so resolution integrity is unchanged.
5. Output `source` block carries the resolved commit SHA (supply-chain integrity preserved).

---

### Shipping UX scenarios proven

| Scenario                                                                                  | Before this PR                                                          | After this PR                                                                  |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------|--------------------------------------------------------------------------------|
| 73-plugin marketplace.json from a single apm.yml (awesome-copilot)                        | Required `eng/merge-external-plugins.mjs` bridge + `plugins/external.json` to fix double-prefix and add metadata | `apm pack` emits spec-equivalent local sources directly; 0 double-prefix entries |
| Curator overrides description/version for a remote package whose upstream has different/empty values | Remote always won; curator's apm.yml values silently ignored            | Curator wins when truthy; verbose `--verbose` log surfaces the override         |
| Curator publishes pass-through metadata (author, license, repository, keywords) on remote entries | Schema rejected unknown keys at parse                                    | All four accepted; keywords merges into tags; emitted verbatim in marketplace.json |
| Mixed-root marketplace authoring (some plugins outside pluginRoot)                        | Verbatim emit -- silently broken on join                                | Per-entry warning, verbatim fallback, exit 0 (per CEO Q2: warn-always)          |
| Misconfigured `pluginRoot: "../escape"`                                                   | Accepted at parse, garbage propagates downstream                         | Rejected at schema load via `validate_path_segments` (S1 guard)                 |

---

### Reproducer

```
git clone --depth=1 https://github.com/danielmeppiel/awesome-copilot.git \
  -b feat/apm-pack-marketplace ac
cd ac
apm pack --offline --marketplace-output /tmp/out.json   # 64 plugins, 0 double-prefix

# Finding 2 synthetic:
mkdir /tmp/syn && cd /tmp/syn
# write the apm.yml shown above (pinned SHA)
apm pack --marketplace-output /tmp/syn-out.json         # 1 plugin with all new fields
```

Patched binary used: `fix/1061-marketplace-pack-hardening` @ HEAD
(`/Users/danielmeppiel/Repos/awd-cli-1061/.venv/bin/apm`).

Both repos cloned shallow over HTTPS (`--depth=1 --single-branch`) per the
supply-chain review's safety bar.

danielmeppiel:
# [+] Schema conformance + shipping scenarios for #1061

> Follow-up to [the first validation comment](https://github.com/microsoft/apm/pull/1063#issuecomment-4351144369).
> Validates the patched `apm pack` output against the **official Anthropic
> Claude Code schemas** and documents how vendors with multiple plugins should
> ship them.

## TL;DR

- [+] Patched `apm pack` output now validates against the **official Anthropic
  schemas** published on SchemaStore -- 0 errors on both real-world and
  synthetic fixtures.
- [+] `author` is normalized to the spec-compliant object form
  (`{name, email?, url?}`).
- [+] Remote `source` is emitted in the canonical `{source, repo|url, ref?, sha?}`
  shape; subdir entries automatically use the `git-subdir` source type.
- [+] Two shipping models are supported and documented below: **Scenario A**
  (multi-plugin repo) and **Scenario B** (single-plugin repo).
- [i] `apm pack` does **not** generate per-plugin `.claude-plugin/plugin.json`
  files. Authors hand-write those (or rely on `strict: false` and let the
  marketplace entry be authoritative). `apm pack` aggregates plugin
  references into one `marketplace.json`.

## Problem (WHY)

The Claude Code plugin system is governed by two JSON schemas:

| Schema | URL | Governs |
|---|---|---|
| Plugin manifest | https://json.schemastore.org/claude-code-plugin-manifest.json | The `plugin.json` shipped inside each plugin |
| Marketplace | https://json.schemastore.org/claude-code-marketplace.json | The `.claude-plugin/marketplace.json` aggregator |

Per the [marketplace docs](https://code.claude.com/docs/en/plugin-marketplaces#marketplace-schema):

> Marketplace metadata files use this JSON Schema...

and the [plugin-reference docs](https://code.claude.com/docs/en/plugins-reference#plugin-manifest-schema):

> The plugin manifest defines metadata and capabilities for your plugin.

Before this PR, two of `apm pack`'s emit branches diverged from those schemas:

1. **`author`** was emitted as a free-form string. The schema requires an
   object with a required `name` field (and optional `email` / `url`).
2. **Remote `source`** was emitted as `{type, repository, commit}`, but the
   schema's `source` discriminator key is literally named `source`, the repo
   field is `repo` (not `repository`), and the commit field is `sha` (40-char
   lowercase hex). Subdir variants must use the `git-subdir` source type
   with `url` + `path` -- not `github` + `path`.

These schema gaps blocked downstream Claude Code clients from consuming
APM-generated marketplaces.

## Approach (WHAT)

1. **Normalize `author` at parse time** to the spec object form. Curators
   may still write the convenience string form (`author: "Microsoft"`) -- the
   parser promotes it to `{"name": "Microsoft"}`. The object form is also
   accepted with strict key whitelisting (`name`, `email`, `url`) and
   `name` is required.
2. **Rewrite the remote `source` emit** to produce the canonical Anthropic
   shape:
   - Plain remote: `{source: "github", repo: "owner/name", ref?, sha?}`
   - Subdir remote: `{source: "git-subdir", url: "owner/name", path: "sub/dir", ref?, sha?}`
3. **Vendor the official schemas** under `tests/fixtures/schemas/` and add a
   hermetic `test_schema_conformance.py` that runs `Draft7Validator` on
   freshly composed `marketplace.json` documents covering all three emit
   branches (local, remote, remote+subdir).

## Validation evidence

### Patched output: synthetic Finding 2 (azure-skills as remote dep)

Source apm.yml:

```yaml
marketplace:
  owner:
    name: APM 1061 Validation
    email: validation@example.com
  packages:
    - name: azure
      source: microsoft/azure-skills
      ref: 5544f427264d972b0e406d0b11a8ac31db9b18dc
      description: Microsoft Azure MCP and Skills integration ...
      version: 2.0.0
      author: "Microsoft <https://www.microsoft.com>"
      license: MIT
      repository: https://github.com/microsoft/azure-skills
      keywords: [azure, cloud, mcp, skills]
```

Generated `marketplace.json` plugin entry:

```json
{
  "name": "azure",
  "description": "Microsoft Azure MCP and Skills integration ...",
  "version": "2.0.0",
  "author": {
    "name": "Microsoft <https://www.microsoft.com>"
  },
  "license": "MIT",
  "repository": "https://github.com/microsoft/azure-skills",
  "tags": ["azure", "cloud", "mcp", "skills"],
  "source": {
    "source": "github",
    "repo": "microsoft/azure-skills",
    "ref": "5544f427264d972b0e406d0b11a8ac31db9b18dc",
    "sha": "5544f427264d972b0e406d0b11a8ac31db9b18dc"
  }
}
```

### Patched output: github/awesome-copilot PR #1570 (64 local plugins)

Sample plugin entry (one of 64):

```json
{
  "name": "ai-team-orchestration",
  "description": "Bootstrap and run a multi-agent AI development team ...",
  "version": "1.0.0",
  "source": "./ai-team-orchestration"
}
```

Top-level shape:

```json
{
  "name": "awesome-copilot",
  "owner": {"name": "GitHub", "email": "copilot@github.com"},
  "plugins": [ ... 64 entries ... ]
}
```

### Schema validation results

Run with `jsonschema` Draft 7 against the official SchemaStore-published
marketplace schema:

| Fixture | Plugin count | Schema errors |
|---|---|---|
| `awesome-copilot` (PR #1570 fork, 64 local plugins) | 64 | **0** |
| Synthetic Finding 2 (`microsoft/azure-skills` as remote dep) | 1 | **0** |

Equivalent Python harness used:

```python
from jsonschema import Draft7Validator
schema = json.load(open("schemas/marketplace.schema.json"))
v = Draft7Validator(schema)
for path in ["ac-patched-marketplace.json", "synthetic-marketplace.json"]:
    errs = list(v.iter_errors(json.load(open(path))))
    print(path, len(errs), "errors")
# awesome-copilot: 0 errors
# synthetic Finding 2: 0 errors
```

### Test suite

```
tests/unit/marketplace/  869 passed in 7.19s
tests/unit + tests/test_console.py  6790 passed, 1 warning, 27 subtests passed in 38.25s
```

A new `test_schema_conformance.py` (3 tests) runs the official Anthropic
schema against composed marketplace docs covering local, remote, and
remote+subdir emit branches.

## Scenario A -- Multi-plugin repo (one repo ships several plugins)

Use this when a vendor (e.g. a platform team) curates multiple distinct
plugins under one repository and wants Claude Code clients to discover them
all from a single marketplace.

```mermaid
flowchart LR
  subgraph repo["my-vendor-plugins (git repo)"]
    yml[apm.yml]
    subgraph plugins["plugins/"]
      pa[plugin-a/.claude-plugin/plugin.json]
      pb[plugin-b/.claude-plugin/plugin.json]
      pc[plugin-c/.claude-plugin/plugin.json]
    end
  end
  yml -- "apm pack" --> mkt[".claude-plugin/marketplace.json (aggregates all 3)"]
  pa -. referenced .-> mkt
  pb -. referenced .-> mkt
  pc -. referenced .-> mkt
```

Required apm.yml:

```yaml
name: my-vendor-plugins
version: 1.0.0
description: My vendor's plugin collection.

marketplace:
  owner:
    name: My Vendor
    email: support@my-vendor.example
  metadata:
    pluginRoot: ./plugins
  packages:
    - name: plugin-a
      source: ./plugins/plugin-a
    - name: plugin-b
      source: ./plugins/plugin-b
    - name: plugin-c
      source: ./plugins/plugin-c
```

Each plugin directory ships its own `.claude-plugin/plugin.json` (authored
by hand or generated by Claude Code's own scaffolder). `apm pack` does
**not** create those files; it composes `.claude-plugin/marketplace.json`
that points at each plugin via its relative `source` path. Claude Code
clients then resolve each `plugin.json` from disk.

Default `strict: true` (per the schema) means the per-plugin `plugin.json`
is the authority for `description` / `version` / `author` etc.; the
marketplace entry is purely a router. Set `strict: false` per package if
you want the marketplace entry to override.

## Scenario B -- Single-plugin repo (whole repo is one plugin)

Use this when the entire repository is a single plugin -- the most common
shape for community contributions.

```mermaid
flowchart LR
  subgraph repo["my-plugin (git repo)"]
    yml[apm.yml]
    p[.claude-plugin/plugin.json]
  end
  yml -- "apm pack" --> mkt[".claude-plugin/marketplace.json (1 entry)"]
  p -. referenced .-> mkt
```

Required apm.yml:

```yaml
name: my-plugin
version: 1.0.0
description: Does one thing well.

marketplace:
  owner:
    name: My Vendor
  packages:
    - name: my-plugin
      source: ./
      version: 1.0.0
      description: Does one thing well.
      author: My Vendor
      license: MIT
```

The repo's root `.claude-plugin/plugin.json` is authoritative. The
marketplace entry's `source: ./` tells Claude Code to look at the repo
root for `.claude-plugin/plugin.json`.

## Trade-offs

- **String -> object author normalization is silent.** A curator who writes
  `author: "Jane Doe <jane@example.com>"` ends up with
  `{"name": "Jane Doe <jane@example.com>"}`. We do not parse RFC-5322 -- if
  the curator wants structured `email`/`url`, they must write the object
  form. Documented in the new manifest-schema docs page.
- **`apm pack` does not generate per-plugin `plugin.json` files.** Plugin
  authors must hand-write them (or rely on `strict: false`). This keeps APM
  squarely in the dependency-resolver lane and avoids competing with Claude
  Code's own scaffolder.
- **`sha` field requires lowercase 40-char hex.** Curators who use
  `ref: main` (without an explicit SHA pin) get `ref: main` in the output
  and no `sha` field. APM rejects `ref: <branch>` for remote deps unless
  `--allow-head` is passed (defence against unpinned supply chain).

## How to test

```bash
# Run new schema-conformance suite
uv run --extra dev pytest tests/unit/marketplace/test_schema_conformance.py -v

# Real-world validation against awesome-copilot fork (offline)
git clone https://github.com/danielmeppiel/awesome-copilot /tmp/ac-test
cd /tmp/ac-test
git checkout add-apm-yml  # the PR #1570 branch
apm pack --offline --marketplace-output /tmp/ac-marketplace.json

# Validate against official schema
pip install jsonschema
python -c "
import json
from jsonschema import Draft7Validator
schema = json.load(open('https://json.schemastore.org/claude-code-marketplace.json'))
v = Draft7Validator(schema)
errs = list(v.iter_errors(json.load(open('/tmp/ac-marketplace.json'))))
print(f'{len(errs)} errors')
"
```

---

cc @apm-maintainers -- ready for review. The change is additive
(normalization at parse time + emit-shape correction); no
breaking API surface.

danielmeppiel:
# [+] Plugin-manifest schema conformance (follow-up)

Good question raised on the last comment: did we also validate against the
**plugin** manifest schema, not just the marketplace schema? Now we do.

## Why this matters

`apm pack` does **not** generate `plugin.json` files directly. But when a
marketplace entry uses `strict: false` (per the
[official marketplace schema](https://json.schemastore.org/claude-code-marketplace.json)),
the marketplace entry becomes authoritative for the plugin's metadata and
must therefore satisfy the
[plugin manifest schema](https://json.schemastore.org/claude-code-plugin-manifest.json)
just like a hand-written `plugin.json` would.

To prevent silent drift between APM's emit shape and the plugin-manifest
contract, every marketplace entry that APM emits is now validated against
**both** schemas.

## How

For each generated marketplace entry we strip the marketplace-only fields
(`source`, `category`, `strict`) and validate the remainder as a synthetic
`plugin.json`. This proves that any APM-generated entry could be lifted
verbatim into a `plugin.json` and would still validate.

```python
MARKETPLACE_ONLY = {"source", "category", "strict"}
synthetic = {k: v for k, v in entry.items() if k not in MARKETPLACE_ONLY}
list(plugin_validator.iter_errors(synthetic))  # must be []
```

## Results

| Fixture | Plugin entries | Marketplace-schema errors | Plugin-schema errors |
|---|---|---|---|
| awesome-copilot PR #1570 (real-world) | 64 | **0** | **0** |
| Synthetic Finding 2 (`microsoft/azure-skills`) | 1 | **0** | **0** |

```
tests/unit/marketplace/test_schema_conformance.py
  test_remote_entry_with_all_passthrough_fields_validates    PASSED
  test_local_entry_validates                                 PASSED
  test_remote_subdir_entry_uses_git_subdir_form              PASSED
  test_remote_entry_validates_as_synthetic_plugin_json       PASSED  <-- NEW
  test_local_entry_validates_as_synthetic_plugin_json        PASSED  <-- NEW
  test_minimal_entry_validates_as_synthetic_plugin_json      PASSED  <-- NEW
6 passed in 0.72s
```

The plugin-manifest schema (`claude-code-plugin.schema.json`) was already
vendored under `tests/fixtures/schemas/` in commit `d91d2018`; the
follow-up commit `593e578a` adds the three new conformance tests that
exercise it.

## Caveat

These tests validate the **shape** of an entry against the plugin-manifest
schema, not the **completeness** of a real `plugin.json` (which may also
declare `commands`, `agents`, `skills`, `mcpServers`, etc., depending on
the plugin's capabilities). Those capability sections live inside the
plugin's source tree and remain the author's responsibility -- `apm pack`
intentionally does not synthesize them.

danielmeppiel:
## Update: `apm pack --format` default flipped (BREAKING)

Folding into this PR: the canonical `apm pack` output is now the Claude Code plugin directory with a schema-conformant `plugin.json`. The previous default (legacy APM bundle layout) moves under `--format apm`.

### Correction to earlier comments

[A prior comment](https://github.com/microsoft/apm/pull/1063#issuecomment-4351273234) stated that `apm pack` does not emit `plugin.json`. That was **incorrect** -- it has been emitting `plugin.json` behind `--format plugin` for some time. With this commit, `--format plugin` becomes the **default** (no flag needed), making the plugin manifest the canonical APM publish artifact.

### New flag matrix

| Command | Output |
|---|---|
| `apm pack` | Claude Code plugin directory + schema-conformant `plugin.json` (default) |
| `apm pack --format plugin` | Same as default (explicit) |
| `apm pack --format apm` | Legacy APM bundle layout (with `apm.lock.yaml`) |

### Schema conformance: now tested end-to-end

The `plugin.json` produced by `export_plugin_bundle` is now validated against the [official Claude Code plugin manifest schema](https://json.schemastore.org/claude-code-plugin.json) by `tests/unit/test_plugin_exporter_schema.py` (6 cases: synthesized minimal, synthesized rich, authored minimal, authored rich, authored-with-stale-keys-stripped, convention-dirs-still-reachable).

A real bug was caught and fixed: `_update_plugin_json_paths` was emitting `{"agents": ["agents/"]}` and similar entries that violated the official schema (which requires array items to match `^\\.\\/.*` and the per-component file extension). The fix: drop those keys entirely. The convention directories (`agents/`, `skills/`, `commands/`, `instructions/`) are auto-discovered by Claude Code per the spec; the `agents`/`skills`/`commands`/`instructions` arrays in `plugin.json` exist only for *additional* files outside the convention layout, which APM does not emit.

### Two packing scenarios this PR documents

1. **Single plugin per repo**: `apm pack` -> `<name>-<version>/plugin.json` directory ready to drop into a Claude Code plugins folder.
2. **Multi-plugin / marketplace repo**: `apm.yml` declares a `marketplace:` block; `apm pack` ALSO emits `.claude-plugin/marketplace.json` aggregating each declared package. Curators can choose whether each entry is its own plugin (run `apm pack` per subdirectory to produce per-plugin `plugin.json` files) or rely on the marketplace entry's pass-through fields (`description`, `version`, `author`, `license`, `repository`, `keywords`) under `strict: false`.

### Companion changes in flight

- `microsoft/apm-action` PR (separate branch in `~/Repos/apm-action`): pins `--format apm` on the action's `runPackStep` so the existing restore path keeps working. Opened by a sibling subagent in parallel with this comment -- link will follow.
- Documentation updates under `docs/src/content/docs/` and `packages/apm-guide/.apm/skills/apm-usage/commands.md` (also in flight via the doc-writer agent).
- CHANGELOG: `[Unreleased]` BREAKING entry under "Changed".

### Verification

- `tests/unit/test_plugin_exporter_schema.py`: 6 new tests, all green
- Full unit suite: `6811 passed, 1 warning in 31.89s`

danielmeppiel:
## APM Review Panel Verdict: REJECT

> All five active specialists agreed the architecture and positioning are sound; one required CLI-logging fix (one-line warning when an authored plugin.json has keys silently stripped) blocks merge.

### Required before merge (1 item)

- [cli-logging-expert] Silent strip of authored plugin.json keys emits no warning to the user at `src/apm_cli/bundle/plugin_exporter.py:385`
  - Why: `_update_plugin_json_paths()` unconditionally pops `agents`/`skills`/`commands`/`instructions` from an authored manifest without logging. The established pattern in the same file (lines 347-362, 564-571) is to emit `_rich_warning` when user-authored content is modified or ignored. Silently rewriting a user's manifest violates the "Name the thing" and "Include the fix" rules: the user cannot know their keys were scrubbed or why, and cannot act on it.
  - Suggested fix: When any of the four keys are actually present in the input dict (i.e. were authored), emit a warning like: `_rich_warning("Stripped schema-invalid keys from authored plugin.json: {', '.join(stripped)} -- convention directories are auto-discovered by Claude Code")`. Accept an optional logger param to route through `CommandLogger` when available.

### Nits (9 items, skip if you want)

- [python-architect] Docstring for `_update_plugin_json_paths` over-claims `instructions` is schema-invalid at `src/apm_cli/bundle/plugin_exporter.py:367`
- [cli-logging-expert] `--format` help text uses semicolon-joined compound sentence; could be two shorter sentences for scannability at `src/apm_cli/commands/pack.py:61`
- [cli-logging-expert] CHANGELOG `Added` entry mixes em-dash `--` with trailing comma-separated clause, run-on listing 5 changes at `CHANGELOG.md:18`
- [devx-ux-expert] Silent strip of authored plugin.json keys has no runtime feedback at `src/apm_cli/bundle/plugin_exporter.py:605` (same root cause as the cli-logging required item; addressing the required item resolves this)
- [devx-ux-expert] `pack-distribute.md` devDependencies row says "Included" for APM format -- potentially misleading at `docs/src/content/docs/guides/pack-distribute.md:131`
- [supply-chain-security-expert] Vendored JSON schemas lack integrity pinning or documented refresh process at `tests/fixtures/schemas/claude-code-plugin.schema.json`
  - Suggested fix: Add a comment at the top of each schema file with the fetch URL + sha256 of the canonical source, and a `scripts/update-vendored-schemas.sh` that re-fetches and verifies the checksum (failing loudly on mismatch). Reference the script in CONTRIBUTING.md.
- [supply-chain-security-expert] Unconditional key strip may silently drop valid additional-file references from third-party plugin.json at `src/apm_cli/bundle/plugin_exporter.py:385`
  - Suggested fix: Before stripping, check each entry against the schema pattern (`^\./.*` with `.md` extension). Preserve entries that are schema-valid paths (and ensure those files are collected into the bundle). Only strip entries that look like convention-dir stubs (e.g., `agents/`, `skills/`).
- [oss-growth-hacker] README line 67 still describes `apm pack` as "a zipped package or a standalone plugin" -- missed top-of-funnel positioning win at `README.md:67`
  - Suggested fix: Change to: `**[Pack & distribute](...)** -- `apm pack` produces a Claude Code plugin directory by default; `--format apm` preserves the legacy bundle for CI restore workflows`
- [oss-growth-hacker] CHANGELOG BREAKING entry leads defensively ("must add `--format apm`") rather than offensively ("Claude Code is now the canonical target") at `CHANGELOG.md:11`
  - Suggested fix: Restructure: lead with the value ("`apm pack` now produces a schema-conformant Claude Code plugin directory with zero flags"), then add migration as a secondary note.

### CEO arbitration

The CLI-Logging expert's REQUIRED finding stands. APM's operating principle #1 is unambiguous: silent breaking changes are not allowed. Silently stripping keys from an authored `plugin.json` is a silent behavioral change to the user's own artifact -- the user typed those keys, ran pack, and got a different manifest with no terminal feedback explaining why. The established pattern in the same file (lines 347-362, 564-571) already emits `_rich_warning` when user content is modified or ignored. Adding one conditional warning when authored keys are actually present in the input dict is a trivial, non-controversial fix that brings this code path into alignment with the file's own conventions. The fix is one line plus a format string; it should not delay this PR.

Strategically, this PR is exactly where APM needs to be. Flipping the default to plugin-native output is a positioning inflection: APM becomes the tool that produces Claude Code plugins with zero ceremony. The breaking change is well-communicated (CHANGELOG entry, migration is one flag), the schema-conformance posture is defensible, and the 6 new tests lock the contract. The supply-chain nit about unconditional strip of valid additional-file references is a real correctness gap worth a follow-up issue (it affects re-export of upstream packages that legitimately reference root-level `.md` files outside convention dirs), but it does not block this PR because the current behavior is strictly more conservative -- it removes references that would otherwise point to files not copied into the output. A follow-up to handle that edge case with a warning + copy is appropriate.

The python-architect's docstring nit and the devx-ux doc-table concern are both valid polish items that can land in this PR or immediately after. Neither blocks.

**Dissent resolved:** CLI-Logging classified the silent-strip-without-warning as REQUIRED; DevX UX classified the same root issue as NIT (reasoning: docs mention the behavior). I side with CLI-Logging. Documentation is not a substitute for terminal feedback at the moment of action -- a user running `apm pack` should not need to have pre-read `pack-distribute.md` line 172 to understand why their manifest was rewritten. The project's own CLI design guidelines (`cli.instructions.md`) and the established warning pattern in `plugin_exporter.py` both require visible feedback when user-authored content is silently modified. This is a one-line fix that should not be controversial.

**Growth/positioning note:** This is a genuine positioning inflection point worth a launch beat. The one-liner for social/release announcement: "`apm pack` now produces a Claude Code plugin directory by default -- zero extra flags, schema-validated `plugin.json`, convention dirs auto-discovered." The README bullet at line 67 should be updated in this PR or an immediate follow-up so the top-of-funnel reflects the new default before any announcement lands. Consider a 15-second GIF showing `apm init --plugin demo && apm pack` producing a valid plugin dir.

---

<details>
<summary>Per-persona findings (full)</summary>

#### Python Architect

**OO / class diagram**

```mermaid
classDiagram
    direction LR
    class BuildOptions {
      <<ValueObject>>
      +project_root Path
      +apm_yml_path Path
      +bundle_format str
      +bundle_target Any
    }
    class BuildOrchestrator {
      <<Facade>>
      +run(options, logger) BuildResult
    }
    class BundleProducer {
      <<ConcreteProducer>>
      +produce(options, logger) ProducerResult
    }
    class ArtifactProducer {
      <<Protocol>>
      +kind OutputKind
      +produce(options, logger) ProducerResult
    }
    class export_plugin_bundle {
      <<IOBoundary>>
      +export_plugin_bundle(project_root, output_dir, ...) PackResult
    }
    class _update_plugin_json_paths {
      <<Pure>>
      +_update_plugin_json_paths(plugin_json, output_files) dict
    }
    class _find_or_synthesize_plugin_json {
      <<IOBoundary>>
      +_find_or_synthesize_plugin_json(root, yml, logger) dict
    }
    class PackResult {
      <<ValueObject>>
      +bundle_path Path
      +files list
    }
    BuildOrchestrator *-- ArtifactProducer : delegates
    BundleProducer ..|> ArtifactProducer : implements
    BundleProducer ..> export_plugin_bundle : calls when format=plugin
    export_plugin_bundle ..> _find_or_synthesize_plugin_json : step 3
    export_plugin_bundle ..> _update_plugin_json_paths : step 14
    export_plugin_bundle ..> PackResult : returns
    BuildOrchestrator ..> BuildOptions : reads
    BundleProducer ..> BuildOptions : reads bundle_format
```

**Execution flow diagram**

```mermaid
flowchart TD
    A["apm pack (commands/pack.py)"] -->|default --format plugin| B["BuildOrchestrator.run()"]
    B --> C{detect_outputs apm.yml}
    C -->|dependencies block| D["BundleProducer.produce()"]
    C -->|marketplace block| M["MarketplaceProducer.produce()"]
    D -->|bundle_format == plugin| E["export_plugin_bundle()"]
    E --> F["LockFile.read(lockfile_path)"]
    F --> G["APMPackage.from_apm_yml()"]
    G --> H["_find_or_synthesize_plugin_json()"]
    H --> I{plugin.json exists on disk?}
    I -->|yes| J["Read + parse authored plugin.json"]
    I -->|no| K["Synthesize from apm.yml metadata"]
    J --> L["Collect deps + root components"]
    K --> L
    L --> N["_update_plugin_json_paths() -- strips schema keys"]
    N --> R["Write plugin.json to bundle_dir"]
    R --> S["Return PackResult"]
```

**Design patterns**
- Used in this PR: Protocol (`ArtifactProducer`) -- `BuildOrchestrator` delegates to any producer implementing the Protocol; `BundleProducer` is the concrete implementation whose internal dispatch path changed.
- Used in this PR: ValueObject (`BuildOptions`, `PackResult`) -- frozen-ish dataclasses flow config and results between layers.
- Pragmatic suggestion: none -- the current shape is the simplest correct design at this scope.

**Findings**
- (nit) Docstring for `_update_plugin_json_paths` over-claims `instructions` is schema-invalid at `src/apm_cli/bundle/plugin_exporter.py:367` -- the vendored schema does not define `instructions` at top-level (and there is no `additionalProperties: false`), so stripping is defensive hygiene rather than required by the pattern constraint. One-word docstring tweak: "unrecognized by the schema" instead of "invalid".

#### CLI Logging Expert

- (required) Silent strip of authored plugin.json keys emits no warning to the user at `src/apm_cli/bundle/plugin_exporter.py:385`. See Required section above for full rationale and suggested fix.
- (nit) Help text for `--format` uses a semicolon-joined compound sentence; could be two shorter sentences for scannability at `src/apm_cli/commands/pack.py:61`.
- (nit) CHANGELOG `Added` line is run-on listing 5 changes mid-sentence at `CHANGELOG.md:18`.

#### DevX UX Expert

- (nit) Silent strip of authored plugin.json keys has no runtime feedback at `src/apm_cli/bundle/plugin_exporter.py:605`. Same root cause as the cli-logging required item -- addressing that item resolves this.
- (nit) `pack-distribute.md` devDependencies row says "Included (full install layout)" for APM format -- potentially misleading at `docs/src/content/docs/guides/pack-distribute.md:131`. If intentional, deserves a sentence of explanation; if a doc error, should say "Excluded" like plugin format.

- (nit) Vendored JSON schemas lack integrity pinning or documented refresh process at `tests/fixtures/schemas/claude-code-plugin.schema.json` (3671 lines vendored from `json.schemastore.org` with no SHA, no refresh script, no CONTRIBUTING.md guidance). See Nits section above for suggested fix.
- (nit) Unconditional key strip may silently drop valid additional-file references from third-party `plugin.json` at `src/apm_cli/bundle/plugin_exporter.py:385`. The official schema defines these arrays as `^\./.*\.md` paths to additional files outside convention dirs; the strip pops them unconditionally and `_collect_root_plugin_components` does not walk root-level `.md` files. Correctness gap with supply-chain implications when re-exporting upstream packages. See Nits section above for suggested fix.

#### Auth Expert

Inactive -- PR touches marketplace pack logic, plugin.json path stripping, docs, tests, and CHANGELOG; no auth/token/credential/host files modified.

#### OSS Growth Hacker

- (nit) README.md line 67 still describes `apm pack` as "a zipped package or a standalone plugin" -- missed top-of-funnel positioning win at `README.md:67`. See Nits section above for suggested rewording.
- (nit) CHANGELOG BREAKING entry leads defensively rather than offensively at `CHANGELOG.md:11`. See Nits section above for suggested restructure.

Side-channel growth signal echoed by the CEO above.

</details>

<sub>Verdict computed deterministically: 1 required finding across 5 active panelists. APPROVE iff N == 0. Push a new commit to clear this verdict label automatically.</sub>

danielmeppiel:
## Panel findings applied

PR HEAD: `19b42af9`. All required findings + most nits from the [apm-review-panel comment](https://github.com/microsoft/apm/pull/1063#issuecomment-4351980514) addressed.

**Required (blocks-merge): DONE**
- [x] `_update_plugin_json_paths()` now emits `_rich_warning` (or routes through a supplied logger) when it strips schema-invalid `agents`/`skills`/`commands`/`instructions` keys from an *authored* `plugin.json`. Synthesized manifests stay silent. Two new unit tests pin the contract.

**Nits applied:**
- [x] Docstring reword: `instructions` is "unrecognized by the schema" (it's not defined at all), not "invalid".
- [x] `--format` help text split into two sentences.
- [x] CHANGELOG `Changed` entry leads with value (zero-flag plugin output, schema-validated `plugin.json`, convention dirs auto-discovered) before migration.
- [x] CHANGELOG `Added` "pack hardening" bullet split for scannability.
- [x] DevX nit on silent-strip resolved by the required cli-logging fix.

**Nits intentionally skipped (with rationale):**
- `pack-distribute.md:131` "devDependencies Included (full install layout)" for `--format apm` -- **verified intent, not a bug**: `is_dev` filtering only runs in `plugin_exporter.py`; `--format apm` ships the full install layout including dev deps so `apm-action`'s restore mode gets the same tree it would have post-install. Plugin format excludes them because plugins are runtime artifacts.
- `README.md:67` rewrite -- repo rules require user approval for README changes; flagging for a separate, owner-approved follow-up rather than slipping it into this PR.
- Vendored schema SHA pinning + refresh script -- worth doing but scope-creep relative to this PR. Filed as a follow-up consideration.
- Schema-valid `^\./*.md` preservation in `_update_plugin_json_paths` -- CEO already scoped this to a follow-up issue, not blocking.

**Companion PR:** [microsoft/apm-action#31](https://github.com/microsoft/apm-action/pull/31) pins `--format apm` so `pack: true` keeps emitting the bundle layout. Its own panel run flagged 1 required + 5 nits, [all addressed](https://github.com/microsoft/apm-action/pull/31#issuecomment-4351952231) (CHANGELOG.md added, "legacy" -> "classic" wording, longevity reassurance, apm-version pin guidance).

**Validation:** Full unit suite green -- 6813 passed (was 6811; +2 new warn-on-strip tests).

PR #991, opened 2026-04-27
chore(compile): tighten link_resolver._resolve_path input guards

mvanhorn:
Closes #841.

## Summary

Adds an early-return guard for empty / whitespace-only input in \`_resolve_path\` (Minor 1 + Minor 2 from the issue). Previously \`_resolve_path(\"\")\` and \`_resolve_path(\"   \")\` returned the base directory via \`Path(base_dir) / \"\"\`, which is semantically wrong even though the existing \`.exists()\` checks in the call sites masked the bug.

\`\`\`python
if not path or not path.strip():
    return None
\`\`\`

## Tests (Minor 3)

Adds \`TestResolvePathInputGuards\` to \`tests/unit/compilation/test_link_resolver.py\` covering each case the issue called out:

- empty string -> \`None\` (new guard)
- whitespace-only (spaces, tab, newline) -> \`None\` (new guard)
- embedded NUL byte -> must not crash; the containment check is the caller's job, so the test only locks in \"no exception\"
- POSIX backslash traversal (\`foo\\\\..\\\\..\\\\etc\\\\passwd\`) -> stays a literal segment under \`base_dir\`
- \`file://\` URI on POSIX -> treated relative, joined under \`base_dir\`
- non-existent relative target -> resolves normally (happy path)

## Verification

\`pytest tests/unit/compilation/test_link_resolver.py\` -> 23 / 23 pass (17 existing + 6 new).

PR #1065, opened 2026-04-30
fix: address Copilot review findings on #1055 and #991 (consolidated)

danielmeppiel:
## Summary

Folds the still-actionable findings from the post-merge Copilot reviews of #1055 and #991 into **one** consolidated PR (per board policy: no follow-up issue/PR sprawl; admin-merge gate now requires reviewing Copilot findings).

### From #1055 (CLAUDE_CONFIG_DIR support)
| Finding | Resolution |
|---------|------------|
| Path traversal via `..` segments | **Already mitigated** in merged code: `resolve(strict=False)` collapses `..` BEFORE `relative_to(home)`. No change needed. |
| `root_dir` contract violation (abs path stored) | **Comment-only fix.** Documented why the absolute-path fallback is safe: every downstream consumer joins as `project_root / target.root_dir` and `pathlib.Path / <abs>` evaluates to `<abs>`. Behavior is correct. |
| Test coverage gap for CLAUDE_CONFIG_DIR install/uninstall | **Already covered**: `test_scope_install_uninstall.py:382` + `test_scope_integration.py:212-279` (outside-home, traversal, whitespace, project-scope). |
| Stale docs hardcoding `~/.claude/` | **Already updated** in merged code: `guides/dependencies.md:338,346` + `integrations/ide-tool-integration.md:172`. |
| `if self.name == "claude"` extensibility | Refactor scope; not a bug. Deferred. |

### From #991 (link_resolver guards)
| Finding | Resolution |
|---------|------------|
| NUL-byte docstring claim (`Path.exists()` does not silently 'reject') | **Docstring corrected.** Test behavior unchanged. |

## Validation

- Full unit suite for affected modules:
  ```
  uv run pytest tests/unit/compilation/test_link_resolver.py \
                tests/unit/integration/test_scope_integration.py \
                tests/unit/integration/test_scope_install_uninstall.py
  61 passed in 2.43s
  ```
- `uv run --extra dev ruff check src/ tests/` -> All checks passed
- `uv run --extra dev ruff format --check` -> 598 files already formatted

### DIFF
--- a/src/apm_cli/compilation/link_resolver.py
+++ b/src/apm_cli/compilation/link_resolver.py
@@ -433,6 +433,13 @@ def _resolve_path(path: str, base_path: Path) -> Path | None:
     """
     if not path or not path.strip():
         return None
+    # NUL bytes survive ``Path()`` construction on POSIX but every downstream
+    # filesystem call (``.exists()``, ``.is_file()``, ``.read_text()``) raises
+    # ``ValueError``. Callers in this module do not catch ``ValueError`` so an
+    # unguarded NUL would abort markdown link resolution / validation. Reject
+    # at the resolver boundary instead.
+    if "\x00" in path:
+        return None
     try:
         if Path(path).is_absolute():
             return Path(path)
--- a/src/apm_cli/integration/targets.py
+++ b/src/apm_cli/integration/targets.py
@@ -235,6 +235,12 @@ class TargetProfile:
                     # Keep ``root_dir`` home-relative so cleanup prefix matching holds.
                     new_root = abs_path.relative_to(home).as_posix()
                 except ValueError:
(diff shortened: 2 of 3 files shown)


danielmeppiel:
Both Copilot findings folded in (`HEAD`):

1. **NUL byte in `_resolve_path`** -- you were right, downstream `.exists()` raises `ValueError` and callers don't catch it. Now rejected at the resolver boundary; test tightened to assert `None`.

2. **`targets.for_scope` absolute fallback** -- correct that `_deployed_path_entry` would `RuntimeError` on out-of-tree paths. Verified the path is unreachable today (user-scope CLAUDE installs don't flow through that translator), but rewrote the comment to record the constraint so a future refactor that lockfiles user-scope deploys treats absolute `root_dir` as a dynamic-root case.

PR #1068, opened 2026-04-30
fix: extend explicit UTF-8 encoding to 5 remaining open() call sites

danielmeppiel:
## Why

Windows users on cp950 / cp1252 locales hit `UnicodeDecodeError` on any `apm` operation that reads or writes UTF-8 content through the 5 modules listed below. Issue #604 covered the prompt-file path and was fixed by #607 (script_runner.py only). The original PR #635 covered the broader surface but was closed because the author did not sign CLA. Re-deriving here.

## What

Adds `encoding="utf-8"` to the remaining `open()` call sites in:

- `src/apm_cli/config.py` (5 sites)
- `src/apm_cli/marketplace/client.py` (5 sites)
- `src/apm_cli/marketplace/registry.py` (3 sites)
- `src/apm_cli/models/plugin.py` (1 site)
- `src/apm_cli/runtime/copilot_runtime.py` (1 site)

Adds round-trip unit tests per module that exercise non-ASCII content (e.g. `中文`, `開始`, `cafe`). Skipped binary-mode opens (encoding irrelevant) and `subprocess.Popen` (different code path; already passes `encoding="utf-8"`).

## Refs

- #604 (closed)
- #635 (closed-no-CLA, re-derived here)
- #607 (merged, fixed script_runner.py only)

## Validation

- `uv run ruff check src/ tests/`: clean.
- `uv run pytest tests/unit tests/test_console.py`: 6870 passed.
- 9 new round-trip tests added; manually verified they would fail on a cp1252 locale without the encoding additions (the default codec cannot decode UTF-8 multi-byte sequences such as `e4 b8 ad` for `中`).

### DIFF
--- a/src/apm_cli/config.py
+++ b/src/apm_cli/config.py
@@ -16,7 +16,7 @@ def ensure_config_exists():
         os.makedirs(CONFIG_DIR)
 
     if not os.path.exists(CONFIG_FILE):
-        with open(CONFIG_FILE, "w") as f:
+        with open(CONFIG_FILE, "w", encoding="utf-8") as f:
             json.dump({"default_client": "vscode"}, f)
 
 
@@ -32,7 +32,7 @@ def get_config():
     if _config_cache is not None:
         return _config_cache
     ensure_config_exists()
-    with open(CONFIG_FILE) as f:
+    with open(CONFIG_FILE, encoding="utf-8") as f:
         _config_cache = json.load(f)
     return _config_cache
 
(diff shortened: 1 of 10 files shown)


danielmeppiel:
Addressed Copilot review: replaced raw CJK / accented literals with \\uXXXX escapes across all 5 test files. Same runtime values (still non-ASCII), ASCII-only source as the repo encoding rule requires. 57 affected tests still pass.

PR #1086, opened 2026-04-30
test(windows): fix CLAUDE_CONFIG_DIR scope tests on Windows

danielmeppiel:
## Problem

Three Windows unit tests added in #1055 fail on `windows-latest` (CI run [25190857376](https://github.com/microsoft/apm/actions/runs/25190857376/job/73860213927)):

- `test_user_scope_with_claude_config_dir`
- `test_user_scope_outside_home_keeps_absolute`
- `test_user_scope_collapses_dotdot_segments`

All three patch `HOME` via `monkeypatch` and then exercise code that calls `Path.home()`. On Windows, `Path.home()` ignores `HOME` and reads `USERPROFILE` (with `HOMEDRIVE`+`HOMEPATH` as fallback). So `Path.home()` returned the real runner profile (`C:\Users\runneradmin`), `relative_to(home)` succeeded against `tmp_path` (which lives under `AppData\Local\Temp\...`), and the assertions on `root_dir` failed:

```
AssertionError: assert 'AppData/Loca...g/test-claude' == '.config/test-claude'
AssertionError: assert 'AppData/Loca..._k0/elsewhere' == 'C:\\Users\\r...k0\\elsewhere'
AssertionError: assert 'AppData/Loca...dotd0/outside' == 'C:\\Users\\r...otd0\\elsewhere'
```

## Fix

Add a small `_set_home()` helper in both scope test modules that, on Windows, also patches `USERPROFILE` / `HOMEDRIVE` / `HOMEPATH` so `Path.home()` returns the fake home. Use it in the three failing tests plus `test_user_scope_expands_tilde` for consistency.

Also call `.resolve()` on the expected `outside` path in `test_user_scope_outside_home_keeps_absolute` to match the source's `Path(env).expanduser().resolve(strict=False)` output (covers macOS `/var` -> `/private/var` and Windows 8.3 short-name expansion).

Test-only change. No production code touched.

## Validation

```
uv run --extra dev ruff check tests/unit/integration/test_scope_integration.py tests/unit/integration/test_scope_install_uninstall.py
uv run --extra dev ruff format --check tests/unit/integration/test_scope_integration.py tests/unit/integration/test_scope_install_uninstall.py
uv run --extra dev pytest tests/unit/integration/test_scope_integration.py tests/unit/integration/test_scope_install_uninstall.py
# 38 passed
```

### DIFF
--- a/tests/unit/integration/test_scope_install_uninstall.py
+++ b/tests/unit/integration/test_scope_install_uninstall.py
@@ -7,6 +7,7 @@ For each target x scope combination, verifies:
 - Files at wrong-scope paths are never created
 """
 
+import os
 import shutil
 import tempfile
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


danielmeppiel:
Thanks for the review. Addressed in 82cebf3e:

- **#2 (comment wording):** Updated to `Paths outside $HOME remain absolute and are resolved/normalized.`
- **#3 (explicit `strict=False`):** Adopted; mirrors `for_scope()` and documents non-existent-path behavior.
- **#1 (dedup `_set_home` into a shared fixture):** Intentionally not adopted. Three call sites total (two added here + two preexisting integration tests in `test_deps_update_e2e.py` / `test_marketplace_e2e.py` that already inline the same pattern). Extracting a fixture saves ~10 lines and adds an import + indirection without removing the existing duplication. Happy to revisit if a 4th call site appears.

Issue #24, opened 2025-11-05
Failed to clone from the private organisation repository


IUADE0UQ:
**Describe the bug**
When I'm trying to make apm install organisation/repository, and this repository is private the APM asking me user name and password. For password I tryied both type of tokens with all permissions but nothing work

**To Reproduce**
1. Create private repository in the organisation
2. Generate token with all permissions, share access for organisation for the token
3. Push the apm project in the repo
4. Try to install apm package from the repo using the username and new token

**Expected behavior**

**Environment (please complete the following information):**
 - OS: macOS
 - Python Version: [3.9.6]
 - APM-CLI Version: [e.g. 0.5.0]

**Logs**
Validating 1 package(s)...
✓ my-org/apm-workflows - accessible
Added my-org/apm-workflows to apm.yml
Updated apm.yml with 1 new package(s)
Installing dependencies from apm.yml...
Installing APM dependencies (1)...
  my-org/apm-workflows#main
❌ Failed to install my-org/apm-workflows: Failed to clone repository my-org/apm-workflows using all available methods. Please check repository access permissions and authentication setup. Last 
error: Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --depth=1 --branch=main -- https://github.com/my-org/apm-workflows /var/folders/gz/62ppwmt12fj6mskzj30msvth0000gp/T/tmpzjf4sa_1
  stderr: 'Cloning into '/var/folders/gz/62ppwmt12fj6mskzj30msvth0000gp/T/tmpzjf4sa_1'...
remote: Invalid username or token. Password authentication is not supported for Git operations.
fatal: Authentication failed for 'https://github.com/my-org/apm-workflows/'
'
Installed 0 APM dependencies
No MCP dependencies found in apm.yml

Dependencies installation complete

PR #1099, opened 2026-05-01
feat(install): apm install <local-bundle> + lockfile-embedded plugin packs (#1098)

### LINKED ISSUE #1098: Plugin bundles should embed apm.lock.yaml (supply-chain integrity + restores apm pack/unpack round-trip)
> **Updated after devx-ux review.** Original framing (just "ship the lockfile, extend unpack mapping") was too narrow. The cleaner unification: **`apm unpack` shouldn't exist as a separate verb.** Plugin bundles are pre-resolved packages; `apm install` already knows how to integrate those into a project. This issue now proposes that direction. Original problem statement preserved below for context.

## TL;DR

1. **Supply-chain**: plugin bundles ship no `apm.lock.yaml`, so they have no embedded SBOM / integrity manifest / provenance. Add it.
2. **Mental model**: a plugin bundle IS a pre-resolved package. `apm install` already knows how to route a package's flat `skills/`/`agents/`/`commands/`/`instructions/` layout to the correct target tree. Make `apm install` accept a local plugin bundle (file or directory). This is what `pip install ./foo.whl`, `cargo install --path .`, and `npm install ./pkg.tgz` already do.
3. **Deprecate `apm unpack`** (and the `apm` bundle format) over two release cycles. Single canonical bundle format = plugin. Single canonical entry point = `install`.
4. **Round-trip restored as a free side effect.**

## Background (verified empirically against `apm` HEAD `0.11.0`, commit `2b9501ab`)

- `apm pack` default flipped from `apm` to `plugin` format in 0.11.
- `plugin` bundle: `<root>/plugin.json` + flat `skills/<slug>/SKILL.md`, `agents/<name>.md`, `commands/<name>.md`, `instructions/<file>`. **No `apm.lock.yaml`** -- explicitly excluded at `src/apm_cli/bundle/plugin_exporter.py:140-155`.
- `apm` bundle: `<root>/.github/<kind>/<slug>/<file>` (target-deployed paths) + `apm.lock.yaml` whose `deployed_files` drives `apm unpack`.
- `apm unpack <plugin.tar.gz>` exits 1 with `apm.lock.yaml not found`. Default `pack` -> `unpack` round-trip is broken in 0.11+ unless the user passes `--format apm`.
- `microsoft/apm-action` PR #31 currently defends downstream by defaulting `bundle-format: apm` and rejecting plugin tarballs at restore.

## Why "ship the lockfile" alone isn't the right framing

`LockedDependency` (`src/apm_cli/deps/lockfile.py:20-50`) records `repo_url`, `resolved_commit`, `resolved_ref`, `version`, `deployed_files`, `deployed_file_hashes`, `content_hash`, `is_insecure`, `discovered_via`. It is already an SBOM + integrity manifest. Excluding it from plugin bundles means:

- **No tamper evidence** -- a plugin bundle modified post-pack is indistinguishable from the legitimate one.
- **Not reproducible** -- third parties cannot re-resolve from sources and verify they get the same artifact.
- **No machine-readable dep graph for scanners** -- CVE/license/policy tooling cannot operate on the artifact.
- **Inconsistent with project posture** -- APM enforces lockfile-frozen installs, hash verification, and content-security scanning everywhere else.

So the lockfile **must** be in plugin bundles for supply-chain reasons, full stop. (Precedent: pip `RECORD`, cargo shipping `Cargo.lock` in source crates, OCI manifests with layer digests, SLSA provenance, CycloneDX/SPDX SBOMs.)

But shipping the lockfile is **not** the way to fix unpack, because that path keeps two parallel verbs (`install` + `unpack`), two parallel bundle formats (`apm` + `plugin`), and two parallel routing-mapping codepaths. The integration pipeline that `apm install` already runs is exactly the logic unpack should be calling.

## Proposed direction: unify on `apm install`

A plugin bundle is a **pre-resolved package with an integration-ready flat layout**. The existing `apm install` pipeline already:

- Routes `skills/`, `agents/`, `commands/`, `instructions/` to the correct target tree (`.github/`, `.claude/`, `.gemini/`, `.cursor/`, ...).
- Applies the canonical naming convention.
- Honors `--target <t1,t2,...>` for multi-target deploys.
- Detects collisions via `BaseIntegrator.check_collision()` (skips locally-modified files unless `--force`).
- Supports `--dry-run`.

All of this should "just work" if `apm install` accepts a local bundle:

```bash
apm install ./bundle.tar.gz             # detect plugin.json -> treat as pre-resolved
apm install ./bundle-dir/               # same, directory form
apm install owner/repo                  # remote, full resolution (today's behavior)
```

This matches what pip/cargo/npm already do, eliminating an APM-specific verb users have to learn.

### Concrete behavior

1. **Detection**: `apm install <local-path>` checks for `plugin.json` at root. If present, treat the path as a pre-resolved package and skip dependency resolution.
2. **Lockfile**: read embedded `apm.lock.yaml` (after change 1 below). Use it for **provenance verification** (hash check on each file, integrity check on `content_hash`) -- NOT for routing. This is a **semantic inversion** of the lockfile's role inside a bundle (today: routing manifest; tomorrow: SBOM/integrity).
3. **Routing**: hand off to the same integrator pipeline `apm install owner/repo` already runs. `--target` resolves identically (auto-detect from `apm.yml` or workspace; explicit `--target a,b` for multi-target).
4. **Naming**: derive the package identifier for the `-<package>` suffix from `plugin.json:id`. Fallback: bundle dirname. Error if neither resolves. Add `--as <alias>` to override (matching the existing `owner/repo@alias` pattern).
5. **Collisions**: inherit `install`'s collision behavior (skip locally-modified files unless `--force`). Print skip count + remediation:
   ```
   [!] 3 files skipped (locally modified). Use --force to overwrite.
       .github/instructions/coding-style.md
       .claude/agents/reviewer.md
       .cursor/rules/linting.md
   ```
6. **Offline guarantee**: `apm install <local-path>` must do zero network I/O when the source is a local file. Verify the integrator codepath has no implicit network calls (link resolution, marketplace lookups). The current `apm unpack` is fully offline; that property must be preserved when migrating bundles to `install`.

### Coordinated changes required

**Change 1 -- ship `apm.lock.yaml` in plugin bundles.** Drop it from the exclusion list in `_collect_root_plugin_components` and per-dependency walker (`plugin_exporter.py:140-155`). Place at `<bundle-root>/apm.lock.yaml`. Claude Code consumers ignore it. Bundle gains a verifiable SBOM at no behavioral cost.

**Change 2 -- `apm install` accepts local bundle paths.** Detect `plugin.json` at root, route through the existing integrator pipeline, verify against embedded lockfile.

**Change 3 -- embed pack-time target metadata for round-trip fidelity.** A bundle packed with `--target copilot,claude` and re-installed in a project that auto-detects only `copilot` would silently drop the `claude` artifacts. Embed the original `--target` set in `plugin.json` (or `apm.lock.yaml`) and have `install` warn:
```
[!] Bundle was packed for [copilot, claude]. Current resolved target is [copilot].
    Use --target copilot,claude to restore the full layout.
```

**Change 4 -- deprecate `apm unpack` and the `apm` bundle format.** Suggested timeline:

| Release | `apm install <bundle>` | `apm unpack <bundle>` | `--format apm` |
|---|---|---|---|
| v0.12 | works on plugin bundles via integrator | thin alias -> `install`; deprecation notice | works; no notice |
| v0.13 | unchanged | deprecation notice | deprecation notice |
| v0.14 or v1.0 | unchanged | removed | removed |

Two release cycles is reasonable for a pre-1.0 tool.

## Footguns / blind spots called out

- **Lockfile role inversion** (today: routing; tomorrow: provenance) needs explicit documentation so reviewers and downstream tools don't assume routing semantics.
- **Air-gapped installs** must remain network-free; integrator codepath needs an audit before unifying.
- **Round-trip fidelity** requires embedding target-set metadata in the bundle (Change 3).
- **Behavior migration** for users of today's `apm unpack`: file layout was bundle-baked; with the new path it becomes target-resolved. Migration notes should call this out explicitly.

## Downstream impact

- **`microsoft/apm-action` PR #31**: defensive `bundle-format: apm` default and plugin-rejection error stay useful as defense-in-depth until v0.13. After v0.14, action can drop both and let `bundle-format` follow upstream default.
- **`gh-aw` workflows** that round-trip via `apm pack` + `apm unpack` are silently broken on 0.11+ today; fixed once Change 2 lands.
- **CI integrity gates**: workflows can verify bundle provenance from the embedded lockfile. Capability does not exist today on plugin bundles.

## Suggested labels

`type/feature`, `theme/security`, `theme/portability`, `area/package-authoring`, `status/needs-triage`

## References

- Plugin exporter exclusion list: `src/apm_cli/bundle/plugin_exporter.py:140-155`
- `LockedDependency` schema: `src/apm_cli/deps/lockfile.py:20-50`
- Empirical reproduction + downstream defense: `microsoft/apm-action#31`
- Related downstream issue: `microsoft/apm-action#24` (setup-only) -- separate but driven by the same hardening pass.

---

<details><summary>Original framing (preserved for history)</summary>

Original draft proposed extending the lockfile schema (`deployed_files: list[{bundle, deployed}]`) so `apm unpack` could route plugin bundles via lockfile mapping. devx-ux review pushed back: that keeps two parallel verbs and two parallel formats. The unified-on-`install` direction above is strictly cleaner, matches package-manager precedent (pip/cargo/npm), and the schema extension is unnecessary because the integration pipeline already knows how to route flat plugin layouts.

</details>

danielmeppiel:
# feat(install): `apm install <local-bundle>` + lockfile-embedded plugin packs

## TL;DR

`apm install <path/to/bundle>` now restores a previously packed plugin bundle (directory or `.tar.gz`) without touching the network or the dependency resolver — matching the `pip install ./wheel` / `cargo install --path` mental model. To make this safe by default, `apm pack` now embeds the project's `apm.lock.yaml` inside the bundle with a `pack.bundle_files: path -> sha256` manifest, so installs verify integrity and detect target mismatches before writing a single file. `apm unpack` is `[Deprecated]` and points at the unified entrypoint. Closes #1098.

> [!NOTE]
> Three maintainer-decision items are surfaced in **Trade-offs** below (pre-0.12 bundle handling, `--as` semantics, `integrate_package_primitives` deviation). Flagged as deferrable, not blocking.

## Problem (WHY)

- [x] **Two commands for one job.** Restoring a packed bundle required `apm unpack`; restoring a registry package required `apm install`. Same intent ("materialize this artifact onto disk"), different verbs — diverging from the package-manager convention every developer already has cached. The PROSE principle is explicit: ["Favor small, chainable primitives over monolithic frameworks."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) — but small primitives only compose if their *interface* is consistent.
- [x] **Plugin bundles shipped without a lockfile.** Since the v0.11 plugin-format flip, `apm.lock.yaml` was actively *excluded* from plugin bundles. That means the bundle on disk had no SBOM, no per-file integrity manifest, and no provenance trail — every downstream `unpack` was an unverified file copy.
- [x] **No path-traversal defense at unpack/install time.** A crafted bundle entry (key like `../../../etc/passwd`) would happily be deployed by the existing copy loop. There was no `validate_path_segments` guard at the install boundary.
- [!] **`apm install` had a silent footgun.** Passing a path that *looked* like a bundle (`./bundle.tar.gz`) fell through to the registry resolver, which would try to clone it as `org/repo`. The error surfaced was about Git, not the user's actual intent.

## Approach (WHAT)

| # | Change |
|---|--------|
| 1 | `apm install <ARG>` early-detects local bundles when `Path(ARG).exists()` and routes to a new `install_local_bundle()` orchestrator. Registry path unchanged. |
| 2 | `apm pack` (plugin format, the default) now embeds `apm.lock.yaml` at the bundle root with new `pack.bundle_files` (sha256 manifest) and `pack.target` fields. |
| 3 | New `src/apm_cli/bundle/local_bundle.py` provides 4 pure functions: `detect_local_bundle`, `verify_bundle_integrity`, `check_target_mismatch`, `read_bundle_plugin_json` — each with two-gate path-traversal defense (`validate_path_segments` + `ensure_path_within`). |
| 4 | New `--as ALIAS` flag for log namespacing on local installs; rejected with `UsageError` on registry installs. |
| 5 | Resolver/registry/MCP/policy flags rejected on local installs with one consolidated error message. |
| 6 | `apm unpack` marked `[Deprecated]`, emits a one-line warning pointing at `apm install <bundle-path>`. |
| 7 | Local installs are an *imperative deploy*: written paths recorded under `local_deployed_files` / `local_deployed_file_hashes` in the project lockfile; `apm.yml` is never mutated. |

## Implementation (HOW)

- **`src/apm_cli/bundle/local_bundle.py`** *(new, ~330 LOC)* — `LocalBundleInfo` dataclass plus the four pure functions above. Python 3.10/3.11 compat: `tarfile.extractall(filter="data")` is gated on `sys.version_info >= (3, 12)`; older interpreters get manual symlink/abs-path/`..`-segment rejection as the primary gate. Strict `sha256:` hash-prefix validation (unknown algorithm prefixes raise `ValueError`).
- **`src/apm_cli/install/local_bundle_handler.py`** *(new, ~180 LOC)* — Orchestrator extracted to honor the 1800-LOC ceiling on `commands/install.py`. Uses `click.Abort()` (not `sys.exit(1)`) so Click owns the exit code. Dry-run uses `tree_item` to make the diff visible at default verbosity.
- **`src/apm_cli/install/services.py`** — Adds `integrate_local_bundle()` (~150 LOC). Two critical correctness fixes landed here: (a) `if not scope:` was always False (enum members are truthy) → fixed to `if scope == InstallScope.USER:` so project-scope installs record *relative* paths, restoring portability and `apm prune`/`apm uninstall`; (b) deployed-file hashes now read from the *destination* file post-`shutil.copy2`, not the expected hash from source — defense-in-depth for any future content transform.
- **`src/apm_cli/bundle/plugin_exporter.py`** — Removed `apm.lock.yaml` from the per-package exclusion list and added a post-walk pass that sha256-hashes every file in the bundle, then writes the enriched lockfile to the bundle root before archiving.
- **`src/apm_cli/bundle/lockfile_enrichment.py`** — `enrich_lockfile_for_pack` accepts `bundle_files: dict[str, str]` and writes a canonical comma-joined `pack.target` string.
- **`src/apm_cli/commands/install.py`** — `--as ALIAS` Click option + early-exit branch; targeted `UsageError` for tarball-but-not-bundle paths (bare directories still flow to the local-source-install path because that's a supported pre-existing form).
- **`src/apm_cli/commands/pack.py`** — `apm unpack` docstring prefixed `[Deprecated]`; runtime warning via the existing logger.

## Diagrams

Legend: the install-time decision tree — every path either errors clearly, falls through to the unchanged registry flow, or runs verify-then-deploy. No silent fall-throughs.

```mermaid
flowchart LR
    U["apm install ARG"] --> D{"Path(ARG)<br/>exists?"}
    D -- "no" --> R["registry / git path<br/>(unchanged)"]
    D -- "yes" --> B{"plugin.json at<br/>bundle root?"}
    B -- "no, .tar.gz" --> E1["UsageError:<br/>not a valid bundle"]
    B -- "no, dir" --> R
    B -- "yes" --> V["verify_bundle_integrity<br/>(sha256 + path guards)"]
    V -- "fail" --> E2["Abort: integrity error"]
    V -- "ok" --> I["integrate_local_bundle<br/>copy + record paths"]
    I --> L["write project<br/>apm.lock.yaml"]
```

Legend: pack-time SBOM embedding — the lockfile is generated, hashed against the bundle, and written *into* the bundle so the install side has something to verify against.

```mermaid
sequenceDiagram
    participant Author
    participant Pack as apm pack
    participant LE as lockfile_enrichment
    participant PE as plugin_exporter
    participant Bundle as plugin bundle
    Author->>Pack: apm pack (default = plugin)
    Pack->>PE: export_plugin_directory()
    PE->>PE: walk + sha256 every file
    PE->>LE: enrich_lockfile_for_pack(bundle_files, target)
    LE-->>PE: lockfile with pack.bundle_files + pack.target
    PE->>Bundle: write apm.lock.yaml at root
    PE->>Bundle: write plugin.json + skills/ + agents/ + ...
    Bundle-->>Author: ready to ship (SBOM embedded)
```

## Trade-offs

- **Embedded `apm.lock.yaml` in plugin bundles vs. separate sidecar.** Chose embedded because (a) the bundle becomes self-describing (one artifact, one provenance file), (b) Claude Code's plugin format ignores unknown root files so there's no schema collision, and (c) it matches every other package format that ships its own manifest (`PKG-INFO`, `.cargo_vcs_info.json`, `package.json`). Rejected sidecar because two-file artifacts are easy to drift out of sync.
- **Imperative local installs vs. mutating `apm.yml`.** Chose imperative — local installs record their footprint in the lockfile but never touch the declarative source of truth. Rejected `apm.yml` mutation because that would force the user's *declared* dependency list to absorb side-channel artifacts, breaking reproducibility.
- **`apm unpack` deprecated, not removed.** One release of overlap so CI workflows can migrate. Removal scheduled for v0.13.
- **Maintainer-decision items deferred.** (1) Pre-0.12 bundles (no `pack.bundle_files`) currently warn-and-proceed; design called for hard-reject. (2) `--as ALIAS` currently surfaces only as a verbose log line — could be promoted to lockfile namespacing later. (3) `integrate_local_bundle` does a raw file-copy loop; the architect's design called for routing through `integrate_package_primitives`. Documented as `TODO(#1098-v0.13)`.

## Benefits

1. **One verb to learn.** `apm install` is now the single restore primitive — registry, local bundle, or local source path. Removes the cognitive cost of remembering which command to use for which artifact.
2. **Bundles are now SBOMs.** Every plugin pack since this PR carries a sha256 manifest of its own contents. Downstream consumers can audit without re-walking the tree.
3. **Path-traversal defense at the install boundary.** Crafted bundle entries are rejected by `validate_path_segments` + `ensure_path_within` before any file is written.
4. **Air-gapped restore.** Verified by an integration test that patches `urllib`, `socket.create_connection`, `socket.socket`, `httpx`, `requests`, and `subprocess` — local installs do zero network I/O.
5. **CI workflows simpler.** `apm pack` followed by `apm install ./bundle.tar.gz` is now a one-line restore in any GitHub Actions / Azure DevOps job.

## Validation

Lint (CI-mirror, must be silent) — silent:

```
All checks passed!
620 files already formatted
```

Targeted test surface — 69 passed, no skips:

```
$ uv run --extra dev pytest tests/integration/test_install_local_bundle_e2e.py \
                              tests/unit/bundle \
                              tests/unit/install/test_install_local_bundle.py \
                              tests/unit/commands/test_unpack_deprecation.py -q
.....................................................................   [100%]
69 passed in 4.28s
```

Two-round review-panel convergence (5 specialists in round 1, 2 specialists in round 2):

- **Round 1** flagged 7 Critical + 11 Important findings across supply-chain, test-coverage, devx-ux, python-architecture, cli-logging-ux. All Critical and 10/11 Important addressed in commit `0e240941`.
- **Round 2** (supply-chain + test-coverage re-check): both verdicts **SHIP**. Path-traversal guards verified at both verify and deploy sites; air-gap socket sentinels verified in place; all 6 TDD stubs materialized into 29 active tests; multi-target `pack.target` schema pinned.

### Scenario Evidence

| 1 | "I packed a bundle and want to install it on a different machine" | Portability / DevX | `tests/integration/test_install_local_bundle_e2e.py::TestLocalInstallRoundTrip` | integration |
| 2 | "Installing a local bundle does zero network I/O" | Secure by default | `tests/integration/test_install_local_bundle_e2e.py::TestLocalInstallAirGap` | integration |
| 3 | "A tampered bundle entry is rejected before any file is written" | Secure by default | `tests/unit/bundle/test_local_bundle.py::test_bundle_files_path_traversal_rejected` | unit |
| 4 | "An unlisted file in the bundle is flagged at verify time" | Secure by default | `tests/unit/bundle/test_local_bundle.py::test_unlisted_bundle_file_flagged` | unit |
| 5 | "Project-scope local installs record relative paths so the lockfile is portable" | Portability | `tests/unit/bundle/test_local_bundle.py::test_bundle_files_absolute_path_rejected` + round-trip fidelity | unit |
| 6 | "`apm install ./bundle.tar.gz` errors clearly when the tarball isn't an APM bundle" | DevX | `tests/unit/install/test_install_local_bundle.py::test_invalid_tarball_raises_usage_error` | unit |
| 7 | "`--as ALIAS` is rejected on registry installs" | DevX / consistency | `tests/unit/install/test_install_local_bundle.py::test_as_rejected_on_registry_install` | unit |
| 8 | "Resolver/registry/MCP/policy flags are rejected on local installs" | DevX | `tests/unit/install/test_install_local_bundle.py` parametrized rejected-flags matrix (14 cells) | unit |
| 9 | "`apm pack` embeds a multi-target lockfile schema authors can rely on" | Multi-harness | `tests/unit/bundle/test_plugin_exporter_lockfile.py::test_plugin_export_lockfile_multi_target` | unit |
| 10 | "`apm unpack` warns and points at the new entrypoint" | DevX migration | `tests/unit/commands/test_unpack_deprecation.py` | unit |

<details><summary>Full pre-push verification transcript</summary>

```
$ git log --oneline main..HEAD
0e240941 fix(install-local-bundle): apply review-panel findings (CR1-CR7, IM1-IM11)
797ebe8c feat: apm install <local-bundle> + lockfile-embedded plugin packs (#1098)
4f82a66f test: TDD harness for apm install <local-bundle> (#1098)

All checks passed!
620 files already formatted

$ uv run --extra dev pytest tests/integration/test_install_local_bundle_e2e.py tests/unit/bundle tests/unit/install/test_install_local_bundle.py tests/unit/commands/test_unpack_deprecation.py -q
.....................................................................   [100%]
69 passed in 4.28s
```

</details>

## How to test

- [ ] Pack a project: `apm pack --format plugin -o /tmp/myplugin` → confirm `/tmp/myplugin/apm.lock.yaml` exists with a `pack.bundle_files` map.
- [ ] Install the directory: `apm install /tmp/myplugin` → confirm files land in `.claude-plugin/` (or target dir) and project `apm.lock.yaml` records `local_deployed_files`.
- [ ] Re-pack as tarball: `apm pack --format plugin -o /tmp/myplugin.tar.gz` then `apm install /tmp/myplugin.tar.gz` → same outcome.
- [ ] Try an obvious footgun: `apm install ./does-not-exist.tar.gz` → Click `UsageError` with bundle-not-found wording (no Git clone attempt).
- [ ] Confirm legacy: `apm unpack /tmp/myplugin.tar.gz` → still works, prints `[Deprecated]` warning pointing at `apm install`.


danielmeppiel:
## apm-action PR #31 compatibility assessment

Short answer: **#31 is sufficient as a defensive measure on its own; once #1099 ships in apm 0.12, a follow-up apm-action change can additionally light up plugin-format restore and remove the current rejection.**

### Why #31 already works after #1099 lands

| apm-action mode (after #31 merges) | After apm 0.12 ships #1099 | Status |
|---|---|---|
| `bundle-format: 'apm'` (default) | Restores via `apm unpack` (still works) | OK |
| `bundle-format: 'plugin'` (opt-in pack) | Pack succeeds; restore intentionally rejected by `extractBundle` with operator-friendly error | OK (rejection is correct *today* but stale once 0.12 is out) |
| `setup-only: 'true'` | Independent of pack/restore | OK |

The `bundle-format: 'apm'` defensive default in #31 means **no apm-action consumer breaks** when apm flips its pack default to `plugin` upstream.

### What #1099 unlocks for apm-action (follow-up, not blocking)

The reasoning #31 cites for rejecting plugin restore -- *"`apm unpack` itself rejects plugin tarballs (different deployment contract -- no lockfile to drive `deployed_files`). That belongs upstream in `apm unpack`, not here."* -- is exactly what #1099 fixes upstream:

1. `apm pack --format plugin` now embeds `apm.lock.yaml` with a `pack.bundle_files` sha256 manifest.
2. `apm install <bundle-path>` accepts plugin tarballs (and directories), verifies integrity against the embedded lockfile, and records `local_deployed_files` in the project lockfile.

So once a consumer's `apm-version` is `>= 0.12`, plugin-format restore is technically possible -- the action just needs to call `apm install <path>` instead of `apm unpack <path>`.

### Recommended follow-up apm-action PR (post 0.12 release)

Open after apm 0.12 ships:

- Detect installed apm version; on `>= 0.12`, route restore through `apm install <bundle-path>` for both formats.
- On `< 0.12`, keep the current `apm unpack` path (with the `bundle-format: 'plugin'` rejection intact for backwards compatibility).
- Optionally flip the `bundle-format` default to `'plugin'` once the floor is `0.12+`, aligning apm-action's default with the upstream CLI default.
- The `apm unpack` deprecation warning will start appearing in CI logs for `bundle-format: 'apm'` users -- worth migrating off.

### TL;DR

- **Ship #31 now.** It correctly defuses the immediate compatibility break from the upstream pack-default flip, and the `setup-only` mode (#24) is independent of #1099.
- **No blocking changes needed in #31** for #1099. The plugin-rejection error message could optionally be softened once #1099 is released ("supported in apm >= 0.12 -- upgrade your `apm-version`"), but that's a polish pass, not a fix.
- **One follow-up issue to file** in apm-action: "Switch restore to `apm install <bundle>` when `apm-version >= 0.12`; unlocks plugin-format restore and removes the deprecation warning."

PR #1230, opened 2026-05-09
fix(lockfile): pass scope-resolved path to MCPIntegrator.update_lockfile

### LINKED ISSUE #794: Bug: `MCPIntegrator.update_lockfile` uses `Path.cwd()` at `--global` scope, writing MCP audit entries to the wrong lockfile
## Summary

`MCPIntegrator.update_lockfile()` defaults `lock_path` to `get_lockfile_path(Path.cwd())` when the caller does not pass an explicit path. At `--global` scope (`apm install -g`), the lockfile should be written to `~/.apm/apm.lock.yaml`, but callers in `install.py` do not pass the scope-resolved lockfile path to `update_lockfile`. This causes MCP server audit entries to be written to the project-local lockfile instead of the user-scope lockfile.

## Reproduction

1. `cd /tmp/some-project && apm init`
2. `apm install -g some-org/some-mcp-package`
3. Observe that `./apm.lock.yaml` (project) gets MCP entries instead of `~/.apm/apm.lock.yaml`.

## Expected Behavior

At `--global` scope, `update_lockfile` should persist MCP server entries in the user-scope lockfile (`~/.apm/apm.lock.yaml`).

## Relevant Code

- `src/apm_cli/integration/mcp_integrator.py`, `update_lockfile()` method (lines ~639-673)
- `src/apm_cli/commands/install.py` or `src/apm_cli/install/` callers

## Context

Discovered during review of #638 (`apm install --global` MCP scope filtering). The scope-filtering fix in #638 correctly routes MCP installation to global-capable runtimes, but the lockfile path was out of scope for that PR.

## Labels

bug

Aaryan-Dadu:
## Description

`MCPIntegrator.update_lockfile` was using `Path.cwd()` as the default lockfile location even at `--global` scope, causing MCP server audit entries to be written to the project-local `apm.lock.yaml` instead of the user-scope `~/.apm/apm.lock.yaml`.

Most caller in the install pipeline omitted the `lock_path` argument. Since `Path.cwd()` is always the current project directory (not the user's home), MCP entries for a global install were silently written to the wrong file.

#### Fix
Thread the already-computed scope-resolved `_lock_path` (derived from `ctx.apm_dir`, which is `~/.apm/` for `--global` and `Path.cwd()` for project scope) into all four `update_lockfile` call sites.

Fixes #794 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass


Aaryan-Dadu:

danielmeppiel:
Thanks for the fix @Aaryan-Dadu! Closing as superseded -- the same fix landed in #1236 (merged earlier today, closes #794). The diagnosis and call-site routing here matched #1236's approach almost exactly, so the underlying behavior is now correct on main. Really appreciate the contribution -- hope to see you back on another issue!

PR #1301, opened 2026-05-13
fix(deps): reject shorthand '@alias' at parse time with migration error

prateek:
## Summary

`apm.yml` string-form entries like `owner/repo/sub/path@alias` (subpath + trailing `@alias`) were silently miscoerced: the `@alias` ended up glued to `virtual_path` and the install failed downstream with the unhelpful `Subdirectory 'skills/orchestration@orca-stration' not found in repository`.

Shorthand `@alias` was deliberately retired in #340 (npm/go/cargo `@version` collision). The whole-repo case (`owner/repo@alias`) already failed loudly with `Invalid repository path component`, but the subpath case slipped past `_detect_virtual_package` because it never looked for `@` in the trailing segments.

This PR adds a single parse-time guard so all three shorthand-`@alias` shapes — `owner/repo@alias`, `owner/repo/sub@alias`, `owner/repo#ref@alias` — produce the same actionable error pointing the user to the object form.

## Repro

```yaml
# apm.yml
dependencies:
  apm:
    - stablyai/orca/skills/orchestration@orca-stration
```

**Before:**

```
[x] 1 package failed:
  +- orca-orchestration@orca-stration -- Failed to download dependency stablyai/orca: Subdirectory 'skills/orchestration@orca-stration' not found in repository
```

**After:**

```
[x] Failed to parse apm.yml: Invalid APM dependency 'stablyai/orca/skills/orchestration@orca-stration': Shorthand '@alias' is not supported in 'stablyai/orca/skills/orchestration@orca-stration'. Use the object form with an 'alias:' field to install a dependency under a custom directory name. See: https://microsoft.github.io/apm/consumer/manage-dependencies/#reference-formats
```

## What changed

- **`src/apm_cli/models/dependency/reference.py`** — new `_reject_shorthand_alias()` static method called from `parse()` after the local-path / `//` checks. Skips `https://`, `http://`, `ssh://`, and SCP shorthand (`<user>@host:path`) — those forms have dedicated parsers that legitimately extract `@alias`. Anything else with `@` raises with migration guidance.
- **`docs/src/content/docs/consumer/manage-dependencies.md`** — removed the two stale `Aliased` / `Pinned + aliased` rows that survived #340, and folded `alias:` into the `Object form` row description.
- Dropped two stale `user/repo@alias` examples from the public `parse()` docstring.
- Tests: 8 new rejection cases (subpath, deep subpath, subpath+ref, FQDN, URL-encoded `%40`, trailing `@`) plus a positive `https://user@github.com/...` regression test that locks in the guard's HTTPS-userinfo carve-out. Three pre-existing tests that asserted the old "absorb `@` into ref" behavior were tightened to assert the new uniform rejection.

## Compatibility

- **No round-trip risk.** `to_canonical()` never emits `@alias`, and `LockedDependency.from_dependency_ref` / `to_dependency_ref` work on structured fields, never re-`parse()`-ing a string. Historical lockfiles cannot contain `@alias` and so cannot trip the new guard on load.
- **SSH paths unchanged.** `ssh://...@alias` and `git@host:path@alias` still parse and extract the alias as before.
- **Object form unchanged.** `{ git: ..., path: ..., alias: ... }` is unaffected.

## Test plan

- [x] `pytest tests/unit/` — 8,301 passing
- [x] Re-ran the bug-report repro against the patched build — rejected at parse time with the new error
- [x] Re-ran the object-form workaround from the bug report — installs to `.agents/skills/orca-stration/` as expected
- [ ] CI green

### DIFF
--- a/docs/src/content/docs/consumer/manage-dependencies.md
+++ b/docs/src/content/docs/consumer/manage-dependencies.md
@@ -54,8 +54,6 @@ parser. The supported forms:
 |---|---|---|
 | GitHub shorthand | `owner/repo` | Public GitHub repo, latest default branch. |
 | Pinned ref | `owner/repo#v1.0.0` | Pin to a tag, branch, or full commit SHA. |
-| Aliased | `owner/repo@my-alias` | Install under a custom directory name. |
-| Pinned + aliased | `owner/repo#v1.0.0@my-alias` | Combine the two. |
 | FQDN shorthand | `gitlab.com/acme/repo#v2.0` | Any git host, not just github.com. |
 | Virtual subdirectory | `owner/repo/skills/review` | Install one skill folder from a monorepo. |
 | Virtual file | `owner/repo/prompts/review.prompt.md` | Install a single primitive file. |
@@ -63,7 +61,7 @@ parser. The supported forms:
 | SSH SCP-style | `git@gitlab.com:acme/repo.git` | SSH with default port. |
 | SSH protocol | `ssh://git@gitlab.com/acme/repo.git` | SSH with explicit scheme or port. |
 | Local path | `./packages/shared` or `/abs/path` | Sibling package on disk. |
-| Object form | `{ git: <url>, path: <subpath>, ref: <ref> }` | Escape hatch for nested groups, monorepo subpaths, or aliases that the string forms cannot express. |
+| Object form | `{ git: <url>, path: <subpath>, ref: <ref>, alias: <name> }` | Custom directory name (`alias`), nested groups, monorepo subpaths, or anything the string forms cannot express. |
 
 Object form in YAML:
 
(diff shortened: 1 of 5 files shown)


prateek:

danielmeppiel:
Superseding this PR with #1655 because pushing the rebased branch back to the contributor fork was rejected by GitHub: the branch history now includes a workflow change from main and the available token lacks workflow scope. The superseding branch preserves @prateek's original commit, folds the required Breaking changelog entry, docs migration note, and panel follow-ups, and includes author credit in the follow-up commit trailer.

danielmeppiel:
Closed in favor of superseding PR #1655. Thank you @prateek for the original fix; the replacement keeps your authored commit and adds only the shepherding folds needed to make it landing-ready.

PR #1303, opened 2026-05-13
fix: preflight via ssh on ssh based connections (closes #1293)

### LINKED ISSUE #1293: [BUG] Authentication fails for apm install --update with private Git-Repository, SSH based authentication and SSH-Agent
**Describe the bug**

Since apm 0.11 we can not use private Git Repositories as a dependency target in combination with SSH based authentication and a locally running SSH-Agent as the authentication fails for apm install --update. This is a regression as it worked in apm <= 0.10.0 . This only happens with --update as the preflight request fails. Without --update it works.

**To Reproduce**

apm.yml:

```name: example
version: 1.0.0
description: APM project for example
author: Example
dependencies:
  apm:
  - git: ssh://git@example.com/927eb5ab/ai/agentic-components.git
    ref: v1.0
  mcp: []
scripts: {}
```

1. Run apm install --force --verbose --update --target opencode --no-policy

```[>] Installing dependencies from apm.yml...
Parsed apm.yml: 1 APM deps, 0 MCP deps
Loaded apm.lock.yaml for SHA comparison (2 dependencies)
    927eb5ab/ai/agentic-components: locked at 515cfb72
Resolved 1 direct dependencies (no transitive)
Phase: resolve -> 0.010s
[!] Policy enforcement disabled by --no-policy for this invocation. This does NOT bypass apm audit --ci. CI will still fail the PR for the same policy violation.
Phase: policy_gate -> 0.000s
[i] Targets: opencode  (source: --target flag)
Phase: targets -> 0.034s
Phase: policy_target_check -> 0.000s
Preflight: gitlab.example.com/927eb5ab -- auth rejected
Authentication failed for gitlab.example.com
Authentication failed for install --update on gitlab.example.com.
No token available.
APM does not apply GitHub PAT environment variables to generic git hosts; configure git credential fill for gitlab.example.com or use a public repository if available.
Run with --verbose for detailed auth diagnostics.

    No files were modified.
    apm.yml, apm.lock.yaml, and apm_modules/ are unchanged.
[!] Install interrupted after 0.5s.
```

**Expected behavior**

SSH-based authentication should work with a locally configured ssh-agent. apm should automatically use the SSH-Agent for authentication as it did before.

**Environment:**
 - OS: macOS
 - Python Version: 3.14.4
 - APM Version: 0.11+

Alpha200:
## Description

Fixes the ssh preflight bug by doing the preflight via ssh on ssh based connections.

Fixes #1293

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass


Alpha200:

danielmeppiel (changes requested):
Address all panel blockers and recommendations, thanks!

danielmeppiel:
Maintainer sweep [2026-05-15]: still in `needs_rework` state per panel + maintainer review (CHANGES_REQUESTED, 2026-05-13). Four blockers were called out:

1. SSH preflight signal-list misclassification.
2. Zero test coverage on the new SSH preflight path.
3. A deleted regression trap.
4. Missing CHANGELOG entry.

5 cancelled CI checks suggest the latest push didn't complete a green run either. @Alpha200, can you address those four items in a single follow-up push? Once they're in, I'll re-trigger the panel and ship.

danielmeppiel:
## PR-shepherd review (automated by orchestrator session)

**Verdict:** needs-author-changes

This PR is on the right track and the testing surface has grown meaningfully since the first panel pass, but a load-bearing blocker called out by the most recent panel run (2026-05-18) remains unaddressed, and the merge state is currently red on three independent gates.

### Status gates (current `head`)

- `reviewDecision`: `CHANGES_REQUESTED`
- `mergeable`: `CONFLICTING` (needs rebase on `main`)
- CI `Lint` job: `FAILURE` -- `ruff format --check` reports `tests/unit/test_github_host.py` needs reformatting (single file, mechanical fix: `uv run --extra dev ruff format src/ tests/`)
- CI `Build & Test (Linux)`, `APM Self-Check`, `CodeQL`, `Merge Gate`: green

### Root-cause alignment: PARTIAL

The PR does fix the originally-reported failure mode in `src/apm_cli/install/pipeline.py` -- `use_ssh=False` is no longer hard-coded into `_dl._build_repo_url(...)`, and an explicit `git: ssh://...` scheme is now honoured. For the bug-report repro (Alpha200's `ssh://git@host/...` + ssh-agent, `apm install --update`), this PR will produce the correct behaviour.

However, the `not dep_ctx.token` half of the new heuristic introduces a fresh regression on a different generic-host code path -- see the auth-expert blocker below. The fix addresses the symptom in #1293 but does not stay in lock-step with `TransportSelector.select()`, so a different subset of generic-host users will now hit a `false-AuthenticationError` they did not have before.

### Specialist panel findings (material only -- from 2026-05-18 PR Review Panel run)

- **[BLOCKING, auth-expert] Transport mismatch with `TransportSelector.select()`** -- `src/apm_cli/install/pipeline.py:147`
  - Probe: `_use_ssh = is_generic and (_explicit_scheme == 'ssh' or not dep_ctx.token)`
  - Reality (verified at `src/apm_cli/deps/transport_selection.py:313-316`): for shorthand generic-host deps at `ProtocolPreference.NONE`, `initial = [_AUTH_HTTPS] if has_token else [_PLAIN_HTTPS]`; SSH is the chained fallback only.
  - Consequence: a user with no token, no SSH key, and a public Gitea/Forgejo dep (anonymous HTTPS) -- which `apm install` (without `--update`) handles fine via `_PLAIN_HTTPS` -- now has the preflight raise `AuthenticationError` over a fabricated SSH probe.
  - Suggested fix: replace the inline heuristic with a delegated call to `TransportSelector.select(...).initial_scheme` (or the equivalent first-attempt scheme), wired with a `NoOpInsteadOfResolver` so probe transport stays bound to the real clone transport.

- **[recommended, auth-expert / supply-chain] `"host key verification failed"` is misclassified as auth failure** -- `src/apm_cli/utils/github_host.py:_SSH_AUTH_FAILURE_SIGNALS`
  - A genuine MITM / host-key-rotation event currently surfaces the `ssh-add -l` remediation rather than a "verify the remote fingerprint" message. Move this signal to a separate bucket and emit a distinct diagnostic, or defer to the download phase.

- **[recommended, python-architect] `_SSH_CONNECTIVITY_SIGNALS` is dead code** -- `src/apm_cli/utils/github_host.py`
  - Defined, never imported, never consumed. The "DNS/firewall failures defer" guarantee currently rides on the implicit disjoint-set property of `_SSH_AUTH_FAILURE_SIGNALS`. Either consume `_SSH_CONNECTIVITY_SIGNALS` as a short-circuit inside `is_ssh_auth_failure_signal`, or delete it and pin the exclusion list in the docstring.

- **[recommended, python-architect] `'publickey'` substring is too permissive** -- `src/apm_cli/utils/github_host.py`
  - With `GIT_SSH_COMMAND='ssh -v'`, OpenSSH emits `debug1: Offering public key: ...` and `debug1: Next authentication method: publickey` on stderr. These can false-positive `is_ssh_auth_failure_signal`. Strip `GIT_SSH_COMMAND` from `probe_env` alongside `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_NOSYSTEM`/`GIT_ASKPASS` in `_preflight_auth_check`, and/or tighten the substring (e.g. require it on a line with `permission denied` or `no more authentication methods`).

- **[recommended, supply-chain] Raw git stderr embedded verbatim in diagnostic** -- `pipeline.py` SSH branch
  - HTTPS path goes through `auth_resolver.build_error_context(...)` (controlled message). SSH path inlines `{stderr_text.strip()}` with no ANSI/control-char stripping and no length cap. Suggest `re.sub(r'[\x00-\x08\x0b-\x1f\x1b\x7f]', '', stderr_text).strip()[:512]`.

- **[nit, cli-logging] British spelling `authorised`** -- `pipeline.py` SSH branch. Every other auth message in the repo uses American spelling.

### E2E coverage: PARTIAL

Present (good additions):
- `TestSshPreflightAuthRejection` (permission denied, no more auth methods)
- `TestSshPreflightConnectivityDeferred` (connection refused, DNS failure -- proves the defer contract)
- `TestSshPreflightExplicitScheme` (explicit `ssh://` forces SSH probe even with token)
- `TestIsSshAuthFailureSignal` -- full positive/negative matrix on the classifier

Missing (required before merge):
1. **Regression-trap test for the auth-expert blocker.** Today the `_use_ssh=False` branch in `TestPreflightGenericHostAllowsCredentialHelpers` asserts env-var stripping but not that the probe URL is HTTPS. If `_use_ssh` silently flipped to `True` for all generic hosts, that test would still pass. Drop-in:

```python
@patch("subprocess.run")
def test_generic_host_with_token_uses_https_probe_url(self, mock_run):
    """When a token is present on a generic host (no explicit ssh:// scheme),
    the preflight probe URL must be HTTPS, not SSH -- otherwise we diverge
    from TransportSelector.select() and raise spurious AuthenticationError
    on hosts where the real clone would have succeeded over HTTPS."""
    mock_run.return_value = MagicMock(returncode=0, stderr="", stdout="")
    from apm_cli.install.pipeline import _preflight_auth_check

    dep = _make_ssh_dep(host="git.corp.internal", repo_url="org/repo")
    dep.explicit_scheme = None
    ctx = _make_ctx(deps=[dep])
    resolver = _make_ssh_resolver(token="ghp_dummy_token")  # token present

    _preflight_auth_check(ctx, resolver, verbose=False)

    # Probe argv: ["git", "ls-remote", "<url>", ...]
    probe_argv = mock_run.call_args[0][0]
    probe_url = probe_argv[2]
    assert probe_url.startswith("https://"), (
        f"Expected HTTPS probe URL when token is present on generic host, "
        f"got: {probe_url}"
    )
```

2. **Public-anonymous-HTTPS defer test** (proves the blocker is closed once the heuristic is fixed):

```python
@patch("subprocess.run")
def test_generic_host_no_token_no_explicit_scheme_uses_https_probe(self, mock_run):
    """No token + no explicit ssh:// scheme on a generic host: probe must be
    HTTPS (matches TransportSelector initial=_PLAIN_HTTPS), so a public repo
    on Gitea/Forgejo with anonymous HTTPS is NOT blocked by a fabricated
    SSH AuthenticationError."""
    mock_run.return_value = MagicMock(returncode=0, stderr="", stdout="")
    from apm_cli.install.pipeline import _preflight_auth_check

    dep = _make_ssh_dep(host="gitea.public.example", repo_url="org/repo")
    dep.explicit_scheme = None
    ctx = _make_ctx(deps=[dep])
    resolver = _make_ssh_resolver(token=None)  # no token

    _preflight_auth_check(ctx, resolver, verbose=False)

    probe_url = mock_run.call_args[0][0][2]
    assert probe_url.startswith("https://"), (
        f"Expected HTTPS probe (anonymous) for tokenless generic dep with "
        f"no explicit scheme; got: {probe_url}"
    )
```

3. (Optional, recommended) `GIT_SSH_COMMAND` stripping test -- assert `'GIT_SSH_COMMAND'` is absent from `probe_env` after the strip block, closing the `publickey` false-positive surface.

### Empirical repro

Deferred -- the SSH preflight needs a real ssh-agent + a generic Git host fixture, which is infeasible from this orchestrator session. Confidence rests on:
- The PR's own unit tests (mocked `subprocess.run`) -- pass and cover the right shapes for the rejected/deferred classification matrix.
- A direct source-level cross-check against `src/apm_cli/deps/transport_selection.py:313-316`, which confirms the auth-expert blocker is real and not a false alarm.

### Suggested follow-ups (post-merge or in this PR)

- Move `'host key verification failed'` into a separate `_SSH_HOST_KEY_SIGNALS` bucket with a distinct diagnostic.
- Audit the download-phase error handler so deferred DNS/firewall failures still surface a structured `apm.yml / apm.lock.yaml / apm_modules/ unchanged` footer instead of raw git stderr.
- Add a `docs/src/content/docs/getting-started/authentication.md` note explaining that `apm install --update` probes via SSH for generic hosts without a configured token, with an `ssh-add -l` pointer.

---
Orchestrator session: danielmeppiel/triage-issues-1363-1364-1366

danielmeppiel:
## Ready to merge

Rebased onto current `main` and addressed the three blocking issues from the prior shepherd review.

### Changes since last review

**1. Heuristic alignment with `TransportSelector` (was: heuristic divergence)**

Previously, the PR set:
```python
_use_ssh = is_generic and (_explicit_scheme == "ssh" or not dep_ctx.token)
```
The `or not dep_ctx.token` clause diverged from `TransportSelector.select()` (`src/apm_cli/deps/transport_selection.py:312-316`), which for shorthand deps without an explicit scheme returns `[_PLAIN_HTTPS]` when `has_token=False` — never SSH. The PR's logic would have routed anonymous-HTTPS Gitea/Forgejo deps through an SSH probe, breaking tokenless public access for users with no configured SSH key.

Fixed to mirror the selector exactly:
```python
_use_ssh = is_generic and _explicit_scheme == "ssh"
```
Only an explicit `ssh://` scheme triggers SSH preflight. Everything else (shorthand + token, shorthand + no token, explicit `https://`) goes through HTTPS, matching `TransportSelector.select()`.

**2. Two regression-trap tests added** (`tests/unit/install/test_pipeline_auth_preflight.py`, new class `TestPreflightTransportMirrorsSelector`):
- `test_tokenless_generic_shorthand_probes_https` — guards the blocker case (would have caught the divergence above).
- `test_tokened_generic_shorthand_probes_https` — guards the silent `_use_ssh=True` regression for token-present generic hosts.

**3. Lint fix** — `ruff format` clean on `tests/unit/test_github_host.py`.

**4. Rebase** — branch is now on top of `main` (was CONFLICTING).

### Verification

- `uv run --extra dev ruff check src/ tests/` — silent
- `uv run --extra dev ruff format --check src/ tests/` — silent
- `uv run --extra dev pytest tests/unit/install/test_pipeline_auth_preflight.py tests/unit/test_github_host.py` — 65 passed
- `uv run --extra dev pytest tests/unit/install/ tests/unit/deps/` — 947 passed
- All CI checks green: Lint, Build & Test (Linux), CodeQL, Analyze, APM Self-Check, NOTICE Drift Check, gate

### Commits

- `d6537a9e` (Alpha200): original SSH preflight fix, rebased clean
- `e0818e92` (Alpha200): integrated requested changes from earlier review, rebased clean
- `95963d1d` (follow-up): heuristic alignment with TransportSelector + two regression traps + lint fix

Ready for maintainer re-review and merge. Thanks @Alpha200 for the original fix — the divergence was subtle and only surfaced when reconciling with the selector's tokenless-shorthand branch.

Issue #1204, opened 2026-05-08
[BUG] outdated option is missing help text


rcollette:
**Describe the bug**
outdated option is missing help text

**To Reproduce**
run `apm --help`

**Expected behavior**
The row containing the `outdated` command should have help text.

**Environment (please complete the following information):**
 - OS: macOS
 - Python Version: 3.12.0
 - APM Version: 0.12.4

PR #1307, opened 2026-05-13
fix(build): keep Click command docstrings in release binaries

### LINKED ISSUE #1298: [BUG] Help missing for view command
**Describe the bug**
`apm --help` is missing summary information and `apm view --help` is missing detailed help for the `apm view` command.

 - APM Version: 0.8.13

edenfunf:
## Problem

`apm --help` shows the `view` row with no summary, and `apm view --help` shows only `Usage:` / `Options:` with no description, fields, or examples — in release binaries only. Source builds render the help correctly.

Reported in #1298 against 0.8.13 (the first release containing `view`; the same shape previously affected `outdated` in #1204 / #1216).

## Root cause

`build/apm.spec` sets PyInstaller `optimize=2`, which is `python -OO` — it strips both asserts AND `__doc__`. Every Click `@click.command()` that omits an explicit `help=` kwarg loses its summary and detailed help in the binary, because Click's documented fallback is `func.__doc__`.

The project has been working around this by defensively setting `help=` on every command. `view` was the only command that relied on the docstring fallback, so it was the only one broken in v0.13.0; `outdated` hit the same trap previously (#1216 patched it the same defensive way).

## Fix

Lower `optimize=2` → `optimize=1` in `build/apm.spec`. `-O` still strips asserts (the actual perf / size benefit), but `__doc__` survives, so Click's documented help-from-docstring fallback works as advertised. The `apm info` hidden alias (which reads `view_cmd.help`) is fixed automatically as a side-effect.

One-line spec change with a one-line comment recording the trade-off.

## Validation

Built two binaries from this branch on Windows (PyInstaller 6.16.0):

- **Baseline (`optimize=2`)** reproduces #1298: `view$` (empty summary) and `apm view --help` missing description / Fields / Examples.
- **Fix (`optimize=1`)**: `apm view --help` is byte-identical to source `apm view --help` (only difference is `argv[0]` program name in `Usage:`). All 25 visible commands still have summaries; `apm info` hidden alias also renders correctly.

Functional tests against the fix binary:
- `apm view <installed-pkg>` — renders Rich panel
- `apm view <missing-pkg>` — exits 1 with available-package list
- `apm view <pkg> <invalid-field>` — exits 1 with valid-field hint
- `apm info <pkg>` — works (hidden alias)

Test suite (`tests/unit/`): 8275 passed, 5 skipped. Pre-existing 3 failures + 2 errors (Windows path absoluteness in `test_config_command`, missing optional `jsonschema` dep) reproduce on `origin/main` without this change.

Binary size delta: bundle 29.6 MB → 31.0 MB (+1.4 MB / +4.6 %); exe 8.86 MB → 10.1 MB. Trade-off documented in the spec comment.

Fixes #1298


sergio-sisternes-epam (approved):

edenfunf:
@sergio-sisternes-epam  To resolve the conflict, thank you!

danielmeppiel (approved):

PR #1165, opened 2026-05-06
fix(targets): explicit, auditable target resolution (closes #1154)

danielmeppiel:
# fix(targets): explicit, auditable target resolution

## TL;DR

Replaces APM's silent `copilot` fallback with a strict three-tier resolution chain (`--target` flag > `apm.yml targets:` > auto-detect), surfaces what APM picked via a `[i] Targets: ... (source: ...)` provenance line, adds an `apm targets` discovery command (table + `--json`) and an `apm compile --all` flag (deprecates `--target all`), and exits 2 with a teaching message on empty repos. Closes a 4-month, 9-issue cluster of "APM silently deployed to `.github/`" reports.

> [!NOTE]
> **Closes:** #1154 (root report), #805 (`.github/`-only false-positive -- now requires Copilot marker).
>
> **Partially addresses:** #1122 (picker shows which harnesses are present, but does not inventory existing agent context -- skills, instructions, agents, hooks -- per harness; full brownfield inventory deferred), #1130 (same scope: harness presence only).
>
> **Supersedes (recommend closing):** PR #1069 (Copilot-marker requirement) and PR #1115 (treat `.github/`-only as ambiguous) -- both are subsumed by this PR's Tier 4 marker-only Copilot detection plus the explicit-targets precedence chain.
>
> **Out of scope (NOT closed -- separate surfaces):** #518 (agentrc Next Steps), #650 (user-scope `install -g`), #702 / #863 (Kiro / Cline vendor coverage), #888 (`--root` flag), #891 (custom skill target dir), #1056 (missing apm.yml silent skip), #1138 (Claude dedup -- handled by PR #1146).
>
> An apm-review-panel advisory comment with `stance: ship_now` and 4 follow-ups (symlink hardening, compile-side parity, README L57-62 refresh, release-note story) is already posted.

## Problem (WHY)

- [x] **Silent harness fallback.** `CLAUDE.md` present without `.claude/` dir routed a Claude install to `copilot` and deployed to `.github/` -- no warning, no override, no way to debug. Original report #1154; replicated across 9 issues over 4 months.
- [x] **No auditable provenance.** Users had no way to see *what* target APM picked or *why*. The CLI output never named the resolved targets or the signal source.
- [x] **No discovery surface.** No command answered "what targets does APM see in my repo?" -- users had to read source.
- [x] **`--target all` magic string.** `compile --target all` mixed an enum value with a meta-token; surface area for typos and ambiguous flag semantics.
[...]

Issue #1154, opened 2026-05-05
Claude skills silently skipped when CLAUDE.md exists but .claude/ directory does not


tillig:
## Summary

When a project has a `CLAUDE.md` file (strong signal that Claude Code is in use) but no `.claude/` directory, `apm install` and `apm compile --targets all` silently skip Claude skill deployment. No warning is emitted, and the user has no indication that Claude was excluded from skill integration.

## To Reproduce

1. Start with a repo that has `CLAUDE.md` and `.github/` but no `.claude/` directory:
   ```
   CLAUDE.md          ← exists (file)
   .github/           ← exists (directory)
   apm.yml            ← no explicit target: field
   ```

2. Run `apm install` followed by `apm compile --targets all`

3. Observe that `.agents/skills/` is populated (for Copilot, Cursor, Codex, etc.) but `.claude/skills/` is never created.

## Expected behavior

One of:
- `.claude/skills/` is created and populated (since `CLAUDE.md` indicates Claude Code usage), OR
- A warning is emitted explaining that Claude skill deployment was skipped because `.claude/` does not exist, with guidance on how to enable it (e.g., "create `.claude/` or set `target: claude` in apm.yml")

## Root cause

Three factors combine to produce the silent skip:

1. **Auto-detection doesn't see Claude.** `target_detection.py:167` checks `(project_root / ".claude").exists()` — this looks for a path literally named `.claude`, not for `CLAUDE.md`. A project with only `CLAUDE.md` (no `.claude/` directory) does not register Claude as a detected target.

2. **`auto_create=False` for Claude.** Even when `--targets all` resolves and includes Claude in the target list, the directory-creation loop in `install/phases/targets.py:161` skips it:
   ```python
   if not _t.auto_create and not _explicit:
       continue
   ```
   The `--targets all` value from auto-detection (triggered by ≥2 detected folders) is not considered "explicit" in the `_explicit` variable sense (which tracks `--target` CLI flag or `apm.yml target:` field).

3. **Skill integrator bails out.** `skill_integrator.py:341` independently guards on the directory:
   ```python
   if not target.auto_create and not target_root_dir.is_dir():
       continue
   ```
   So even if Claude somehow ends up in the resolved target set, skills aren't deployed.

## Relationship to existing issues

[...]

Issue #1074, opened 2026-04-30
[BUG] apm compile -t <target> generates outputs for ALL targets, ignoring the -t flag


ic4-y:
## Bug Description

`apm compile -t <target>` generates output files for ALL supported targets regardless of what is passed to `-t`. The flag only controls which harness directories get skills deployed to during `apm install`, not what `compile` outputs.

## Environment

- **APM version:** 0.11.0
- **OS:** Linux (NixOS)

## Steps to Reproduce

```bash
mkdir -p /tmp/apm-compile-bug/.apm/instructions
cd /tmp/apm-compile-bug

cat > apm.yml << 'EOF'
name: test-bug
version: 1.0.0
type: hybrid
target: opencode
EOF

cat > .apm/instructions/test.instructions.md << 'EOF'
---
description: Test instruction
applyTo: "*.md"
---

Test content.
EOF

# Compile with opencode target only
apm compile -t opencode
```

## Expected Behavior

`apm compile -t opencode` should only generate `AGENTS.md` (which is what opencode reads). It should not generate `CLAUDE.md`, `GEMINI.md`, or any other target-specific output.

## Actual Behavior

```
$ apm compile -t opencode
[*] Starting context compilation...
[i] Compiling for AGENTS.md + CLAUDE.md + GEMINI.md + .github/ + .claude/ +
.cursor/ + .opencode/ + .codex/ + .gemini/ + .agents/ - explicit --target flag
```

Running `ls` after the command shows both `AGENTS.md` and `CLAUDE.md` (and likely more if the project had matching target files). The `-t opencode` flag had no effect on which files were generated.

## Additional Context

- The `-t` flag description in `--help` says `Target platform: copilot (AGENTS.md), claude (CLAUDE.md), cursor, opencode, or all.` — implying it controls compile output, but it does not.
- This produces unwanted files (e.g., a `CLAUDE.md` in a repo that only uses opencode) that must then be `.gitignore`\d or manually cleaned.
- Verified on 0.11.0 with a clean Python environment (no PYTHONPATH pollution).

## Suspected Location

The compile command likely ignores the `-t`/`--target` parameter and always compiles for all registered targets. The flag is only respected by `apm install` for harness directory deployment.

ic4-y:
Closing — verified this is fixed in 0.11.0. The `-t` flag now correctly restricts compile output to the specified targets.

PR #1351, opened 2026-05-15
fix(compile): --watch path honors apm.yml targets and --target flag

### LINKED ISSUE #1345: [BUG] apm compile --watch regenerates GEMINI.md when targets is [claude, cursor] — regression of #1019 in watch code path
**Describe the bug**
`apm compile --watch` regenerates `GEMINI.md` on every triggered recompile, even when `apm.yml`'s `targets` field excludes Gemini. The non-watch path (`apm compile`) correctly omits `GEMINI.md` for the same project, which suggests the family-collapse fix from #1019 / #1074 was applied to the one-shot compile path but not to the watcher's recompile loop.

**To Reproduce**
1. Create a fresh APM project with `apm.yml`:
     ```yaml
     name: Repro
     version: 1.0.0
     targets:
     - claude
     - cursor
2. Add any .apm/instructions/*.instructions.md file.
3. Remove any pre-existing GEMINI.md: rm GEMINI.md (Windows: Remove-Item GEMINI.md).
4. Run apm compile — verify GEMINI.md is not generated. Output reports only "Generated 2 AGENTS.md files" / "Generated 2 CLAUDE.md files".
5. Run apm compile --watch. On the initial watcher emission, GEMINI.md is generated at the repo root.

Expected behavior

apm compile --watch should produce identical output to apm compile. Since Gemini is not in targets, GEMINI.md should not be generated by either invocation.

Actual behavior

GEMINI.md is generated by --watch (as a 5-line @./AGENTS.md redirect stub) but not by plain apm compile. The two paths diverge.

Environment
- OS: Windows 11 Enterprise 10.0.26100
- APM Version: 0.13.0
- Shell: PowerShell 7

Additional context

#1019 identified the root cause for the non-watch path:
_resolve_compile_target() in src/apm_cli/commands/compile/cli.py collapsed any multi-family target list (e.g., claude + cursor) into "all", and should_compile_gemini_md() in src/apm_cli/core/target_detection.py returned True for "all".
#1019 (closed 2026-04-29) and #1074 (closed 2026-04-30) appear to have fixed this for the one-shot compile path.

This report observes the same symptom in the --watch recompile loop in 0.13.0, suggesting the watcher invokes a parallel code path that was not updated by the #1019 / #1074 fixes. Likely fix: route the watcher's internal compile call through the same target-resolution function that the one-shot path now uses.

Related issues
- #1019 [BUG] apm compile --targets with multi-target list generates GEMINI.md even when Gemini is not a requested target
- #1074 [BUG] apm compile -t  generates outputs for ALL targets, ignoring the -t flag

edenfunf:
# fix(compile): `--watch` path honors `apm.yml` `targets:` and `--target` flag

## TL;DR

`apm compile --watch` now respects the same target inputs that `apm compile` does (`apm.yml`'s `target:` / `targets:` and the CLI `--target` flag), restoring parity that was lost when the watcher diverged from the one-shot path. Achieved structurally: the compile body is extracted into a shared `_run_compile_once` helper that both surfaces call, so future target-resolution work can't drift between them. Closes #1345.

> [!IMPORTANT]
> Two related side-effect changes inside watch mode:
> 1. `apm.yml` is **re-read on every recompile** (it wasn't before — the watcher's `CompilationConfig` was effectively a snapshot of the initial process state with `target="all"`). Mid-session edits to `targets:` now take effect on the next file event.
> 2. `apm compile --watch --clean` now **prints a warning** instead of silently dropping the `--clean` flag. The flag was being ignored before; this surfaces the existing behavior.

## Problem (WHY)

- [x] **`apm compile --watch` ignored `apm.yml` `targets:`.** A project with `targets: [claude, cursor]` (no gemini) produced a stub `GEMINI.md` containing `@./AGENTS.md` on every `--watch` recompile, while the same `apm.yml` under one-shot `apm compile` correctly omitted GEMINI.md. The two surfaces had diverged.
- [x] **`apm compile --watch --target X` was silently ignored too.** Same root cause — the watcher's `_watch_mode(output, chatmode, no_links, dry_run, verbose)` signature never accepted `target` at all, so the CLI flag never reached the watch path. This wasn't called out in #1345 but is the same bug from a different input angle.
- [x] **Both regressed #1019 / #1074.** Those PRs fixed the equivalent symptoms on the one-shot path by introducing `_resolve_compile_target` and routing CLI / `apm.yml` targets through `detect_target`. The watcher had never been wired through that pipeline.

### Root cause

`_watch_mode` reimplemented config construction inside the watcher with a direct `CompilationConfig.from_apm_yml(output_path=..., chatmode=..., resolve_links=..., dry_run=...)` call. That call:

1. Doesn't accept `target=`, so the CLI flag is dropped on the floor.
2. Reads only `apm.yml`'s nested `compilation.target` key — not the top-level `target:` / `targets:` keys (the canonical schema documented for end users and supported on the install path since #1154).

With both inputs ignored, `CompilationConfig.target` stayed at its default `"all"`, which makes `should_compile_gemini_md("all")` return `True` and unconditionally emits the `GEMINI.md` redirect stub.

## Solution (WHAT)

**Extract the compile body, don't add a second target-resolution path to the watcher.** A "watcher copy of the resolution logic" would have re-fixed the symptom but left the same parallel-path shape that caused #1345 in the first place — a future PR fixing target resolution would need to remember to update two places. Instead, the compile body in `cli.py::compile` (target resolution → config build → `compiler.compile()` → result reporting) is extracted into a shared private helper:

```python
def _run_compile_once(*, target, output, chatmode, no_links, dry_run,
                      single_agents, verbose, local_only, clean,
                      with_constitution, logger) -> CompilationResult:
    ...
```

Both surfaces now call it:

- **One-shot** (`compile()` after the `if watch:` guard): single call, exits on `result.errors` or `result.has_critical_security`.
- **Watch** (`_watch_mode`'s initial compile + every debounced file event): wrapped in a local `_do_compile` that catches exceptions so the watcher keeps running after a failed recompile.

`apm.yml` is read fresh inside `_run_compile_once` on every call, so mid-session `targets:` edits in watch mode reflect on the next recompile.

`watcher.py` no longer imports `CompilationConfig` or `AgentsCompiler` at all. To break the import cycle (`cli.py` imports `_watch_mode` from `watcher.py`), `watcher.py` imports `_run_compile_once` lazily inside `_watch_mode` — documented in-line at [watcher.py:28-30](../blob/fix/1345-watch-target-resolution/src/apm_cli/commands/compile/watcher.py#L28-L30).

### Behavior delta

| Input | Before | After |
|---|---|---|
| `apm compile --watch` + `targets: [claude, cursor]` | Emits `AGENTS.md`, `CLAUDE.md`, `GEMINI.md` (stub) | Emits `AGENTS.md`, `CLAUDE.md` only |
| `apm compile --watch --target claude` | Emits all default files; `--target` ignored | Emits `CLAUDE.md` only, matching one-shot |
| `apm compile --watch` + `target: claude` (singular key) | Ignored; defaults to "all" | Honored; emits `CLAUDE.md` only |
| `apm compile --watch` + `--all` | Emits all (was always "all" anyway) | Emits all (positive control unchanged) |
| Editing `apm.yml`'s `targets:` mid-watch | No effect until process restart | Takes effect on next file event |
| `apm compile --watch --clean` | `--clean` silently ignored | `--clean` ignored with explicit `[!]` warning at startup |
| `apm compile --watch --dry-run` | Worked | Works; provenance line `[i] Targets: ... (source: ...)` now visible per recompile |

### Reproducer from the issue

```bash
mkdir -p /tmp/repro-1345/.apm/instructions
cd /tmp/repro-1345

cat > apm.yml <<'EOF'
name: Repro
version: 1.0.0
targets:
- claude
- cursor
EOF

cat > .apm/instructions/style.instructions.md <<'EOF'
---
description: style
applyTo: "**/*.py"
---

snake_case.
EOF

rm -f GEMINI.md

apm compile          # baseline: produces AGENTS.md + CLAUDE.md (no GEMINI.md)
apm compile --watch  # before fix: emits stub GEMINI.md on initial compile.
                     # after fix:  no GEMINI.md, log shows
                     #             [i] Targets: claude, cursor  (source: apm.yml)
                     #             [i] Compiling for AGENTS.md + CLAUDE.md (apm.yml target: [claude, cursor])
```

### Watch-vs-one-shot parity diff (the issue's literal ask)

Run `apm compile` in one directory and `apm compile --watch` in an identical sibling directory, then `diff` the outputs file-by-file. Pre-fix the watch directory has an extra `GEMINI.md`; post-fix the directories are byte-identical.

### Mid-session `apm.yml` edit reflects in next recompile

With watch running on `targets: [claude]`:
1. Confirm only `CLAUDE.md` exists.
2. Edit `apm.yml` to `targets: [claude, gemini]`, save.
3. Touch any `.instructions.md` to trigger a recompile.
4. Confirm log line changes to `[i] Targets: claude, gemini  (source: apm.yml)` and `GEMINI.md` now appears.
5. Reverse: edit back to `targets: [claude]`, delete `GEMINI.md` manually, touch again — confirm `GEMINI.md` does **not** come back.

### `--clean --watch` warning

```bash
apm compile --watch --clean
# Prints: [!] --clean is ignored in watch mode; run 'apm compile --clean' separately to remove orphaned outputs.
# ...then proceeds with normal watch (no destructive cleanup mid-session).
```

## Regression tests

`tests/unit/compilation/test_compile_watch_target_regression.py` — six tests in two classes, all of which **fail on `main` without this PR** (toggle-verified) and pass with it:

**Behavioral guards** (call `_run_compile_once` directly):
- `test_apm_yml_targets_excluding_gemini_does_not_emit_gemini_md` — the core bug.
- `test_cli_target_claude_does_not_emit_gemini_md` — CLI `--target` overrides `apm.yml`.
- `test_apm_yml_targets_including_gemini_does_emit_gemini_md` — positive companion so the negative isn't a tautology.

**Structural guards** (`inspect.getsource` on `watcher.py`):
- `test_watcher_imports_run_compile_once` — watcher must reference the shared helper.
- `test_watcher_does_not_build_compilation_config_directly` — `CompilationConfig.from_apm_yml` literal must not reappear in `watcher.py`.
- `test_watcher_does_not_instantiate_compiler_directly` — `AgentsCompiler(` must not reappear.

The structural guards exist so a future contributor copy-pasting the broken parallel path back into `watcher.py` gets a clear test failure at the exact spot, with an assertion message that points at #1345.

## Trade-offs

- [x] **Watch-mode logging is louder per recompile.** Pre-fix each recompile printed `[*] Recompiling...` plus compiler output. Post-fix each recompile additionally prints `[*] Starting context compilation...` plus the canonical `[i] Targets: ...  (source: ...)` / `[i] Compiling for ...` provenance lines. This is the explicit cost of the `--watch == apm compile` parity the issue asks for, and the provenance lines have real value — when a user edits `apm.yml`'s `targets:` mid-watch, they immediately see whether the change was picked up. The redundant banner could be suppressed via a `quiet_banner` flag if it becomes annoying, but adding the flag now would re-introduce divergence between the two surfaces; deferred.
- [x] **`--clean` is hard-coded off inside the watcher.** `--clean` removes orphaned outputs (e.g., a `GEMINI.md` left over from a previous `targets:` configuration). Running it on every recompile would surprise users mid-session, and selectively running it only on the initial compile would re-introduce a watcher-specific code path — the same trap this PR exists to remove. The warning path means users who explicitly pass `--clean --watch` know what's happening. To clean orphans, run `apm compile --clean` once between watch sessions.
- [x] **Initial compile still triggers one self-recompile.** The initial compile writes `AGENTS.md`, watchdog observes the write, debounce passes (`last_compile=0` at startup), and a redundant second compile fires. **This is pre-existing on `main` — the old watcher had the same behavior** — and is out of scope for #1345. Fix would be `event_handler.last_compile = time.time()` immediately before the initial `_do_compile()` call. Not landed here to keep this PR purely focused on the target-resolution drift.
- [x] **`compile_has_critical` local variable removed.** During code review the previous shape (`compile_has_critical = result.has_critical_security` ... mutate local ... `result.has_critical_security = compile_has_critical`) was identified as a foot-gun: future code adding a third critical-flipping branch could silently miss either the local or the result depending on insertion point. The new shape mutates `result.has_critical_security` directly at the one site that flips it (`if verdict.has_critical: result.has_critical_security = True`); the returned `result` is the single source of truth.

## Files changed

- `src/apm_cli/commands/compile/cli.py` — `_run_compile_once` extracted from `compile()`; one-shot path now calls it; watch dispatch updated to forward `target` and other flags to `_watch_mode`; `--clean --watch` warning added.
- `src/apm_cli/commands/compile/watcher.py` — `_watch_mode` signature expanded to keyword-only `target`, `output`, `chatmode`, `no_links`, `dry_run`, `single_agents`, `verbose`, `local_only`, `with_constitution`; recompile body delegates to `_run_compile_once`. `watcher.py` no longer imports `CompilationConfig` or `AgentsCompiler`.
- `tests/unit/compilation/test_compile_watch_target_regression.py` — new; six tests as described above.
- `CHANGELOG.md` — `### Fixed` entry under `[Unreleased]`.

`_resolve_compile_target`, `detect_target`, `CompilationConfig`, and `AgentsCompiler` are unchanged.

## Related

- Regresses fix for: #1019 (`--targets claude,codex` emits GEMINI.md), #1074 (`-t opencode` emits all targets).
- Builds on plural-`targets:` parsing from #1154 / #1165.


edenfunf:
Pushed fixup `3d9cc354` addressing the review:

- **Type hints on `_run_compile_once`** — added explicit annotations for all 11 parameters + `-> CompilationResult` return. Imported `CompilationResult` from `apm_cli.compilation`.
- **Em dash → ASCII hyphen** — both the `_run_compile_once` docstring and the CHANGELOG entry; matches the `RUF002` / `RUF003` ASCII convention declared (deferred) in `pyproject.toml`.
- **CHANGELOG entry** — condensed from 4 long sentences to 3, kept the WHY (was-regression-of-#1019/#1074) and the two non-obvious side effects (apm.yml re-read each recompile; `--clean --watch` warning).

The fourth comment (`parse_targets_field` in a broad `except Exception: pass`) is a verbatim relocation of the existing block from `cli.py::compile` on `main` — I did not modify the logic, only its enclosing function. Filing a follow-up issue for it makes more sense than expanding this PR's scope, since the silent-fallback affects both the one-shot and watch paths equally and predates #1345.

`ruff check`, `ruff format --check`, and the 82 target/regression tests are green on the fixup.

danielmeppiel:
Thanks for this — the core bug was independently fixed via #1349, which landed while this was in flight. Closing as superseded, but the mid-session `apm.yml` reload and the `--clean --watch` warning are real improvements we'd love to ship. Would you be up for a small follow-up PR (~100 lines) that adds just those two behaviors on top of current main? We'll fast-track review. Appreciate the thorough work here.

PR #1513, opened 2026-05-27
fix(opencode): validate-and-warn on incompatible agent frontmatter at install (Phase 1 of #581)

### LINKED ISSUE #581: Feature: per-target agent frontmatter transformation during deployment
## Problem

When APM deploys agent `.md` files from a marketplace plugin to multiple targets (Claude Code, OpenCode, Codex, etc.), it copies the same file verbatim to each target directory (`.claude/agents/`, `.opencode/agents/`, `.codex/agents/`).

This causes hard crashes on platforms with different frontmatter schemas.

### `tools` field
- **Claude Code** expects: comma-separated string or array → `tools: Read, Grep, Glob` or `tools: ["Read", "Grep"]`
- **OpenCode** expects: record/object → `tools: {Read: true, Grep: true}` (Zod schema: `z.record(z.string(), z.boolean())`)
- **Result**: OpenCode throws `InvalidError` ("expected record, received array") — **hard crash, unrecoverable**

### `color` field
- **Claude Code** expects: named colors → `color: cyan`, `color: yellow`
- **OpenCode** expects: hex `#RRGGBB` or theme names (`primary`, `secondary`, `accent`, `success`, `warning`, `error`, `info`)
- **Result**: OpenCode throws `InvalidError` ("Invalid hex color format")

### Impact on OpenCode

OpenCode's `loadAgent()` function calls `Agent.safeParse()` on the parsed frontmatter. On validation failure, it `throw`s an `InvalidError` with **no catch** — a single bad agent file kills the entire application startup. There is no graceful degradation or per-file skip on validation failure.

### APM already detects the problem but only during compile

`apm compile` warns about these files with "Failed to parse primitive ... mapping values are not allowed in this context" — but `apm install` (which deploys the files) does not validate or warn. The broken files are deployed silently.

### Option A: Transform frontmatter per target during `apm install`
During `apm install --target <platform>`, APM transforms agent frontmatter fields to the target platform's expected format:
1. **`tools`**: Convert between array/string (Claude) ↔ record (OpenCode) formats
2. **`color`**: Convert between named colors (Claude) ↔ hex/theme names (OpenCode)
3. **Unknown fields**: Strip fields not supported by the target platform

This would allow plugin authors to write agent files once in any supported format, and APM would handle the translation during deployment.

### Option B: Validate and warn during install (not just compile)
`apm install` should validate agent frontmatter against the target platform's schema and warn about incompatible fields before deploying, rather than silently deploying broken files that crash the target tool. The validation logic already exists in `apm compile` — it should also run during `apm install`.

## Related

- #293 — Support tool overrides for installed agent packages (consumer-side)
- OpenCode agent schema: `z.record(z.string(), z.boolean())` for tools, `z.union([hex regex, theme enum])` for color ([source](https://github.com/sst/opencode/blob/dev/packages/opencode/src/config/config.ts))

## Environment

- APM version: 0.8.10
- Plugin: `data-goblin/power-bi-agentic-development` v0.17.2
- Targets affected: OpenCode (confirmed hard crash), potentially Codex and Cursor

danielmeppiel:
## Context

Refs #581 (does NOT close -- Phase 2 still required to fully resolve).

OpenCode's `loadAgent()` calls `Agent.safeParse()` on parsed YAML frontmatter and `throw`s an uncaught `InvalidError` on validation failure -- one bad agent file kills OpenCode startup. `AgentIntegrator.integrate_agents_for_target` previously dispatched to `copy_agent()` for the `opencode_agent` format_id, so a Claude-style agent file (`tools` as string/array, `color` as a named color outside the OpenCode theme enum) was deployed verbatim to `.opencode/agents/` and then crashed OpenCode at startup. `apm install` did not re-run the compile-time frontmatter checks, so users had no warning.

## Scope: Phase 1 only

The issue thread agreed on a two-phase split:

- **Phase 1 (this PR):** at install time, validate each `.agent.md` frontmatter against the known Zod-fatal shapes for OpenCode and emit a loud, ASCII warning naming the offending file. The file is still copied -- warnings do NOT block install, matching the behavior the maintainer asked for in the triage thread.
- **Phase 2 (separate issue/PR):** per-target frontmatter transformer that rewrites `tools` and `color` to the target schema. Intentionally **not** implemented here.

## What changed

- New module `src/apm_cli/integration/opencode_frontmatter.py` -- pure validator returning a list of human-readable, ASCII warning messages for:
  - `tools` not a `dict[str, bool]` (list/string/other shapes are Zod-rejected).
  - `color` not a hex (`#abc` / `#aabbcc`) and not in OpenCode's theme enum (`primary, secondary, accent, success, warning, error, info`).
- `AgentIntegrator.integrate_agents_for_target` now invokes `_warn_opencode_frontmatter` in the `opencode_agent` dispatch branch **before** `copy_agent()` runs. Warnings flow through the existing `DiagnosticCollector` so they render with the same `[!]` prefix as other install warnings. No new console plumbing.
- Malformed YAML, symlink sources, and read errors are treated as silent no-ops by the validator hook (the existing security guards in `copy_agent` handle them).

## How to test

```
uv run --extra dev pytest tests/unit/integration/test_opencode_frontmatter_validation.py -v
```

Manual repro of the bug:
1. Author an agent with `tools: [Read, Grep]` and `color: cyan`.
2. `apm install -t opencode` against a project with `.opencode/` present.
3. Before this PR: install succeeds silently; OpenCode crashes on startup.
4. After this PR: install still succeeds, plus two `[!]` warnings name the offending file and the field that will make OpenCode reject the agent.

## Validation evidence

- `uv run --extra dev pytest tests/unit/integration/` -- **1449 passed**.
- `uv run --extra dev ruff check src/ tests/` -- clean.
- `uv run --extra dev ruff format --check src/ tests/` -- clean.
- `uv run --extra dev python -m pylint --disable=all --enable=R0801 --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/` -- 10.00/10, no R0801.
- `bash scripts/lint-auth-signals.sh` -- clean.
- **Mutation-break gate:** with the dispatch-site hook reverted, the new install-path tests fail (`AssertionError: []`); with the hook restored they pass.

## Trade-offs

- Warn rather than block: keeps install non-failing per the maintainer's call in the triage thread, at the cost that an unwitting user could ignore the warning and still hit the OpenCode crash.
- Hard-coded enum + regex instead of importing OpenCode's Zod schema: avoids a runtime dependency on the OpenCode source tree; the enum is small and stable (`config.ts` in `sst/opencode`).
- Validator does not mutate the frontmatter -- that's Phase 2's job. Source files remain the source of truth.

## Follow-ups

- Phase 2 (per-target transformer) still needed to fully close #581.
- The same validation idea likely applies to `commands` deployed to OpenCode, but #581 is scoped to agents and we don't have a confirmed crash report on commands yet.


danielmeppiel:

> Phase 1 of #581 lands validate-and-warn on OpenCode-incompatible agent frontmatter at install, defending multi-harness portability without breaking any existing install.

This PR converts a silent portability hazard into a visible, install-time warning: when an agent declares OpenCode-incompatible frontmatter (tools-as-list, named color outside the theme enum), the user now sees it instead of debugging a broken harness later. The panel converges cleanly. Zero blocking findings across all seven personas; every return is recommended or nit. python-architect and supply-chain-security both validated the Copilot pass on HEAD bb6bfbc3 (helper-extracted ASCII sanitization of dynamic frontmatter, schema-corrected dict|None, four new sanitization-path tests). test-coverage-expert confirmed the install-pipeline floor was met with a real mutation-break gate (integration-with-fixtures, not just unit coverage), which is the load-bearing evidence here: the user-facing promise (install succeeds AND warns on incompatible frontmatter) is asserted on this commit and holds.

The strategic call is validate-and-warn vs validate-and-block. Phase 1 is correctly scoped to warn: blocking would break installs that work today on non-OpenCode harnesses, which is precisely the multi-harness commitment the PR exists to defend. Phase 2 can revisit elevation per harness once we have signal on how often the warning fires in the wild. Note: auth-expert was intentionally not spawned (no auth surface).

The only real copy-direction tension is between cli-logging (drop the consequence trailer, keep the warning terse) and devx-ux (add a remediation pointer so the user knows where to fix it). Both can win: drop the redundant consequence, add a one-line remediation hint plus a package-qualified source path. That is a Phase 1.1 follow-up, not a merge blocker.

**Dissent.** cli-logging-expert wants the warning shortened (drop the consequence trailer); devx-ux-expert wants it lengthened (add a remediation pointer and package-qualified source path). I side with devx-ux on substance (remediation > consequence for a warning the user can act on) and with cli-logging on form (one line, not two). The synthesis is: replace the consequence trailer with a remediation hint, and qualify the file with its package.

**Aligned with:** _portable by manifest_ -- Manifest-declared agents now surface harness-incompatibility at install time instead of failing opaquely downstream.; _multi-harness, multi-host_ -- The clear principle this PR defends: OpenCode-incompatible frontmatter is named explicitly so multi-harness portability stays a first-class promise, not a debug-it-yourself contract.; _secure by default_ -- ASCII sanitization of dynamic frontmatter values (per the Copilot pass) keeps package-controlled strings from injecting into install-time output; supply-chain flagged an ANSI-escape gap in _ascii_safe_name worth tightening in follow-up.; _pragmatic as npm_ -- Warn-don't-block matches npm's deprecation-warning ergonomics: existing installs keep working, new signal is visible, escalation is a separate decision..

**Growth signal.** oss-growth-hacker flagged this as a credible portability-story beat: 'APM warns you before your agent breaks on another harness' is concrete, demoable, and reinforces the multi-harness positioning without name-dropping incumbents. Worth a release-note line and a follow-up amplification when Phase 2 (per-target transformer) lands.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 2 | Validator/dispatch split is clean and mirrors codex precedent; main architectural watch-item is the growing format_id if/elif tower in integrate_agents_for_target. |
| CLI Logging | 0 | 0 | 2 | Warnings are actionable, ASCII-safe, and route through DiagnosticCollector correctly; only minor copy nits. |
| DevX UX | 0 | 2 | 1 | Warn-don't-block is the right call for multi-agent installs, but the warning text lacks a next-action and a copy-pasteable path; doc fix-path is solid. |
| Supply Chain Security | 0 | 1 | 1 | Validator is dependency-neutral, auth-neutral, uses yaml.safe_load with proper fail-closed-to-no-warn on YAMLError; one minor asymmetry in filename sanitizer. |
| OSS Growth | 0 | 1 | 1 | Adoption-blocker removed for OpenCode users; CHANGELOG line is correct but buried -- recommend a one-liner amplification beat and a discoverability tweak in the new docs subsection. |
| Doc Writer | 0 | 1 | 2 | New skill-resource subsection is accurate and well-placed, but the user-facing Starlight page for agents still teaches the OpenCode-incompatible shape it now warns about. |
| Test Coverage | 0 | 2 | 2 | Install-pipeline floor met: pure-validator unit tests + integration-with-fixtures tests on integrate_agents_for_target form a real mutation-break gate; 27/27 pass. Two minor defensive-branch gaps. |

### Top 5 follow-ups

1. **[Doc Writer]** Starlight page docs/src/content/docs/producer/author-primitives/instructions-and-agents.md still teaches tools-as-list and labels opencode 'verbatim' without the new caveat; either annotate the row or fix the example so the docs stop teaching the exact pattern the installer now warns about. -- User-facing doc drift on the same surface this PR validates is the single most confusing thing a new author can hit. Highest-signal follow-up.
2. **[DevX UX]** Add a remediation pointer to the warning (one short line: where to fix it, or a doc URL) so the user can act, not just acknowledge. -- A warning the user cannot act on is noise. This is the highest-leverage UX edit and pairs naturally with dropping the consequence trailer cli-logging flagged.
3. **[Supply Chain Security]** Tighten _ascii_safe_name to also strip ASCII control characters (notably ESC 0x1B); current filter only excludes >0x7E, leaving room for package-controlled ANSI-escape injection into install-time output. -- Low-likelihood but real injection vector on package-controlled strings; cheap to fix and aligns with secure-by-default.
4. **[DevX UX]** Qualify the offending file in the warning by package, not just basename; with many installed packages, basename alone is not locatable. -- Pairs with the remediation pointer; both are one-line edits that turn the warning from a notification into a fix instruction.
5. **[Python Architect]** The format_id if/elif tower will grow as Phase 2 adds per-harness handlers; extract a handler map (format_id -> {pre_copy_hook?, writer}) when Phase 2 lands, not now. -- Architectural debt to track, not pay today. Premature extraction with one harness would be over-engineering; flag it for the Phase 2 PR.

### Architecture

```mermaid
classDiagram
    direction LR
    class BaseIntegrator {
      <<AbstractBase>>
      +find_files_by_glob(path, pattern)
      +sync_remove_files(...)
      #_check_adopt_or_skip(...)
    }
    class AgentIntegrator {
      +integrate_agents_for_target(target, package_info, project_root, managed_files, force, diagnostics)
      +copy_agent(source, target) int
      -_write_codex_agent(source, target)$
      -_write_windsurf_agent_skill(source, target, diagnostics)
      -_warn_opencode_frontmatter(source, diagnostics, package_name)$
      -_FRONTMATTER_RE re.Pattern
    }
    class opencode_frontmatter {
      <<Module:Pure>>
      +validate_opencode_frontmatter(fm, source) list~str~
      -_ascii_safe_name(name) str
      -_ascii_repr(value) str
      -_is_valid_opencode_color(value) bool
      -OPENCODE_THEME_COLORS frozenset
      -_HEX_COLOR_RE re.Pattern
    }
    class DiagnosticCollector {
      <<Collector>>
      +warn(message, package)
      +info(message, package)
    }
    class TargetProfile {
      <<ValueObject>>
      +primitives dict
      +root_dir str
    }
    class PrimitiveMapping {
      <<ValueObject>>
      +format_id str
      +subdir str
      +extension str
    }
    BaseIntegrator <|-- AgentIntegrator
    AgentIntegrator ..> opencode_frontmatter : delegates validation
    AgentIntegrator ..> DiagnosticCollector : pushes warnings
    AgentIntegrator ..> TargetProfile : reads
    TargetProfile *-- PrimitiveMapping : agents
    class AgentIntegrator:::touched
    class opencode_frontmatter:::touched
```

```mermaid
flowchart TD
    A[CLI: apm install] --> B[install pipeline]
    B --> C[AgentIntegrator.integrate_agents_for_target]
    C --> D{source is symlink?}
    D -- yes --> E[skip]
    D -- no --> F{ensure_path_within ok?}
    F -- no --> G[diagnostics.warn rejected path]
    F -- yes --> H{_check_adopt_or_skip}
    H -- proceed --> J{mapping.format_id?}
    J -- codex_agent --> K[_write_codex_agent MD to TOML]
    J -- windsurf_agent_skill --> L[_write_windsurf_agent_skill]
    J -- opencode_agent --> M[_warn_opencode_frontmatter]
    J -- else --> N[copy_agent]
    M --> M5[yaml.safe_load]
    M5 -- ok dict --> M6[validate_opencode_frontmatter pure]
    M6 --> M7{messages empty?}
    M7 -- yes --> N
    M7 -- no --> M8[diagnostics.warn per message]
    M8 --> N
    N --> Z[files_integrated plus 1]
    K --> Z
    L --> Z
    Z --> R[IntegrationResult]
```

### Recommendation

Ship it. Zero blocking findings, lint clean, tests green, Copilot's six inline items already addressed on HEAD bb6bfbc3, and test-coverage confirmed a real mutation-break gate on the install pipeline. Track the doc-writer follow-up as the highest-signal next edit: the Starlight authoring page still teaches the exact frontmatter shape the installer now warns against, and fixing that closes the loop between the validator and the docs that produce its inputs.

---

<details>

#### Python Architect

- **[recommended]** Per-format dispatch in integrate_agents_for_target is becoming an if/elif tower; extract a handler map before Phase 2 lands. at `src/apm_cli/integration/agent_integrator.py:168`
  The inner loop in AgentIntegrator.integrate_agents_for_target now branches on mapping.format_id three ways (codex_agent -> _write_codex_agent, windsurf_agent_skill -> _write_windsurf_agent_skill, opencode_agent -> _warn_opencode_frontmatter then copy_agent, else copy_agent). Phase 2's per-target frontmatter transformer (#581) will add at least one more branch (opencode_agent -> transform-then-copy) plus likely future targets. Each new format currently requires reordering this nested if/elif and risks dropping the pre-copy hook (the opencode warn lives inside the 'else' branch today and would be silently bypassed if a future format_id is added that takes the codex/windsurf path but also needs warnings). Pushing this to a small format_id -> {pre_copy_hook?, writer} table (or a polymorphic FormatHandler protocol next to TargetProfile.primitives) keeps O(work) per new target proportional to one registry entry, not a re-edit of the dispatch chain. Design patterns: Used in this PR: Pure function (validate_opencode_frontmatter is side-effect-free and reusable for tests/Phase 2 transformer); Adapter (_warn_opencode_frontmatter adapts the pure validator to the DiagnosticCollector + file-on-disk surface). Pragmatic suggestion: Strategy / Registry keyed on format_id -- one entry per target with optional pre_copy_hook + required writer; would absorb both the codex/windsurf writers and the new opencode warn, and is the only refactor that pays off at the size Phase 2 will push this loop to.
  *Suggested:* When introducing the Phase 2 transformer, replace the nested if/elif with a class-level dict like _FORMAT_HANDLERS: dict[str, FormatHandler] = {'codex_agent': FormatHandler(writer=_write_codex_agent), 'windsurf_agent_skill': FormatHandler(writer=_write_windsurf_agent_skill), 'opencode_agent': FormatHandler(pre_copy_hook=_warn_opencode_frontmatter, writer=copy_agent)} and look up handler = self._FORMAT_HANDLERS.get(mapping.format_id, _DEFAULT_HANDLER).

- **[nit]** Source file is read twice on the opencode path (_warn_opencode_frontmatter then copy_agent). at `src/apm_cli/integration/agent_integrator.py:280`
  _warn_opencode_frontmatter calls source.read_text() to extract frontmatter, and copy_agent immediately reads the same file again. For typical agent files this is negligible (OS page cache absorbs it) but if you extract the handler map per the recommended finding, threading the already-read content through avoids the duplicate I/O for free.

- **[nit]** YAMLError and OSError in _warn_opencode_frontmatter are silently swallowed with no diagnostic. at `src/apm_cli/integration/agent_integrator.py:289`
  The early returns on OSError and yaml.YAMLError are correct for warn-and-continue semantics (copy_agent will surface a real error if the file is unreadable), but a single diagnostics.info() on YAML parse failure would help users debugging why no opencode warning fired on a file they expected to flag. Skip if you consider this noise.

#### CLI Logging

- **[nit]** Trailing 'OpenCode will reject this agent at load time.' duplicates per-warning when multiple shapes are wrong on the same file. at `src/apm_cli/integration/opencode_frontmatter.py:82`
  Each message ends with the same consequence sentence; an agent with both bad tools and bad color renders two yellow lines that each repeat it. Newspaper rule: state consequence once, lead with the fact. Could be folded into the verbose-only detail field of diagnostics.warn() so default output stays crisp while --verbose still spells the consequence out.
  *Suggested:* Drop the trailing sentence from each message and pass it (or a single 'OpenCode will refuse to load this agent') via the detail= kwarg of diagnostics.warn so it only surfaces under --verbose, matching how _render_warning_group already gates detail on self.verbose.

- **[nit]** 'tools as NoneType' reads awkwardly when frontmatter has 'tools: null'. at `src/apm_cli/integration/opencode_frontmatter.py:78`
  type(tools).__name__ returns 'NoneType' / 'str' / 'list'. 'NoneType' is a Python implementation token, not user vocabulary; a Claude-style author writing 'tools: null' will see 'has tools as NoneType' which obscures the real fix (use a mapping). Minor polish; the rest of the message still points the user at the right shape.
  *Suggested:* Map type names to user-facing labels (None -> 'null', str -> 'string', list -> 'list', etc.) before interpolation, or special-case tools is None with 'has tools set to null'.

#### DevX UX

- **[recommended]** Warning tells the user what's wrong but not what to do next or where the doc fix-path lives. at `src/apm_cli/integration/opencode_frontmatter.py:227`
  npm/cargo/pip warnings consistently end with a one-shot remediation ('run X', 'see Y'). The current message ends with 'OpenCode will reject this agent at load time.' -- a diagnosis, not a next action. The fix path exists in packages/apm-guide/.apm/skills/apm-usage/package-authoring.md (the new 'OpenCode target: frontmatter constraints' subsection), but a user reading install output has no pointer to it. For a warn-don't-block contract where exit code is 0, the warning IS the entire UX surface; if the user skims past it, the next signal is OpenCode crashing at startup with no APM context. Adding 'Edit the source file and reinstall; see apm docs (package-authoring -> OpenCode target frontmatter constraints).' (or a stable docs URL) brings this in line with the persona's Recovery lens: every error names one concrete next action one copy-paste away.
  *Suggested:* Append a fixed trailer to each message, e.g. 'Edit the source and reinstall. See: https://microsoft.github.io/apm/reference/package-authoring/#opencode-target-frontmatter-constraints' (or the local equivalent). Apply uniformly to both tools and color branches so the trailer is one constant string, not duplicated phrasing.

- **[recommended]** Warning identifies the offending file only by basename; user with multiple packages cannot jump straight to the source. at `src/apm_cli/integration/opencode_frontmatter.py:217`
  Messages currently read 'OpenCode agent demo.agent.md has tools as list...'. The package name IS attached to the diagnostic via diagnostics.warn(package=package_name), but the message STRING itself -- which is what a user copy-pastes, shares in an issue, or greps in CI logs -- carries only the filename. When a project has many APM dependencies, 'demo.agent.md' alone does not locate the file; the user has to cross-reference the renderer's package column (which may or may not be prominent in the install summary). npm/pip surface the package origin inline in the message itself for exactly this reason. Including 'apm_modules/<package>/<relpath>' (or at minimum '<package>: <relpath>') in the message string makes the warning self-contained.
  *Suggested:* Pass the package name (and ideally a portable_relpath() from project_root) into validate_opencode_frontmatter() and interpolate it into the message, e.g. "OpenCode agent at 'apm_modules/<pkg>/.apm/agents/demo.agent.md' has tools as list; ...". Keeps the diagnostics.warn(package=...) metadata for the rendered table AND makes the message string standalone.

- **[nit]** Symlinked agents skip validation silently; same crash will hit OpenCode but the user gets zero warning. at `src/apm_cli/integration/agent_integrator.py:290`
  _warn_opencode_frontmatter() early-returns on source.is_symlink(). The copy_agent path still deploys the symlink target, so the same incompatibility crashes OpenCode -- but without any APM warning, the user's mental model ('install was clean') is wrong. Acceptable for Phase 1 (symlinks are uncommon in agent packages) but worth a follow-up: either dereference the symlink for validation, or emit a single 'symlinked OpenCode agent at <path>; APM did not validate frontmatter' info-level note so the silence is at least acknowledged.

#### Supply Chain Security

- **[recommended]** _ascii_safe_name lets ASCII control chars (ESC 0x1B, BEL, etc.) pass through; package-controlled filenames could inject ANSI escapes into the warning. at `src/apm_cli/integration/opencode_frontmatter.py:188`
  _ascii_safe_name uses .encode('ascii','replace') which only replaces codepoints > 127. ASCII control characters (0x00-0x1F, 0x7F) including ESC (0x1B) are valid ASCII and pass through unchanged. The sibling helper _ascii_repr correctly escapes them via ascii() (verified: ascii('\x1b[31m') -> '\\x1b[31m'). Because a third-party package controls the agent filename written to disk and POSIX filesystems permit control characters in filenames, a crafted name like evil\x1b[31m.agent.md would be interpolated verbatim into the warning string and then rendered through the terminal by DiagnosticCollector, potentially injecting ANSI color/cursor sequences in CLI output. This is the same class of risk the PR already mitigates for tool keys/values and color values. Exploitability is bounded (cosmetic ANSI injection at install time on POSIX; the install does not execute attacker code) so this is recommended, not blocking. Treat as part of APM's untrusted input rendering contract.
  *Suggested:* Make _ascii_safe_name strip ASCII control characters in addition to non-ASCII bytes, e.g.: return ''.join(c if 0x20 <= ord(c) < 0x7F else '?' for c in name.encode('ascii','replace').decode('ascii')). Add a unit test asserting that a source path containing ESC (0x1B) in the stem produces a warning with no raw control bytes (assert '\x1b' not in msg).

- **[nit]** YAMLError fallback is silent; consider a one-line debug note for diagnosability. at `src/apm_cli/integration/agent_integrator.py:131`
  yaml.safe_load correctly enforces the no-arbitrary-construction contract (no yaml.load, no eval/exec). On YAMLError the hook returns without emitting any diagnostic. This is fail-open from a do we warn the user? angle: a package author with malformed YAML escapes the new OpenCode validator entirely and OpenCode will still crash at runtime, which is exactly the failure mode #581/Phase 1 set out to surface. Not a security issue (no code execution, no integrity bypass; the file is copied verbatim through the normal copy_agent path which still applies path_security), but the user-facing diagnostic gap is worth a debug-level note so the silent return is observable.
  *Suggested:* Optional: emit a low-severity diagnostic (could not parse YAML frontmatter for {package_name}/{source.name}; skipping OpenCode frontmatter validation) so that a malformed-YAML package is at least noticed at install time. Keep the file-copy path unchanged.

#### OSS Growth

- **[recommended]** Surface this fix as a release beat -- 'APM warns before OpenCode rejects your agent' is a story-shaped multi-harness proof point. at `CHANGELOG.md`
  Multi-harness portability is one of APM's headline promises (per README/why-apm). Silent install + runtime crash on a competing harness is precisely the failure mode that erodes the 'works across Claude/Copilot/Codex/OpenCode' narrative; the fix is the inverse story and worth amplifying beyond a single CHANGELOG Fixed bullet. Existing entry is accurate but reads as maintenance; a tweet-sized line ('apm install now tells you why OpenCode would reject your agent -- before you ship it') plus inclusion in the next release notes converts a bug fix into a trust signal for OpenCode-curious users arriving from issue #581.
  *Suggested:* When the next release post is drafted, lift this entry into a 'Multi-harness reliability' beat with a 2-line example showing the warning text. No code change needed in this PR.

- **[nit]** Docs subsection is discoverable for authors already in package-authoring.md; add a one-line back-reference from the OpenCode target row in the targets/runtimes docs so authors hit it before they hit the install warning. at `packages/apm-guide/.apm/skills/apm-usage/package-authoring.md`
  The new 'OpenCode target: frontmatter constraints' block lives under 'Agent format' inside package-authoring.md. An author writing their first .agent.md for OpenCode will more likely arrive via a targets/runtimes page than scroll the authoring guide. A single cross-link ('See: OpenCode frontmatter constraints') compounds: it moves the constraint from 'discovered at install warn' to 'discovered while writing', which is the friction-reduction lens.

#### Doc Writer

- **[recommended]** Starlight agents page still shows tools: as a list and describes opencode as 'verbatim' with no caveat about Zod-schema rejection. at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md:170`
  The new constraints live only in packages/apm-guide/.apm/skills/apm-usage/package-authoring.md, which is a skill resource consumed by the apm-guide agent -- not a user-facing doc. The user-facing agent authoring page in docs/src/content/docs/producer/author-primitives/instructions-and-agents.md is where someone wiring an OpenCode-targeted agent actually looks. That page (a) shows an example with 'tools:\n  - read\n  - grep' (lines 130-132), which is exactly the list shape OpenCode rejects and the new validator warns about; (b) the prose at lines 146-149 says 'model and tools reach Copilot, Claude, Cursor, and OpenCode verbatim' without noting that OpenCode then refuses to load the agent; (c) the 'What compiles where' table at line 170 lists opencode as 'verbatim' with no asterisk pointing at the new constraint. Per doc-sync Rule 1, code changes that affect what a user sees at install time must update docs/src/content/docs/. Per Rule 3 the patch should stay pragmatic -- a one-line caveat in the opencode row plus a short note under the 'tools' key (or replacing the example with the dict-of-bool form) is enough. The skill-resource update is correct under Rule 4 and should stay; this finding is about the parallel user-doc surface that was missed.
  *Suggested:* Either (a) annotate the opencode row in the 'What compiles where' table with '* OpenCode requires tools: as a name->bool mapping and color: as hex or theme token; mismatches install with a warning and then fail to load in OpenCode -- see package-authoring resource', or (b) add a one-paragraph 'OpenCode frontmatter constraints' note after the table linking to the same authoring guidance, AND change the 'tools: - read - grep' example (lines 130-132) to the dict form so the canonical example does not teach an OpenCode-incompatible shape.

- **[nit]** New skill-resource subsection is accurate against the validator (theme enum, #abc/#aabbcc regex, warn-not-block) and placed correctly under section 5 'Agent'. at `packages/apm-guide/.apm/skills/apm-usage/package-authoring.md:252`
  Voice, bold-key formatting, and the trailing 'until APM ships a per-target frontmatter transformer' caveat match the surrounding workflow-frontmatter and copilot-app subsections. Theme tokens enumerated in the doc (primary, secondary, accent, success, warning, error, info) match OPENCODE_THEME_COLORS in src/apm_cli/integration/opencode_frontmatter.py. The hex shapes '#abc' and '#aabbcc' match _HEX_COLOR_RE. The 'file is still copied so you can fix it in place, but OpenCode will fail to start until you do' framing matches the warn-not-block dispatch in AgentIntegrator._warn_opencode_frontmatter. No nit on the subsection itself; recording this as positive signal so the CEO can weigh the drift finding against the rest of the doc surface being clean.

- **[nit]** CHANGELOG Unreleased > Fixed entry is well-formed (concise, accurate, links #1513 and #581). at `CHANGELOG.md:21`
  Entry sits under '### Fixed' in '## [Unreleased]', uses backticks for paths and YAML keys per .apm/instructions/changelog.instructions.md, ends with the '(Phase 1 of #581, #1513)' citation, and accurately summarises the user-visible behavior (warn at install, file still copied, transformer tracked separately). No change needed.

#### Test Coverage

- **[recommended]** Install-path tests certify the OpenCode validate-and-warn promise at the install-pipeline floor and form a real regression trap. at `tests/unit/integration/test_opencode_frontmatter_validation.py`
  The dispatch-site hook is on a critical-promise surface (install pipeline). TestOpencodeInstallEmitsWarnings drives integrator.integrate_agents_for_target against a real tempdir, a real .opencode marker, a real package on disk with real frontmatter, and a real DiagnosticCollector - no mocks at the asserted boundary. The assertions probe the diagnostic stream AND verify the file is still deployed, so deleting the _warn_opencode_frontmatter call in agent_integrator.py would make test_tools_as_list_emits_warning and test_color_named_emits_warning fail (mutation-break gate holds). The parametrized non-OpenCode test additionally pins that the warning is scoped to format_id == opencode_agent. Surfaced as a finding (rather than silently passing) so the CEO sees the load-bearing evidence.
  *Proof (passed, integration-with-fixtures):* `tests/unit/integration/test_opencode_frontmatter_validation.py::TestOpencodeInstallEmitsWarnings::test_tools_as_list_emits_warning` -- proves: When a package author ships an OpenCode-incompatible agent (tools as list, named color), apm install writes the file AND surfaces a diagnostic warning naming the file and the bad field, so the user learns why OpenCode will reject it. [devx,multi-harness-support]
  `assert result.files_integrated == 1; assert any('tools' in m and 'demo.agent.md' in m for m in msgs), msgs`

- **[recommended]** Pure-validator branch coverage is thorough and ASCII-sanitization paths are covered at the unit tier. at `src/apm_cli/integration/opencode_frontmatter.py`
  TestValidateOpencodeFrontmatter exercises tools-as-dict OK, tools-as-list, tools-as-string, tools-as-dict-with-non-bool, hex color, short hex, every theme enum value, named-not-in-enum, non-string color, empty/None frontmatter, multi-problem accumulation, ASCII-only output, plus the three sanitization paths added in the Copilot-review pass (non-ASCII filename via _ascii_safe_name, non-ASCII color via _ascii_repr, non-ASCII tool key+value). The fm=None contract (annotation dict | None) has an explicit test. No genuine branch in opencode_frontmatter.py is unasserted.
  *Proof (passed, unit):* `tests/unit/integration/test_opencode_frontmatter_validation.py::TestValidateOpencodeFrontmatter` -- proves: The pure validator emits ASCII-only diagnostics for every documented OpenCode-incompatible shape and accepts None/empty frontmatter without raising. [devx]
  `assert '\\xe1' in msgs[0]  # non-ASCII color repr; msgs[0].encode('ascii')  # raises if non-ASCII`

- **[nit]** No integration-with-fixtures coverage for the end-to-end install path under a non-ASCII agent filename. at `tests/unit/integration/test_opencode_frontmatter_validation.py`
  _ascii_safe_name is exercised at the unit tier (test_non_ascii_filename_sanitized calls the validator directly with Path('cy\u00e1n-agent.md')), but no install-path test writes a non-ASCII filename into a package directory and drives integrate_agents_for_target to confirm the full chain (frontmatter read + warning emission + diagnostics.warn) stays ASCII-clean on Windows cp1252. Low value: the install path passes the same Path object straight to the unit-tested validator and the diagnostic message is constructed there; the only added risk is filesystem encoding on the read_text() side, which would surface as an OSError that the code already swallows. Surfacing only as nit because the unit assertion essentially covers the message shape.
  *Proof (missing, integration-with-fixtures):* `tests/unit/integration/test_opencode_frontmatter_validation.py::TestOpencodeInstallEmitsWarnings::test_non_ascii_filename_install_path (would-be)` -- proves: The full install pipeline keeps diagnostics ASCII-only even when the source agent file has a non-ASCII filename, guaranteeing no UnicodeEncodeError on Windows cp1252 terminals. [devx,multi-harness-support]
  `pkg = self._write_agent_named('cy\u00e1n.agent.md', 'tools:\n  - Read\n'); self.integrator.integrate_agents_for_target(...); for m in _warning_messages(diagnostics): m.encode('ascii')`

- **[nit]** Defensive symlink and OSError no-op branches in _warn_opencode_frontmatter are not asserted. at `src/apm_cli/integration/agent_integrator.py:278`
  agent_integrator.py:278 short-circuits when source.is_symlink() and :282 swallows OSError from read_text; neither branch has a test that would notice if the guard were removed. test_malformed_yaml_does_not_crash covers the yaml.YAMLError path. These are defensive no-ops on a non-critical sub-surface (the symlinked-source case is rare in real .apm/agents trees), so silent-drift severity is nit per the tier-floor matrix. A one-line parametrize entry adding a symlinked agent file and asserting no warnings + no crash would close the gap.
  *Proof (missing, integration-with-fixtures):* `tests/unit/integration/test_opencode_frontmatter_validation.py::TestOpencodeInstallEmitsWarnings::test_symlinked_agent_skipped_no_crash (would-be)` -- proves: Symlinked agent sources do not crash the install pipeline and do not emit OpenCode frontmatter warnings (defensive no-op). [devx]
  `real_pkg/.apm/agents/real.agent.md; sym = pkg/.apm/agents/link.agent.md; sym.symlink_to(real); integrate_agents_for_target(...); assert _warning_messages(diagnostics) == []`

</details>

danielmeppiel:

> Phase 1 validate-and-warn ships clean: 8 panelists, zero blocking findings, all wave-1 followups folded, CI green at 1459 tests.

All eight panelists converge on the same signal: this PR is well-scoped Phase 1 work that delivers a non-blocking install-time safety net for OpenCode users without introducing new attack surface, architectural debt, or UX regressions. The python-architect confirms correct extraction of a pure validator with a clean Phase 2 boundary. Supply-chain security validates the symlink guard, safe_load usage, and ANSI-stripping defenses. CLI-logging and DevX-UX both affirm the warning format follows APM conventions (outcome-first, package-qualified, actionable Fix pointer) with only cosmetic polish suggestions.

No dissent exists between panelists. The test-coverage-expert surfaced one missing-evidence finding (diagnostics=None guard untested) but classified it as nit-severity given the low likelihood of that code path being exercised without a collector today. That guard protects against a crash and the install still succeeds -- it does not meet the threshold for a blocking signal.

Strategically, this PR is the first concrete proof point for APM's multi-target safety-net positioning. It ships a user-visible improvement (catching broken agents before OpenCode does) with zero breaking-change cost and a one-line CHANGELOG story. The wave-2 iteration folded all four CEO-recommended followups from wave 1, demonstrating strong iteration discipline. The deferred format_id handler-map refactor is correctly scoped to Phase 2.

**Aligned with:** Multi-harness/multi-host, Secure by default, Pragmatic as npm, OSS community-driven

**Growth signal.** OpenCode validate-and-warn is APM's first shipped proof point for multi-target safety-net positioning. When Phase 2 (auto-transform) lands, this becomes a launch beat: 'APM catches what would crash your agent before it ever loads.' OpenCode Discord and r/opencode are seeding channels. The CHANGELOG entry should lead with user benefit for release-note mining: 'APM now catches agent frontmatter that would crash OpenCode at startup and tells you exactly what to fix.'

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 2 | Clean pure-validator extraction with correct Phase 1 scope boundary; no architectural concerns after wave-2 hardening. |
| CLI Logging Expert | 0 | 0 | 2 | Warnings follow APM conventions well: outcome-first, package-qualified identifier, actionable Fix pointer, ASCII-safe. Two polish nits on the color warning. |
| DevX UX Expert | 0 | 0 | 3 | Solid validate-and-warn UX: non-blocking install with actionable Fix pointers meets npm-style expectations. No blocking issues. |
| Supply Chain Security Expert | 0 | 0 | 1 | Phase 1 is read-only validation with no new attack surface; symlink guard, safe_load, and ANSI defenses are sound. |
| OSS Growth Hacker | 0 | 0 | 2 | Phase 1 ships a clear install-time safety net for OpenCode users; CHANGELOG entry is strong but could lead with user benefit. |
| Auth Expert | -- | -- | -- | No auth surface touched; no auth findings. |
| Doc Writer | 0 | 0 | 2 | Doc updates are accurate, well placed, and consistent with APM voice; one minor gap: color is referenced but not listed in the agent frontmatter key table. |
| Test Coverage Expert | 0 | 0 | 1 | Strong coverage at integration-with-fixtures tier; cross-module flow and key user promises defended. One nit for untested diagnostics=None defensive guard. |

### Top 5 follow-ups

1. **[Test Coverage Expert]** Add test for diagnostics=None early-return guard in _warn_opencode_frontmatter -- Missing coverage on a defensive guard; low risk today but prevents silent drift if a future caller omits the collector.
2. **[CLI Logging Expert]** Fix pointer should mention '#rgb' short-hex in addition to '#rrggbb' -- Validator accepts 3-char hex but Fix text only mentions 6-char; mismatch between behavior and remediation guidance.
3. **[OSS Growth Hacker]** Link #581 in package-authoring.md where Phase 2 transformer is mentioned -- Contributor funnel: interested producers can self-assign Phase 2 work.
4. **[Doc Writer]** Add 'color' row to agent frontmatter key table in instructions-and-agents.md -- Producers hitting the color warning have no canonical reference for what the key does or what values are valid.
5. **[DevX UX Expert]** Phase 2: expose validate_opencode_frontmatter via a non-mutating command (apm check / --dry-run) -- CI lint gates and pre-commit users need to discover warnings without running a full install.

### Architecture

```mermaid
classDiagram
    direction LR
    class BaseIntegrator {
        <<Abstract>>
        +find_files_by_glob()
        +resolve_links()
        +sync_remove_files()
    }
    class AgentIntegrator {
        +integrate_agents_for_target()
        +copy_agent(source, target) int
        -_warn_opencode_frontmatter(source, diag, pkg)$ void
        -_FRONTMATTER_RE Pattern
    }
    class DiagnosticCollector {
        <<Collect-then-render>>
        +warn(message, package)
        +render_summary()
    }
    class validate_opencode_frontmatter {
        <<Pure Function>>
        +__call__(fm, source, package_name) list~str~
    }
    class _ascii_safe_name {
        <<Pure Function>>
    }
    class _is_valid_opencode_color {
        <<Pure Function>>
    }
    class OPENCODE_THEME_COLORS {
        <<frozenset>>
    }
    BaseIntegrator <|-- AgentIntegrator
    AgentIntegrator ..> validate_opencode_frontmatter : delegates validation
    AgentIntegrator ..> DiagnosticCollector : emits warnings
    validate_opencode_frontmatter ..> _ascii_safe_name : sanitizes identifiers
    validate_opencode_frontmatter ..> _is_valid_opencode_color : checks color
    _is_valid_opencode_color ..> OPENCODE_THEME_COLORS : membership test
    class AgentIntegrator:::touched
    class validate_opencode_frontmatter:::touched
    class _ascii_safe_name:::touched
    class _is_valid_opencode_color:::touched
    class OPENCODE_THEME_COLORS:::touched
```

```mermaid
flowchart TD
    A["apm install -t opencode"] --> B["integrate_agents_for_target()"]
    B --> C{"format_id == opencode_agent?"}
    C -- No --> D["copy_agent / other targets"]
    C -- Yes --> E["_warn_opencode_frontmatter()"]
    E --> F{"diagnostics None or symlink?"}
    F -- Yes --> G["skip validation"]
    F -- No --> H["source.read_text()"]
    H --> I{"frontmatter regex match?"}
    I -- No --> G
    I -- Yes --> J["yaml.safe_load(frontmatter)"]
    J --> K["validate_opencode_frontmatter() -- pure, no I/O"]
    K --> L{"warnings returned?"}
    L -- Yes --> M["diagnostics.warn() per message"]
    L -- No --> N["no-op"]
    M --> O["copy_agent()"]
    N --> O
    G --> O
```

```mermaid
sequenceDiagram
    participant AI as AgentIntegrator
    participant OC as opencode_frontmatter
    participant DC as DiagnosticCollector
    participant FS as Filesystem
    CLI->>AI: integrate_agents_for_target()
    AI->>FS: source.read_text()
    FS-->>AI: content
    AI->>OC: validate_opencode_frontmatter(fm, source, pkg)
    OC-->>AI: [warnings]
    AI->>DC: diagnostics.warn(message)
    AI->>FS: copy_agent(source, target)
    AI-->>CLI: IntegrationResult
```

### Recommendation

All wave-1 followups are folded, all 8 panelists return nit-only findings, CI is green at 1459 tests, and the PR delivers a clear user-visible improvement with zero breaking-change cost. The remaining followups are polish and Phase 2 prep -- none gate shipping. Merge at maintainer discretion.

---

<details>

#### Python Architect

- **[nit]** Double read of source file: _warn_opencode_frontmatter reads then copy_agent reads again at `src/apm_cli/integration/agent_integrator.py`
  The install path reads source.read_text() in _warn_opencode_frontmatter, then copy_agent reads the same file again. For Phase 1 fine -- validation intentionally avoids coupling to copy path. When Phase 2 adds a transformer, the shared read-parse-frontmatter preamble should be extracted.
  *Suggested:* Phase 2 prep: extract a _read_frontmatter(source) helper that both _warn and the future transformer share.
- **[nit]** _ascii_repr is a trivial wrapper over builtin ascii(); consider inlining at `src/apm_cli/integration/opencode_frontmatter.py`
  The function body is return ascii(value). Defensible for docstring clarity but borderline over-abstraction at one call site.

#### CLI Logging Expert

- **[nit]** Color warning embeds Python list repr for theme enum at `src/apm_cli/integration/opencode_frontmatter.py`
  sorted(OPENCODE_THEME_COLORS) renders as ['accent', 'error', ...] with Python-style brackets and quotes. A comma-separated list without brackets would feel more natural in CLI output.
  *Suggested:* Join as plain comma list: "one of 'accent', 'error', 'info', ..." without Python list brackets.
- **[nit]** Fix pointer says '#rrggbb' but short hex '#rgb' is also valid per the validator at `src/apm_cli/integration/opencode_frontmatter.py`
  _HEX_COLOR_RE accepts #RGB (3-char hex) but Fix only mentions '#rrggbb'. Minor mismatch between validator behavior and remediation text.
  *Suggested:* Fix: replace the color with a '#rrggbb' (or '#rgb') hex literal or one of the listed theme names.

#### DevX UX Expert

- **[nit]** No non-mutating path (--dry-run / apm check) to discover warnings without installing
  CI lint gates or pre-commit users have no way to discover frontmatter issues without running a full install. Noting for Phase 2 planning.
  *Suggested:* Track as Phase 2 follow-up: expose validate_opencode_frontmatter via a non-mutating command surface.
- **[nit]** Warning message repeats 'OpenCode will reject this agent at load time' -- slightly verbose in multi-warning scenarios at `src/apm_cli/integration/opencode_frontmatter.py`
  Each warning contains 4 clauses. The rejection consequence is implied by context. Minor noise in multi-agent-file scenarios.
- **[nit]** Fix pointer for color says 'listed theme names' but the list appears in the prior clause, not directly after Fix: at `src/apm_cli/integration/opencode_frontmatter.py`
  A user scanning for just the Fix: line might miss the valid theme options.
  *Suggested:* Append theme list after Fix: e.g. 'Fix: use #rrggbb or one of: primary, secondary, accent, success, warning, error, info.'

- **[nit]** No file-size cap before read_text() -- maliciously large file could cause memory exhaustion at `src/apm_cli/integration/agent_integrator.py`
  source.read_text() reads entire file. A multi-GB file could OOM the process. Low likelihood but cheap to defend.
  *Suggested:* Add size check: if source.stat().st_size > 1_048_576: return

#### OSS Growth Hacker

- **[nit]** CHANGELOG entry buries the hook -- leads with implementation detail rather than user benefit
  The CHANGELOG opens with 'apm install against target: opencode now warns...' which is correct but reads as a patch note. For release-note mining, leading with the user pain eliminated would make it repostable.
  *Suggested:* Lead with: 'APM now catches agent frontmatter that would crash OpenCode at startup and tells you exactly what to fix.'
- **[nit]** Docs mention Phase 2 transformer without linking to #581 -- misses contributor funnel opportunity
  package-authoring.md says 'until APM ships a per-target frontmatter transformer' but doesn't link the tracking issue.
  *Suggested:* Add '(tracked in #581)' after the transformer mention in package-authoring.md.

#### Auth Expert -- inactive

No auth-related files in diff; PR changes frontmatter validation and warning emission, not token management, credential resolution, or host classification.

#### Doc Writer

- **[nit]** Agent frontmatter key table omits `color` as a recognized key at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  The new constraints paragraph tells producers that color must be a hex literal or theme name on OpenCode, but the preceding frontmatter key table never introduces color. A producer hitting the warning for the first time has no canonical row for this key.
  *Suggested:* Add a color row to the agent frontmatter table: 'Display tint used by harnesses that render persona swatches (e.g., OpenCode)'.
- **[nit]** compile-targets table opencode row says 'verbatim' with no hint that install-time validation runs at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  Producers who skim the compile-targets table never see that opencode triggers install-time warnings. A four-word qualifier in the Transform column would catch the skimmer.
  *Suggested:* Change opencode Transform cell from 'verbatim' to 'verbatim; install-time schema check'.

#### Test Coverage Expert

- **[nit]** diagnostics=None early-return in _warn_opencode_frontmatter has no direct test at `tests/unit/integration/test_opencode_frontmatter_validation.py`
  Line 276-277 silently skips validation when diagnostics is None. If a future caller omits diagnostics, the validation becomes a silent no-op with no test catching the drift. Low risk today.
  *Suggested:* Add a test that calls integrate_agents_for_target with diagnostics=None and asserts files_integrated == 1 (file still deployed, no exception).
  *Proof (missing):* `tests/unit/integration/test_opencode_frontmatter_validation.py::test_diagnostics_none_does_not_crash` -- proves: Install pipeline does not crash when diagnostics collector is not provided

</details>

danielmeppiel:

> Ship: Phase-1 frontmatter validator is clean, tested (34/34 green), and three shepherd passes have folded all prior recommendations.

Three shepherd iterations have already folded six Copilot inline items plus four CEO recommended followups plus four wave-3 nits. The PR is converged. All eight panelists returned ship-ready assessments; zero blocking findings; zero required changes. The sole recommended-tier finding (doc-writer: canonical tools example allegedly switched to OpenCode mapping form) does not hold on current HEAD -- the example at instructions-and-agents.md:130-132 remains the list form, and the surrounding prose is consistent. The finding is a phantom; no action required.

The remaining nits cluster around two themes: (1) `sorted()` list repr producing Python brackets in the color warning message (cli-logging-expert and devx-ux-expert independently flagged it), and (2) minor doc cross-linking and deduplication. Both are polish that can land in a fast follow without any risk to the user promise. Architecture is sound (pure validator, correct hook placement, no heavy imports), supply-chain surface unchanged, and the test-coverage expert confirmed all critical promises covered at integration-with-fixtures tier.

**Dissent.** Doc-writer recommended finding ('canonical tools example switched to OpenCode mapping form') was investigated and found inaccurate on commit eb0bb0d1 -- example remains in list form. Downgraded to no-action. No other inter-panelist dissent; cli-logging-expert and devx-ux-expert independently flagged the same sorted() repr nit, confirming it as highest-signal polish item.

**Aligned with:** multi-harness-multi-host (validator warns without blocking install), pragmatic-as-npm (warn-not-block mirrors npm peer-dependency warnings), portable-by-manifest (frontmatter validated against target schema at install time), oss-community-driven (Phase 2 issue #581 tagged "contributions welcome" with clear scope).

**Growth signal.** Frame: 'APM now catches what OpenCode can't explain' -- turns a silent churn moment (agent fails to load in OpenCode with no error) into an actionable warning at install time. Ensure #581 carries the 'help wanted' label to convert debugging users into Phase-2 contributors.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 1 | Clean Phase-1 architecture: pure validator module, correct hook placement in format_id dispatch, good separation of I/O from logic. Ship-ready. |
| CLI Logging Expert | 0 | 0 | 1 | Warning output is well-structured, ASCII-safe, actionable, and consistent with DiagnosticCollector patterns. No blocking issues. |
| DevX UX Expert | 0 | 0 | 2 | Warn-not-block pattern is sound; warnings are actionable with fix pointers and package-qualified identifiers. Ship-ready for Phase 1. |
| Supply Chain Security | 0 | 0 | 1 | No supply-chain risks introduced; validator is read-only, uses safe_load, sanitizes all interpolated values, and skips symlinks. |
| OSS Growth Hacker | 0 | 0 | 3 | Good contributor-funnel language and clear user-facing warning; CHANGELOG entry is dense but accurate; minor polish opportunities. |
| Doc Writer | 0 | 1 | 2 | Docs accurate against the validator; one recommended finding investigated and found inaccurate on current HEAD (no action needed). |
| Test Coverage Expert | 0 | 0 | 1 | All critical surfaces covered at correct tier; 34/34 pass. No gaps found. Ship. |

### Top 3 follow-ups

1. **[CLI Logging Expert]** Replace sorted() list repr with ', '.join() in color warning -- Python brackets in user-facing terminal output break the established pattern of clean comma-separated lists elsewhere in APM warnings.
2. **[Doc Writer]** Deduplicate theme-color enum to single source in instructions-and-agents.md -- Seven theme names appear three times on the same page creating drift risk on any future palette change.
3. **[OSS Growth Hacker]** Add before/after YAML snippet to OpenCode pitfall section -- 60-second copy-paste fix lowers time-to-resolution for producers hitting the warning.

### Architecture

```mermaid
classDiagram
    direction LR
    class BaseIntegrator {
      <<Abstract>>
      +find_files_by_glob()
      +copy_agent(source, target) int
      +resolve_links()
      +_check_adopt_or_skip()
    }
    class AgentIntegrator {
      +integrate_agents_for_target()
      +find_agent_files()
      +get_target_filename_for_target()
      -_warn_opencode_frontmatter()$ void
      -_FRONTMATTER_RE$ Pattern
    }
    class IntegrationResult {
      <<Dataclass>>
      +files_integrated int
      +files_skipped int
      +target_paths list
      +links_resolved int
    }
    class DiagnosticCollector {
      <<Collect-then-render>>
      +warn(message, package)
      +render_summary()
    }
    class validate_opencode_frontmatter {
      <<Pure Function>>
      +__call__(fm, source, package_name) list~str~
    }
    class OPENCODE_THEME_COLORS {
      <<Constant>>
    }
    BaseIntegrator <|-- AgentIntegrator
    AgentIntegrator ..> validate_opencode_frontmatter : calls
    AgentIntegrator ..> DiagnosticCollector : pushes warnings
    AgentIntegrator ..> IntegrationResult : returns
    validate_opencode_frontmatter ..> OPENCODE_THEME_COLORS : reads
    class AgentIntegrator:::touched
    class validate_opencode_frontmatter:::touched
```

```mermaid
flowchart TD
    A["apm install -t opencode"] --> B["integrate_agents_for_target()\nsrc/apm_cli/integration/agent_integrator.py"]
    B --> C{"mapping.format_id?"}
    C -->|codex_agent| D["_write_codex_agent()"]
    C -->|windsurf_agent_skill| E["_write_windsurf_agent_skill()"]
    C -->|opencode_agent| F["_warn_opencode_frontmatter()\n[I/O] read source file"]
    C -->|other| G["copy_agent()\n[FS] write target"]
    F --> H{"diagnostics is None?"}
    H -->|yes| G
    H -->|no| I{"source.is_symlink()?"}
    I -->|yes| G
    I -->|no| J["[I/O] source.read_text()"]
    J --> K{"_FRONTMATTER_RE match?"}
    K -->|no| G
    K -->|yes| L["yaml.safe_load()"]
    L --> M["validate_opencode_frontmatter()\nsrc/apm_cli/integration/opencode_frontmatter.py"]
    M --> N{"warnings returned?"}
    N -->|yes| O["diagnostics.warn() per message"]
    N -->|no| G
    O --> G
```

### Recommendation

Three shepherd iterations have resolved all blocking and recommended findings. The sole remaining recommended-tier item (doc-writer tools example) is factually inaccurate on current HEAD. CI is all green, 34 tests pass covering every critical promise at integration tier, and the architecture is clean. The three followups above are polish -- they can land in a separate PR without holding this one. Ship.

---

<details>

#### Python Architect

- **[nit]** validate_opencode_frontmatter import is eager; could be lazy inside the opencode_agent branch at `src/apm_cli/integration/agent_integrator.py:17`
  The module is tiny (no heavy deps), cost is negligible today. If Phase 2 adds heavier deps, consider moving the import inside _warn_opencode_frontmatter. Not blocking.
  *Suggested:* No action needed now; note for Phase 2 if the module grows dependencies.

#### CLI Logging Expert

- **[nit]** sorted() list repr in color warning produces Python-flavored brackets in user-facing output at `src/apm_cli/integration/opencode_frontmatter.py:116`
  sorted(OPENCODE_THEME_COLORS) renders as "['accent', 'error', 'info', ...]" -- square brackets and single quotes are Python repr artifacts. Other messages hand-write the list or use a join.
  *Suggested:* Consider ', '.join(sorted(OPENCODE_THEME_COLORS)) to render without Python list syntax.

#### DevX UX Expert

- **[nit]** Warning messages are long single lines that may wrap awkwardly in narrow terminals at `src/apm_cli/integration/opencode_frontmatter.py`
  Each warning is ~180+ chars. 'Fix:' remediation merges visually with problem statement in 80-col terminals. Phase 1 scope -- renderer is shared infrastructure.
  *Suggested:* Consider splitting 'Fix:' onto a second indented line in the warning renderer (not this PR scope).

- **[nit]** sorted(OPENCODE_THEME_COLORS) in color warning produces Python list repr with brackets at `src/apm_cli/integration/opencode_frontmatter.py`
  The warning renders 'or one of ["accent", ...]'. A joined string would read more naturally.
  *Suggested:* Use ', '.join(sorted(OPENCODE_THEME_COLORS)) for cleaner terminal read.

#### Supply Chain Security

- **[nit]** type(tools).__name__ is not passed through _ascii_safe_name, relying on yaml.safe_load only producing ASCII type names at `src/apm_cli/integration/opencode_frontmatter.py`
  yaml.safe_load restricts types to Python builtins with ASCII __name__ values. No actual risk today, but a comment noting the implicit safe_load contract would make the safety argument self-documenting.
  *Suggested:* Add inline comment: '# safe_load guarantees basic Python types with ASCII __name__'

#### OSS Growth Hacker

- **[nit]** CHANGELOG entry leads with implementation detail before user benefit at `CHANGELOG.md`
  Entry opens with 'Zod schema rejects' before the user benefit. Inverting would improve shareability.

- **[nit]** 'contributions welcome' link could be more actionable with a 'help-wanted' label note at `packages/apm-guide/.apm/skills/apm-usage/package-authoring.md`
  Contributors scanning for entry points filter by label. Ensuring #581 has 'help wanted' label would lower friction.

- **[nit]** Pitfall section in docs could add a 60-second copy-pasteable before/after YAML snippet at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  Currently declarative; a quick before/after would let readers fix their agent in under 60 seconds.

#### Auth Expert -- inactive

No auth-related files in diff; auth surface unchanged. PR touches only frontmatter validation (opencode_frontmatter.py, agent_integrator.py hook, docs, CHANGELOG) with no changes to auth.py, token_manager.py, or any credential/host resolution path.

#### Doc Writer

- **[recommended]** Canonical tools: example switched to OpenCode mapping form but prose still says it reaches Claude/Copilot/Cursor 'verbatim' at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  CEO investigation: this finding does NOT hold on current HEAD eb0bb0d1 -- the list form remains in the example and prose is consistent. No action required.
  *Suggested:* Verified as no-issue on current HEAD; downgraded.

- **[nit]** OpenCode theme color enum listed three times on the same page -- drift risk at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  The seven theme names appear in the color table row, in the 'OpenCode is the strictest' paragraph, and in the pitfall bullet. State the full list once; have other mentions point to it.

- **[nit]** No cross-link from new OpenCode paragraph to deeper package-authoring.md section at `docs/src/content/docs/producer/author-primitives/instructions-and-agents.md`
  package-authoring.md now has the canonical 'OpenCode target: frontmatter constraints' section. instructions-and-agents.md OpenCode paragraph does not link to it.
  *Suggested:* Add: 'See also: package-authoring guide -- OpenCode target: frontmatter constraints'

#### Test Coverage Expert

- **[nit]** TestOpencodeInstallEmitsWarnings lives under tests/unit/ but exercises integration-with-fixtures tier behavior at `tests/unit/integration/test_opencode_frontmatter_validation.py`
  The class creates real tempdir fixtures, calls integrate_agents_for_target end-to-end. Acceptable per current repo convention (tests/unit/integration/ is the home for lightweight integration tests). Informational only.
  *Proof (passed):* `tests/unit/integration/test_opencode_frontmatter_validation.py::TestOpencodeInstallEmitsWarnings::test_file_is_still_deployed_when_warning_emitted` -- proves: Install deploys the agent file even when frontmatter triggers a warning -- warnings never block install [devx]
  `assert result.files_integrated == 1; deployed = self.project_root / '.opencode' / 'agents' / 'demo.md'; assert deployed.exists()`

</details>

Issue #757, opened 2026-04-19
[Feature] Document marketplace.json schema 


rcollette:
## Problem
There is no dedicated documentation in the APM project that clearly defines the schema for `marketplace.json`. As the APM tool expands its marketplace model—including support for both Git-backed and URL-backed marketplaces—having an up-to-date, authoritative schema reference is critical for:
- Marketplace publishers (defining their own `marketplace.json`)
- Third-party integrators (validating and generating compatible indexes)
- Users diagnosing issues with marketplace registration or skill indices

## Proposal
- Add a dedicated documentation page or appendix to the current docs that describes the full JSON schema expected for `marketplace.json`.
- Cover all current required and optional fields, types, value constraints, and examples.
- Document how differences are handled between repo-backed and URL-backed marketplace schemas (including references to Agent Skills specification if applicable).
- Include versioning strategy and backward compatibility notes.
- Link to any relevant RFCs/specifications (e.g., Agent Skills, RFC 8615).

## Motivation
- Reduce publisher/integrator confusion and support requests
- Enable better validation, error messages, and automation for marketplace publishing
- Support the planned generalization of the marketplace discovery model

## Example Content to Include
- Example minimal and full-featured `marketplace.json` files
- JSON schema definition (or link)
- Common mistakes and troubleshooting
- Comparison table/references for `.well-known/agent-skills/index.json` vs. repo `marketplace.json`

## Reference material
The guide at https://github.com/microsoft/apm/blob/main/docs/src/content/docs/guides/marketplaces.md is insufficient for agents and other tooling to generate marketplace.json files.

danielmeppiel:
Hi @rcollette , the reference for plugin marketplaces is not an industry standard as of today. The implementation in APM follows the reference schema from [GitHub Copilot CLI](https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-plugin-reference#marketplacejson) and [Claude Code](https://code.claude.com/docs/en/plugin-marketplaces#marketplace-schema), which should both match. Agree this should be clearly stated in the docs.

mizoz:
Hi! I'd love to contribute. Could you assign this issue to me?

danielmeppiel:
> Hi! I'd love to contribute. Could you assign this issue to me?

Feel free to open a PR, reading CONTRIBUTING.md

danielmeppiel:
[...]

PR #1564, opened 2026-06-01
docs: add marketplace.json schema reference

kaizeenn:
Adds a reference page documenting the Claude/Copilot-compatible marketplace.json schema and links it from the reference index. Addresses #757.


danielmeppiel:
I agree we should clearly outline which formats we support, but then I'd lean to directly sending users to the vendors' spec docs.

PR #1574, opened 2026-06-01
feat: make apm update a superset of apm deps update

### LINKED ISSUE #1525: Unify update commands: extend `apm update` with `-g`, deprecate `apm deps update`
## Problem

Two commands refresh dependencies, with divergent surfaces and no superset:

| Surface | `apm update` (top-level, #1244) | `apm deps update` (legacy) |
|---|---|---|
| Project scope (cwd) | ✅ | ✅ |
| User scope (`~/.apm/`, `-g`) | ❌ | ✅ |
| Interactive plan + confirmation | ✅ | ❌ |
| `--dry-run` | ✅ | ❌ |
| `--yes` (CI) | ✅ | ❌ |
| Per-package targeting (`PACKAGES...` arg) | ❌ (refreshes all) | ✅ |
| `--force` | ❌ | ✅ |
| `--parallel-downloads` | ❌ | ✅ |

Neither command is a strict superset, so users must pick the right one based on which feature they need — and the global-update path is buried under `deps` despite `apm install`'s top-level `-g` setting the expectation that scope is a flag, not a subcommand namespace.

## Repro: the UX cliff today

```bash
# Discoverable, modern UX — but cwd-only, no -g
apm update --dry-run

# Hidden under `deps`, no plan/confirmation/dry-run — but the only path to global
apm deps update -g

# Tried by users (myself included) and silently rejected
apm update -g                  # Error: No such option: -g
cd ~/.apm && apm update        # works, but requires --target because no harness markers in ~/.apm
```

## Why one `update` is the right shape

Every neighbouring tool collapses this to a single verb with a scope flag, not a subcommand:

- `npm update -g <pkg>`
- `pip install --upgrade --user <pkg>` / `pipx upgrade <pkg>`
- `cargo install --force` (global by definition)
- `brew upgrade <pkg>` (scope is implicit / per-formula)

`apm install` already follows this pattern: `-g` is a flag on the top-level command. `apm update` should match.

The `deps` subcommand group is the right home for *read-side* introspection (`list`, `tree`, `info`, `clean`) where namespacing makes the command surface easier to scan. `update` is a primary lifecycle verb (peer to `install`, `uninstall`, `prune`) and belongs at the top level — which is exactly the move #1244 made for the project-scope case.

## Proposal

1. **Extend top-level `apm update`** with the missing flags from `apm deps update`:
   - `-g, --global` (the main motivator)
   - `[PACKAGES]...` positional for per-package refresh
   - `--force` for collision overwrite
   - `--parallel-downloads`
2. **Soft-deprecate `apm deps update`** for one release: keep it working, print a one-line banner pointing to `apm update` (mirroring how #1244 deprecated the old self-updater verb).
3. **Remove `apm deps update`** in the next breaking release.
4. **No changes** to `apm deps list / tree / info / clean` — those are read-side and belong in the subcommand group.

## Out of scope

- `apm update` from `~/.apm/` currently requires `--target` because the dir has no harness markers. Once `-g` lives on `apm update`, the global-scope branch should auto-resolve the harness from `~/.apm/apm.yml`'s `targets:` field (or default to `copilot,claude` since those are the global-capable runtimes per `install -g`'s own logic). Can be a follow-up.

## Related

- #1244 — introduced top-level `apm update` and the deprecation pattern this proposal follows.
- #1203 — the lockfile-UX gap that motivated #1244; the global-scope cliff is the same gap, one level up.

Aaryan-Dadu:
## Description

`apm update` now accepts `-g/--global`, positional `[PACKAGES]...`, `--force`, and `--parallel-downloads`, making it a strict superset of `apm deps update`.
One verb now handles project and user scope, single-package refreshes, and collision overwrites - all behind the plan + confirmation + `--dry-run` gate `apm update` already had. `apm deps update` is soft-deprecated (still works, prints a banner, to be removed in the next breaking release).

Closes #1525 

#### Before

Screenshot From 2026-06-01 14-56-42

#### After

Screenshot From 2026-06-01 14-51-11

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass
image

If this PR changes behaviour that an OpenAPM v0.1 `req-XXX` covers,
confirm the three-step ritual (see CONTRIBUTING.md "Adding or
changing a normative requirement"):

- [ ] Spec edit: `docs/src/content/docs/specs/openapm-v0.1.md` updated
      (new/changed `<a id="req-XXX"></a>` anchor + prose + Appendix C
      row).
- [ ] Manifest edit: `docs/src/content/docs/specs/manifests/openapm-v0.1.requirements.yml`
      updated.
- [ ] Test edit: a `@pytest.mark.req("req-XXX")` test under
      `tests/spec_conformance/` added or extended.
- [ ] `CONFORMANCE.{md,json}` regenerated via
      `uv run --extra dev python -m tests.spec_conformance.gen_statement`
      and committed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   one-directional (APM only reads vendor SARIF), and are install-method
   neutral -- they work with the self-contained APM binary with no `pip`
   extra to install.
+- `apm update` now accepts `-g/--global`, positional `[PACKAGES]...`, `--force`, and `--parallel-downloads`, making it a strict superset of `apm deps update`. A single verb now covers project and user scope, per-package refresh, and collision overwrite -- all behind the same interactive plan with `--dry-run`/`--yes`. (closes #1525)
+
+### Deprecated
+
+- `apm deps update` is deprecated in favor of `apm update`, which now exposes every flag it had. It prints a one-line banner pointing to `apm update` and keeps working for one release; it will be removed in the next breaking release. (closes #1525)
+
 ### Fixed
 
 - `apm pack --check-clean` now emits a copy-pasteable recovery recipe when `marketplace.json` drifts from source: `git commit --amend --no-edit` + `git push --force-with-lease` to fold the diff into the current commit, or a follow-up commit variant. Producers get the right command at the point of failure without consulting external docs. (closes #1381)
--- a/docs/src/content/docs/consumer/update-and-refresh.mdx
+++ b/docs/src/content/docs/consumer/update-and-refresh.mdx
@@ -26,15 +26,22 @@ apm update
 
 Re-resolves every dependency in `apm.yml` to its latest matching Git
 reference, prints the planned changes, and prompts for consent before
-rewriting `apm.lock.yaml` and redeploying primitives.
+rewriting `apm.lock.yaml` and redeploying primitives. Pass one or more
+package names (`apm update org/pkg-a org/pkg-b`) to refresh only those.
 
 Useful flags:
 
 - `--dry-run` -- show the resolution plan and exit; never writes.
 - `--yes` -- skip the consent prompt (for scripts and CI).
 - `-v, --verbose` -- show per-dependency resolution detail.
+- `-g, --global` -- refresh user-scope dependencies under `~/.apm/`
+  instead of the current project (pairs with `apm outdated -g`).
+- `--force` -- overwrite locally-authored files on collision.
+- `--parallel-downloads N` -- cap concurrent downloads (default 4;
+  `0` disables parallelism).
 
-This is the command that actually changes versions in your project.
+This is the command that actually changes versions in your project. It
+is a strict superset of the deprecated `apm deps update`.
 
 ## Inspect first
 
--- a/docs/src/content/docs/reference/cli/deps.md
+++ b/docs/src/content/docs/reference/cli/deps.md
@@ -89,6 +89,10 @@ Exit codes: `0` on success, `1` when the package is not installed or the query m
 
 ### `apm deps update`
 
+:::caution[Deprecated]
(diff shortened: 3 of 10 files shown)


danielmeppiel:
## Shepherd advisory: apm-review-panel results

HEAD: efd32f27 (rebased onto main, folds applied, CI pending)

### Panel verdict: ship

Eight panel personas reviewed this PR (python-architect, cli-logging-expert, devx-ux-expert, supply-chain-security-expert, test-coverage-expert, doc-writer, oss-growth-hacker, apm-ceo). CEO arbiter returned **ship**.

### Applied (folded into PR commits)

- **PA-2** -- type annotations on `resolve_requested_packages()`: signature now `(packages: tuple[str, ...] | list[str], all_deps: list) -> list[str] | None`
- **TC-1** -- alias resolution test: `test_alias_maps_to_canonical` added to `TestResolveRequestedPackages`

### Advisory (deferred -- no action required for merge)

- **PA-1** -- collision detection: `resolve_requested_packages()` silently keeps first match when both short-name and alias collide on the same dep. Low priority; no user-visible bug.
- **PA-3** -- return type of `_run_dep_update` could be annotated.
- **CL-1/CL-2** -- deprecation banner: consider collapsing double-warning on repeated invocations (not a blocker).
- **DX-2** -- migration one-liner in banner: consider adding `apm update` as the suggested replacement in the banner text.
- **DX-3/DX-4** -- edge cases: `--target` collision with `-g` and global-mode error messaging.
- **SC-1/SC-2** -- alias-shadowing guard and `--force` scope confirmation for production use.
- **DW-3** -- packages/apm-guide/commands.md: `--legacy-skill-paths` not yet documented.
- **OG-1..5** -- growth/changelog/narrative items; non-blocking.

### Notes

- CL-3 ("remove redundant symbol= kwargs"): panel finding is incorrect -- the `symbol=` args on `_rich_warning` / `_rich_error` ARE required to prepend the `[!]`/`[x]` bracket prefix. Not removed.
- All lint checks pass (ruff clean, pylint 10.00/10, auth-signals clean).
- 62 unit tests pass.

Issue #1571, opened 2026-06-01
[BUG] or [QUESTION] Local dependencies in shared repository


tomdaniel-ae:
**Describe the bug**
We have one repository for our organisation containing all apm packages:
```
apm/
    _shared/
        .apm/
        apm.yml
    frontend/
        .apm/
        apm.yml # REFERENCES _shared
    backend/
        .apm/
        apm.yml # REFERENCES _shared
```

When installing frontend/ or backend/ into one of our projects, we get an error because it's referencing a package in the same remote repository:

```
Refusing to install local_path dependency './_shared' declared by remote package 'development-backend-dotnet': remote packages cannot reference paths on the consumer filesystem. Publish the dependency as a standalone package and reference it via owner/repo or marketplace handle.
```

**To Reproduce**
1. Create two .apm folders (e.g. _shared/ and frontend/)
2. Make sure one apm.yml references the other:
```yml
depdendencies:
  apm:
    - path: ./_shared
```
3. Install the frontend/ folder with `apm install dev.azure.com/org/project/repo/apm/frontend`
4. See the error in the terminal (mentioned above)

**Expected behavior**
I would expect the _shared/ apm package to be installed as well from that same remote repository.

**Environment (please complete the following information):**
 - OS: Windows
 - Python Version: 3.14.3
 - APM Version: 0.16.0
 - GIT Platform: Azure DevOps

**Additional context**
Is this expected behaviour or a bug?  
  
If this is expected behaviour, can there be a section explaining how to achieve a setup like this? Since I can't find a clear explanation for how to set this up (besides using plugins and a complex setup to make this work).

Thanks! :)

ivancea:
++. Discovered this issue while trying to configure a new repo with modules and dependencies.
I had to add absolute git dependencies `git: <org>/<repo>#<branch>`, which makes it very complicated to work with, as now:
- You can't install another branch, unless you replace every "main" branch name too
- You can't use it locally, unless you replace all references with locall ones "temporarily"

danielmeppiel:
## North-star design: same-repo sibling dependencies (monorepo `path:` support)

Thanks `@tomdaniel-ae` -- this is a real, well-formed monorepo pattern, and the
diagnosis in triage is correct. Posting the autopilot **design panel** north-star
so implementation can proceed without re-opening the #940 path-confusion hole.

### (a) Problem framing -- the #940 tension

[...]

PR #1578, opened 2026-06-02
Extract shepherd-driver engine, rewire skill packages, and harden local-dep drift replay

### LINKED ISSUE #857: Local relative paths in transitive deps resolve against root consumer instead of declaring package
## Summary

Relative paths in a non-root package's `apm.yml` are resolved against the **root consumer's** project directory rather than the package's **own source directory**. This makes it impossible to compose two local sibling packages where one declares the other as a dependency — the natural pattern for layering reusable building blocks inside a single repo.

The fix is small (~30 LOC), aligns APM with the semantics every other package manager (npm, pip, cargo, go modules, Bicep) uses for relative paths in manifests, and unlocks clean local-package composition without changing the common case.

## Repro

A monorepo with two local packages — a base and a specialization that builds on it:

```
my-project/
├── apm.yml                                  # consumer
└── packages/
    ├── editorial-pipeline/
    │   └── apm.yml                          # base package
    └── handbook-agents/
        └── apm.yml                          # specialization, depends on the base
```

Consumer apm.yml:
```yaml
dependencies:
  apm:
    - ./packages/handbook-agents
```

`packages/handbook-agents/apm.yml` — the natural way to declare the dep:
```yaml
dependencies:
  apm:
    - ../editorial-pipeline
```

`apm install` fails: APM tries to resolve `../editorial-pipeline` against the **root consumer** (`my-project/`), looking for `my-project/../editorial-pipeline` — i.e. one directory above the project root, which doesn't exist.

The only workaround that "works" today is to write `./packages/editorial-pipeline` inside `handbook-agents/apm.yml` — but that bakes the consumer's filesystem layout into the inner package. Publish that package standalone tomorrow and it's broken for everyone else.

## Root cause

In `src/apm_cli/install/phases/resolve.py` (around line 117), the root project's path is closure-captured once and reused for **every** dep, including transitive ones:

```python
project_root = ctx.project_root  # captured at the root

def download_callback(dep_ref, modules_dir, parent_chain=""):
    ...
    if dep_ref.is_local and dep_ref.local_path:
        result_path = _copy_local_package(
            dep_ref, install_path, project_root, ...   # always root, never the parent's location
        )
```

`_copy_local_package` (`src/apm_cli/install/phases/local_content.py:92-94`) does the resolution:

```python
local = Path(dep_ref.local_path).expanduser()
if not local.is_absolute():
    local = (project_root / local).resolve()
```

So when a transitive dep declares `../editorial-pipeline`, the `project_root` here is the **root consumer**, not the directory whose `apm.yml` actually contains the line. The breadcrumb already exists (`parent_chain` is threaded for error messages), but the parent **package object** isn't passed down, so the call site has no way to ask "where did the apm.yml that declared this dep actually live?"

## Proposed fix

Make relative paths resolve against the **source directory of the package whose apm.yml declared the dep**.

Concretely:

1. **`APMPackage` gains one field**: `source_path: Path` — absolute on-disk directory holding this package's `apm.yml`.
   - Root project: `project_root`
   - Local dep: the **original** path (before being copied into `apm_modules/_local/`), which the resolver already computes
   - Remote dep: the clone directory inside `apm_modules/`

2. **`_try_load_dependency_package`** (`apm_resolver.py:360-438`) sets `package.source_path` when constructing `APMPackage`. For local deps the value is the same `(project_root / local_path).resolve()` expression `_copy_local_package` uses today.

3. **`download_callback` accepts `parent_pkg: Optional[APMPackage]`**. The resolver already knows the parent at every step (it must, to record graph edges); pass it down.

4. **`_copy_local_package`** signature changes from `(dep_ref, install_path, project_root, ...)` → `(dep_ref, install_path, base_dir, ...)` where the caller computes `base_dir = parent_pkg.source_path if parent_pkg else project_root`. The body is unchanged.

That's the entire change. ~30 lines touched, no new abstractions.

## Why this is correct, not just convenient

| Behaviour | Today | With fix |
|---|---|---|
| Root: `./packages/foo` | resolves vs `project_root` ✅ | resolves vs `project_root` ✅ (root's `source_path == project_root`) |
| Local-in-local: `../sibling` | breaks | resolves vs the parent package's source dir ✅ |
| Remote-in-remote: `./internal-helper` | meaningless | resolves vs the clone dir ✅ |
| Absolute paths | works | works (handled by the absolute branch already) |

This semantic — "relative paths in a manifest are relative to that manifest's directory" — is what every other package manager uses (npm `file:`, pip editable installs, cargo path deps, go module replace directives, Bicep module paths). The current behaviour is the surprising one.

## Wins

- **Real local layering.** A repo can ship reusable building blocks (`packages/base/`) and consumers (`packages/specialization/`) that genuinely declare the relationship. Today this requires either flattening at the consumer root (so the dep is invisible to APM) or publishing the base as its own remote repo just to express a sibling-folder dependency.
- **Portability survives extraction.** A local package can be lifted into its own repo later without rewriting its `apm.yml`. The relative path keeps the same meaning.
- **Aligns with developer intuition.** No one reading `../editorial-pipeline` in `packages/handbook-agents/apm.yml` expects it to resolve from somewhere two directories up.
- **Consistent semantic across local and remote deps.** Both anchor on "the directory containing the apm.yml that declared me." Today, transitive local deps anchor on the root consumer, which has no equivalent in the remote case.
- **Doesn't change the common case.** Root-level relative paths behave identically (root's `source_path == project_root`).

## Risks & edge cases

| Concern | Severity | Notes |
|---|---|---|
| **Backward compatibility** | Low | The only behaviour that changes is local-relative paths in **non-root** apm.yml files. That case is broken today (resolves to nonsense), so realistically no one depends on the current resolution. The root case is byte-identical. |
| **Install-path collisions** | Pre-existing | `install_path = apm_modules/_local/<dirname>` (`reference.py:335`). Two parents declaring local deps with the same final dirname collide today and still would. Orthogonal to this issue; addressable later by hashing the source path or namespacing by parent. |
| **Lockfile entries** | Low | `lockfile.py:198-215` serialises `local_path` as a string. For transitive local deps the string's anchor would change (relative-to-source rather than relative-to-root); a regenerate handles it. Direct local deps are unaffected. |
| **Cycle detection** | None | Graph walker already detects cycles by package identity, not by path. Unchanged. |

## Out of scope

- Fixing the install-path collision risk (separate, longer-standing)
- Subpath-remote refs (`org/repo/sub/path`) — already work today, this issue is purely about the local-path branch
- Any change to lockfile schema

---

Happy to send a PR if there's appetite. Analysis was done against `awd-cli` `main` (current).

danielmeppiel:
# refactor(skills): extract shepherd-driver, add issue-autopilot, harden local-dep drift replay

## TL;DR

This PR does three things. It extracts the per-PR convergence engine out of `batch-bug-shepherd` into a standalone, reusable **`shepherd-driver`** package (true DRY); it adds a new **`apm-issue-autopilot`** orchestrator skill that takes a batch of issues from triage to a mergeable PR queue, reusing `shepherd-driver` for the PR loop; and it **fixes a silent, permanent disabling of drift detection** for any repo that has a transitive local dependency. The first two are agentic-primitive (markdown) changes; the third is a code fix in `src/apm_cli/deps`.

> [!NOTE]
> The drift fix (closes #857) is also the resolution to the design question in #1571: a package's `../sibling` local dep now resolves against the parent package's resolved dir, mirroring install, so external repos and the drift gate agree on where a transitive local source lives.

## Problem (WHY)

- [x] **Drift detection was silently, permanently disabled** for any repo with a transitive local dep. Drift replay anchored every local dep's path on `project_root`; a child package declaring `../sibling` resolved to `project_root/../sibling`, escaped the repo, raised `CacheMissError`, and the `ci_checks` gate soft-skipped the **entire** drift check as "cache not populated" -- including this very repo.
- [x] **The per-PR shepherding engine lived inside `batch-bug-shepherd`**, so a second consumer (issue autopilot) could not reuse the classify -> fold -> push -> CI-watch -> mergeability loop without copy-paste. That is a DRY violation across skill packages.
- [!] **No automated path from a raw issue to a mergeable PR.** Triage, solution design, planning, implementation, and PR convergence existed only as separate manual motions.

Why these matter: a soft-skip that hides a disabled safety gate is the opposite of a deterministic gate -- "Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action." The DRY extraction follows "Favor small, chainable primitives over monolithic frameworks." (both [PROSE](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/)).

## Approach (WHAT)

| # | Pillar (and why) |
|---|------------------|
| 1 | **Extract `shepherd-driver`** -- the per-PR engine becomes its own package; `batch-bug-shepherd` and `apm-issue-autopilot` both consume it via an `apm` dep. |
| 2 | **Add `apm-issue-autopilot`** -- an orchestrator skill: triage gate (escalate-by-default) -> Ideate/Plan/Implement wave pipeline -> `shepherd-driver` PR convergence, with an A12 cost gradient bound per spawn in `model-routing.md`. |
| 3 | **Harden drift replay** -- new `deps/path_anchoring.py` walks the lockfile `resolved_by` chain so `../sibling` anchors on the parent package's resolved dir; a corrupt graph hard-fails the gate instead of soft-skipping. |

Also included (listed for completeness; not the narrative focus of this PR): new `docs-grounding-verifier`, `docs-corpus-audit`, and `apm-spec-guardian` skill bundles, plus the `spec-*` and `performance-expert` agent personas. These are additive primitive files validated by `apm audit` + their own evals.

## Implementation (HOW)

**Pillar 1 + 2 -- skill packages (agentic primitives):**

- **`packages/shepherd-driver/`** (new) -- the shared per-PR engine: `shepherd-driver-prompt.md`, `conflict-resolution-prompt.md`, `references/mergeability-gate.md`, `assets/completion-schema.json`. Depends on `apm-review-panel`.
- **`packages/batch-bug-shepherd/`** -- rewired to delegate the per-PR loop to `../shepherd-driver/`; keeps its own triage (`triage-prompt.md`, `strategic-alignment-prompt.md`). Removed assets are moved, not deleted; all references repoint to the sibling package.
- **`packages/apm-issue-autopilot/`** (new) -- orchestrator `SKILL.md` plus pipeline asset prompts (`solution-pipeline-prompt.md`, `ideate-prompt.md`, `plan-panel-prompt.md`) and the `model-routing.md` B12 SKU map. Depends on `../shepherd-driver` and `../apm-triage-panel`.

**Pillar 3 -- drift replay (code):**

- **`src/apm_cli/deps/path_anchoring.py`** (new) -- `resolve_local_dep_dir` walks the `resolved_by` chain, anchoring relative local paths on the parent's resolved dir (absolute paths bypass anchoring); raises new `LocalResolutionError` on a missing/ambiguous/non-local/cyclic parent.
- **`src/apm_cli/install/drift.py`** -- `_materialize_install_path` delegates to the helper (keyword-only `lockfile` param); `run_replay` threads `lockfile=lock`. No `project_root` clamp -- install deliberately permits out-of-tree local sources, so clamping drift would falsely flag deps that install happily.
- **`src/apm_cli/policy/ci_checks.py`** -- `LocalResolutionError` -> `passed=False` (hard fail) before the `CacheMissError` soft-skip; a correctly-resolved-but-missing dir still soft-skips, unchanged.
- **`src/apm_cli/deps/lockfile.py`**, **`models/dependency/reference.py`** -- document/tag local-path anchoring semantics (`source='local'` at parse time).

## Diagrams

Legend: package-ownership and dependency edges -- which of the five packages owns each component, and the `apm` dep arrows between them. Note that the solution pipeline is a set of asset prompts *inside* `apm-issue-autopilot`, not a separate package, and that triage's rubric lives in the external `apm-triage-panel` package.

```mermaid
flowchart TB
    subgraph AIA["PACKAGE apm-issue-autopilot"]
        ORCHp["SKILL.md orchestrator"]
        A1["asset triage-prompt.md"]
        A2["assets solution-pipeline + ideate + plan-panel prompts"]
        A3["asset model-routing.md B12"]
    end
    subgraph SDP["PACKAGE shepherd-driver (shared per-PR engine)"]
        SD1["shepherd-driver-prompt + conflict-resolution + mergeability-gate"]
    end
    subgraph TPP["PACKAGE apm-triage-panel"]
        T1["triage rubric"]
    end
    subgraph RPP["PACKAGE apm-review-panel"]
        R1["review engine"]
    end
    subgraph BBS["PACKAGE batch-bug-shepherd (sibling consumer)"]
        B1["own triage + fix prompts"]
    end

    AIA -->|"apm dep"| SDP
    AIA -->|"apm dep"| TPP
    BBS -->|"apm dep"| SDP
    SDP -->|"apm dep"| RPP

    A1 -.->|"wraps rubric"| TPP
    A2 -.->|"opens PR, then drives"| SDP

    classDef pkg fill:#0e7490,color:#fff,stroke:#083d4d
    class AIA,SDP,TPP,RPP,BBS pkg
```

Legend: the autopilot execution flow with each step's cost properties -- model SKU, brief compression mode (NORMAL vs CAVEMAN_FULL), and spawn type (orchestrator-inline vs fan-out subagent). The purple front gates (triage, Ideate, Plan) are deliberately bound to opus 4.8 because a wrong call there poisons the whole downstream pipeline (A12 buy-quality-where-stakes-are-highest).

```mermaid
flowchart TB
    legend["LEGEND each step shows: model - briefMode - spawnType. ORCH=run by orchestrator inline. FANOUT=spawns subagents"]

    subgraph OWN_AIA["owned by package apm-issue-autopilot"]
        ORCH["Orchestrator - n/a - NORMAL - ORCH inline, sole writer to issue+PR"]
        G1["P1 Triage - opus 4.8 - NORMAL brief - FANOUT 1 per issue"]
        PIPE["P4 Pipeline child - sonnet 4.6 - NORMAL - FANOUT 1 per issue, sole writer of branch"]
        ID["Ideate - opus 4.8 - NORMAL - subagent of pipeline"]
        PL["Plan Architect - opus 4.8 - NORMAL - subagent of pipeline"]
        LN["Plan lenses - haiku 4.5 - CAVEMAN_FULL - up to 4 subagents, conditional"]
        IMP["Implement tasks - sonnet 4.6 - NORMAL - up to 6 subagents per wave"]
        VG["Wave-gate verifier - haiku 4.5 - CAVEMAN_FULL - 2 subagents per wave"]
        ACC["Acceptance close - pipeline model - INLINE - ORCH inside pipeline, no spawn"]
    end
    subgraph OWN_TP["package apm-triage-panel"]
        RUB["triage rubric, loaded by Triage child"]
    end
    subgraph OWN_SD["package shepherd-driver"]
        DRV["per-PR driver - classify review + fold + push + CI watch - FANOUT 1 per PR"]
    end
    subgraph OWN_RP["package apm-review-panel"]
        PANEL["review panel, loaded by driver each iteration"]
    end

    HUMAN["Maintainer escalation target"]
    DONE["ready-to-merge PR"]

    ORCH --> G1
    G1 -->|"loads rubric"| RUB
    G1 -->|"low confidence / red_flags"| HUMAN
    G1 -->|"accept"| PIPE
    PIPE --> ID --> PL --> IMP --> VG
    PL -.->|"conditional"| LN
    VG -->|"pass"| ACC
    VG -->|"fail, replan <=2"| PL
    ACC -->|"opens PR"| DRV
    DRV -->|"loads panel"| PANEL
    DRV -->|"green + mergeable"| DONE

    classDef heavy fill:#5b21b6,color:#fff,stroke:#3c1773
    classDef mid fill:#1d4ed8,color:#fff,stroke:#13327a
    classDef cheap fill:#047857,color:#fff,stroke:#024c37
    classDef human fill:#b45309,color:#fff,stroke:#7a3806
    classDef ext fill:#374151,color:#fff,stroke:#1f2733
    class G1,ID,PL heavy
    class PIPE,IMP,ACC,ORCH mid
    class LN,VG cheap
    class HUMAN human
    class RUB,DRV,PANEL,DONE ext
```

## Trade-offs

- **Drift fix mirrors install rather than clamping to `project_root`.** Chose to anchor on the parent's resolved dir; rejected a `project_root` clamp because install deliberately permits out-of-tree local sources, so a clamp would make drift falsely flag deps that install accepts.
- **Corrupt graph hard-fails; correctly-resolved-but-missing soft-skips.** A new `LocalResolutionError` (not a `CacheMissError` subclass) fails the gate closed; the legitimate "cache not populated" path is preserved.
- **Front-loaded opus on triage + Ideate + Plan.** Accepts higher per-issue cost at the three gates that determine whether the whole pipeline is correctly scoped; the implementer bulk stays sonnet and read-only verifiers/lenses stay haiku.
- **Scenario evidence is scoped to the code (drift) pillar.** The skill-package pillars are additive primitive/asset files validated by `apm audit` + their own evals, per the scenario-evidence rubric's asset-change skip clause; no runtime behavior in `src/` changes for them.

## Benefits

1. **Drift detection is no longer silently disabled** in any repo with a transitive local dep -- including this one; the gate now hard-fails a corrupt graph instead of passing it.
2. **One shared per-PR engine** instead of two copies: `batch-bug-shepherd` and `apm-issue-autopilot` consume the same `shepherd-driver` package.
3. **A manual issue-to-mergeable-PR path** with triage as the paramount, escalate-by-default front gate.
4. **Per-spawn cost is declared, not inferred** -- `model-routing.md` binds every Phase 4 spawn to a concrete SKU and compression mode.

## Validation

`uv run apm audit --ci --no-drift`:

```
[+]      lockfile-exists          Lockfile present
[+]      ref-consistency          All dependency refs match lockfile
[+]      deployed-files-present   All deployed files present on disk
[+]      no-orphaned-packages     No orphaned packages in lockfile
[+]      skill-subset-consistency Skill subset selections match lockfile
[+]      config-consistency       No MCP configs to check
[+]      content-integrity        No critical hidden Unicode or hash drift detected
[+]      includes-consent         'includes:' declared -- local content deployment is explicitly consented
[*] All 8 check(s) passed
```

<details><summary>Targeted drift-replay tests (48 passed)</summary>

```
$ uv run pytest tests/unit/deps/test_path_anchoring.py tests/integration/test_drift_check.py -q
................................................                         [100%]
48 passed in 8.04s
```

</details>

All PR checks are green on CI (Lint, Build & Test shards, APM Self-Check, Drift Check, Spec conformance gate, Binary Smoke, CodeQL).

### Scenario Evidence

| 1 | "Drift detection stays active in a repo whose package declares a `../sibling` local dep" | Portability / Governed by policy | `tests/integration/test_drift_check.py` (Section F nested `./packages/parent` -> `../sibling` regression trap) | integration |
| 2 | "A `../sibling` path resolves the same way for drift as it does for install" | Portability | `tests/unit/deps/test_path_anchoring.py` | unit |
| 3 | "A corrupt local-dep lockfile graph fails the gate closed, not open" | Secure by default / Governed by policy | `tests/unit/policy/test_ci_checks.py` | unit |

## How to test

- [ ] `git checkout danielmeppiel/bug-queue-shepherd && uv sync` -> branch builds.
- [ ] `uv run pytest tests/unit/deps/test_path_anchoring.py tests/integration/test_drift_check.py -q` -> 48 passed.
- [ ] `uv run apm audit --ci --no-drift` -> "All 8 check(s) passed".
- [ ] Inspect `packages/apm-issue-autopilot/.apm/skills/apm-issue-autopilot/assets/model-routing.md` -> triage + Ideate + Architect bind to `claude-opus-4.8`.
- [ ] `grep -r '../shepherd-driver/' packages/batch-bug-shepherd/.apm/skills/` -> BBS references resolve to the sibling package, no dangling in-package paths.


danielmeppiel:
## Review resolution (commit `aacc2b99`)

Thanks @copilot-pull-request-reviewer. All open threads are addressed below. Folding the comments surfaced a genuine shipped bug, so this turned into a substantive hardening pass rather than cosmetics.

### The `apm.lock.yaml` `local_path: "../foo"` threads (3) -- the important ones

The reviewer is correct that `../foo` is **anchored to the dependency's package dir, not the project root**, and that this can break "lockfile replay/drift logic." It did -- but the suggested remedy (rewrite the paths to be project-root-relative) is the **wrong** fix and would itself ship breakage:

- `../foo` is the **correct, intended** value. The install resolver (`apm_resolver._compute_dep_source_path`) deliberately anchors a transitive local dep on its **parent package's** resolved dir, and `install/phases/local_content.py` documents that out-of-tree local sources (`apm install ../pkg-a`) are a supported, opt-in monorepo layout. Forcing project-root-relative paths would diverge the lockfile from how install actually resolves, breaking the sibling-package layout this repo uses.
- The **real defect** was on the consumer side: `install/drift.py::_materialize_install_path` joined every local `local_path` on `project_root`, so a transitive `../sibling` became `project_root/../sibling`, escaped the repo, raised `CacheMissError`, and the `ci_checks` drift gate **soft-skipped the entire check** as "cache not populated." Net effect: **drift detection was silently, permanently disabled for any repo with a transitive local dep -- including this one.**

Fix in `aacc2b99`: drift now **mirrors install**. New shared helper `deps/path_anchoring.resolve_local_dep_dir` walks the lockfile's `resolved_by` chain and anchors `../sibling` on the parent's resolved dir (recursively, root deps anchor on `project_root`; absolute paths bypass). The misleading `lockfile.py:42` docstring that said "relative to project root" -- the premise of these three comments -- is corrected to document direct-vs-transitive anchoring. A genuinely corrupt graph (missing/ambiguous/non-local/cyclic parent) now hard-fails via a new `LocalResolutionError` instead of masquerading as a cold cache.

Evidence it was real and is fixed: `apm audit --ci` on this repo previously skipped drift; it now reports **"Replayed 6 package(s) -> No drift detected"** with the `drift` check present (9/9). A Section F integration regression-trap (nested `./packages/parent -> ../sibling`) fails on the pre-fix engine and passes now.

### The `confidence-gate-rubric.md` `type/` prefix threads (4) and `ground-truth-table.md` verdict thread (2)

Both fixed in `aacc2b99` (source + deployed `.agents/` twin): the rubric now spells implementation types as `type/bug`, `type/feature`, etc. to match the triage schema, and the seed row's `verdict` cell is blanked instead of carrying the out-of-enum `pending-triage` sentinel (the `pending-triage` state already lives in the dedicated `status` column).

### Portability (relates to #1571)

These transitive local deps are **monorepo-internal only**. `#940` dual-rejects a `local_path` dependency declared by a **remote** parent, so an external repo depending on one of these published packages cannot resolve or download their `../sibling` local edges -- consistent with the local-dependency portability questions raised in #1571. This PR does not change that boundary; it only makes the **in-repo** drift audit honest about the local graph it already installs.

Leaving the individual threads open for maintainer resolution.

PR #1588, opened 2026-06-02
fix: send auth header for GitHub release lookup in installer and self-update

### LINKED ISSUE #1582: Installer and self-update ignore token env vars for GitHub release lookup
## Summary

I cannot reliably install or update APM while connected through my company VPN, because APM does the GitHub `releases/latest` lookup anonymously even when I provide `GITHUB_TOKEN` or `GITHUB_APM_PAT`.

That means APM can fail on GitHub's shared anonymous/IP-based quota even though `gh auth status` is healthy and my authenticated GitHub API quota is not exhausted.

Observed with APM CLI `0.14.0`.

## Symptom

With the anonymous GitHub API quota exhausted:

```sh
export GITHUB_TOKEN="$(gh auth token)"
curl -sSL https://aka.ms/apm-unix | sh
```

can fail with:

```text
Error: Invalid API response received
GitHub API Error:
API rate limit exceeded for .
```

The authenticated quota can still be healthy at the same time:

```sh
gh api rate_limit --jq '.resources.core'
```

## What seems to be happening

In `install.sh`, the token env vars are resolved before release lookup, but APM still makes the first `releases/latest` request without auth:

- `install.sh` resolves `GITHUB_APM_PAT` / `GITHUB_TOKEN` into `AUTH_HEADER_VALUE`.
- The first `https://api.github.com/repos/$APM_REPO/releases/latest` call is made without that header.
- The script only retries with auth for curl failure, empty response, or `"Not Found"`.
- A GitHub rate-limit response is valid JSON but has no `tag_name`, so the installer exits as `Invalid API response received`.

`apm self-update` appears to have the same issue one step earlier: it calls `get_latest_version_from_github()` before running the installer, and that helper also requests `releases/latest` without auth headers.

## Expected behavior

If `GITHUB_APM_PAT` or `GITHUB_TOKEN` is set, APM should not start with an anonymous release lookup. It should either:

- use it for the first `releases/latest` request, or
- retry with auth when GitHub returns an anonymous rate-limit response.

## Workaround

Pin the version using authenticated `gh`, which lets the installer skip `releases/latest`:

```sh
VERSION="$(gh api repos/microsoft/apm/releases/latest --jq .tag_name)"
curl -sSL https://aka.ms/apm-unix | VERSION="$VERSION" GITHUB_TOKEN="$(gh auth token)" sh
```

## Related context

I searched existing issues/PRs and did not find this exact case already tracked.

Related but distinct:

- #660 added `VERSION` pinning, which is the current workaround.
- #322 handled unauthenticated GitHub API limits for dependency file downloads, but not installer release metadata.
- #773 mentions GitHub API rate-limit wording in dependency download/error paths, but not the installer or self-update path.

## Small repro note

This form does not pass the token to the installer process:

```sh
GITHUB_TOKEN="$(gh auth token)" curl -sSL https://aka.ms/apm-unix | sh
```

The env assignment only applies to `curl`, not the `sh` process. The repro above uses `export` so the installer receives the token.

danielmeppiel:
## TL;DR

Two anonymous GitHub API calls caused rate-limit failures (403) on shared IPs and corporate NAT. Both now send a conditional `Authorization` header when a token is present, with anonymous fallback preserved.

Closes #1582

## Problem

**install.sh** resolved `AUTH_HEADER_VALUE` (from `GITHUB_APM_PAT` / `GITHUB_TOKEN`) around line 213, but the first `releases/latest` curl call around line 232 ignored it and ran anonymously. Under rate-limiting the first call would fail, triggering the private-repo fallback unnecessarily.

**`src/apm_cli/utils/version_checker.py`** `get_latest_version_from_github` called `requests.get` with no headers; `self_update.py` calls this function, so `apm self-update` also hit the rate limit silently.

## Approach

- **install.sh**: conditional first curl -- send `Authorization: token $AUTH_HEADER_VALUE` when already resolved; anonymous otherwise. Anonymous fallback + private-repo retry path unchanged.
- **version_checker.py**: add `_get_github_token()` following canonical `TOKEN_PRECEDENCE["modules"]` order (`GITHUB_APM_PAT -> GITHUB_TOKEN -> GH_TOKEN`); pass header in `requests.get` only when a token is present.

## Implementation

**install.sh** (around line 231):
```bash
if [ -n "$AUTH_HEADER_VALUE" ]; then
    LATEST_RELEASE=$(curl -s -H "Authorization: token $AUTH_HEADER_VALUE" ...)
else
    LATEST_RELEASE=$(curl -s ...)
fi
```

**version_checker.py**:
```python
def _get_github_token() -> str | None:
    return (
        os.environ.get("GITHUB_APM_PAT")
        or os.environ.get("GITHUB_TOKEN")
        or os.environ.get("GH_TOKEN")
        or None
    )
# ...
token = _get_github_token()
headers = {"Authorization": f"token {token}"} if token else {}
response = requests.get(url, headers=headers, timeout=timeout)
```

## Auth guardrails

- Token precedence: `GITHUB_APM_PAT > GITHUB_TOKEN > GH_TOKEN` (mirrors `TOKEN_PRECEDENCE["modules"]` in `core/token_manager.py`). No new env vars.
- Token value is never logged, echoed, or included in any exception text.
- Anonymous fallback preserved when no token is set (header is omitted, not empty).
- `scripts/lint-auth-signals.sh` passes (no `get_bearer_provider` or unannotated `git ls-remote` violations).

## Validation evidence

```
All checks passed! / 1161 files already formatted

$ uv run --extra dev python -m pylint --disable=all --enable=R0801 --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/

$ bash scripts/lint-auth-signals.sh
[+] auth-signal lint clean

$ bash -n install.sh
(exit 0)

$ uv run --extra dev pytest tests/unit/test_version_checker.py -v
34 passed in 2.44s
```

New tests (`TestGitHubTokenResolution`, `TestGitHubVersionFetchAuth`):
- Assert `Authorization` header is sent only when a token is present
- Assert 403 rate-limit response with no token returns `None` gracefully
- Assert token value does not appear in any log/exception output
- Assert 200 with token returns the version string

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `apm publish` auto-pack now includes `README.md`, `CHANGELOG.md`, and `LICENSE` / `LICENCE` (case-insensitive, symlinks excluded) in the flat registry archive, matching npm's behaviour of bundling standard root-level documentation files alongside the package source.
+- `install.sh` and `apm self-update` now send a conditional `Authorization` header on GitHub release-lookup API calls when `GITHUB_APM_PAT`, `GITHUB_TOKEN`, or `GH_TOKEN` is set, improving reliability for users on shared IPs and corporate NAT that hit anonymous rate limits. Anonymous fallback is preserved when no token is configured. (closes #1582)
 
 ## [0.16.1] - 2026-06-01
 
--- a/install.sh
+++ b/install.sh
@@ -228,8 +228,14 @@ if [ -z "$TAG_NAME" ]; then
 # Get latest release info
 echo -e "${YELLOW}Fetching latest release information...${NC}"
 
-# Try to fetch release info without authentication first (for public repos)
-LATEST_RELEASE=$(curl -s "https://api.github.com/repos/$APM_REPO/releases/latest")
+# Fetch release info; include Authorization header when a token is already resolved
+# (AUTH_HEADER_VALUE set earlier from GITHUB_APM_PAT > GITHUB_TOKEN precedence).
+# This avoids anonymous rate-limiting behind shared IPs / corporate NAT.
+if [ -n "$AUTH_HEADER_VALUE" ]; then
+    LATEST_RELEASE=$(curl -s -H "Authorization: token $AUTH_HEADER_VALUE" "https://api.github.com/repos/$APM_REPO/releases/latest")
+else
+    LATEST_RELEASE=$(curl -s "https://api.github.com/repos/$APM_REPO/releases/latest")
+fi
 CURL_EXIT_CODE=$?
 
 # Check if the response indicates authentication is required (private repo)
--- a/install.sh
+++ b/install.sh
@@ -210,10 +210,13 @@ if [ -e "$APM_INSTALL_DIR" ] && [ ! -w "$APM_INSTALL_DIR" ]; then
 fi
 
 # Resolve auth token (needed for both API and download paths)
+# Precedence: GITHUB_APM_PAT > GITHUB_TOKEN > GH_TOKEN (mirrors version_checker.py)
 if [ -n "$GITHUB_APM_PAT" ]; then
     AUTH_HEADER_VALUE="$GITHUB_APM_PAT"
 elif [ -n "$GITHUB_TOKEN" ]; then
     AUTH_HEADER_VALUE="$GITHUB_TOKEN"
+elif [ -n "$GH_TOKEN" ]; then
+    AUTH_HEADER_VALUE="$GH_TOKEN"
 fi
 
 # When VERSION is provided, skip GitHub API and compute download URL directly
@@ -229,7 +232,7 @@ if [ -z "$TAG_NAME" ]; then
 echo -e "${YELLOW}Fetching latest release information...${NC}"
 
 # Fetch release info; include Authorization header when a token is already resolved
-# (AUTH_HEADER_VALUE set earlier from GITHUB_APM_PAT > GITHUB_TOKEN precedence).
+# (AUTH_HEADER_VALUE set earlier from GITHUB_APM_PAT > GITHUB_TOKEN > GH_TOKEN precedence).
 # This avoids anonymous rate-limiting behind shared IPs / corporate NAT.
 if [ -n "$AUTH_HEADER_VALUE" ]; then
     LATEST_RELEASE=$(curl -s -H "Authorization: token $AUTH_HEADER_VALUE" "https://api.github.com/repos/$APM_REPO/releases/latest")
--- a/src/apm_cli/utils/version_checker.py
+++ b/src/apm_cli/utils/version_checker.py
@@ -1,14 +1,36 @@
 """Version checking and update notification utilities."""
 
+import os
 import re
 import sys
 from pathlib import Path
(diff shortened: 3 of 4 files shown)


danielmeppiel:
## apm-review-panel advisory (shepherd-driver, 1 iteration)

**Verdict: ship_with_followups** -- CI green, auth safe, two items folded, one deferred.

---

### Panel findings

**python-architect**
- [FOLDED] `from typing import Optional, Tuple  # noqa: F401, UP035` removed from `version_checker.py` -- dead code on a touched file; noqa suppressor was masking it.
- [MINOR] Add a sync comment in `_get_github_token()` referencing `GitHubTokenManager.TOKEN_PRECEDENCE["modules"]` so future readers know the order is canonical. Not blocking; defer to a follow-up if desired.

**auth-expert**
- Direct `os.environ.get()` in `version_checker.py` is ACCEPTABLE. `version_checker` is a lightweight non-blocking utility (2 s timeout, fail-gracefully) distinct from `AuthResolver` which is designed for git/dep operations with `(host, org)` context and heavier dependencies. Pattern stays out of scope for `lint-auth-signals.sh` checks.
- [FOLDED] `install.sh` initial auth block now includes `GH_TOKEN` as third fallback, consistent with `version_checker.py` TOKEN_PRECEDENCE order: `GITHUB_APM_PAT > GITHUB_TOKEN > GH_TOKEN`.

**test-coverage-expert**
- COMPLETE. All precedence paths, anonymous fallback, token-leak prevention, and 403-response trap are covered at unit tier in `TestGitHubTokenResolution` and `TestGitHubVersionFetchAuth`. No gaps.

**supply-chain-security-expert**
- Minimal risk introduced. API calls pin to `api.github.com`. Three pre-existing hardening gaps noted (not introduced by this PR):
  - `bash -x` / xtrace exposure of token in `install.sh` (pre-existing)
  - `GITHUB_URL` override without validation (pre-existing)
  - `ASSET_URL` domain not validated against `github.com` (pre-existing)
- [DEFERRED] These are cross-cutting install.sh hardening concerns unrelated to the auth-header fix; tracked for a follow-up security audit.

**cli-logging-expert**
- CLEAN. No token leakage in log output. Authorization header set in-memory and never echoed.

**devx-ux-expert**
- CLEAN. Fully transparent to end users; anonymous fallback preserved; no new flags or prompts required.

**oss-growth-hacker**
- POSITIVE signal. Fixes a recurring pain point for users on shared IPs / corporate NAT hitting anonymous rate limits during install. No docs change needed; fix is self-contained.

---

### Items folded into this PR

| Item | Source | File |
|------|--------|------|
| Remove dead `from typing import Optional, Tuple  # noqa: F401, UP035` | Copilot LEGIT + python-architect | `src/apm_cli/utils/version_checker.py` |
| Add `GH_TOKEN` as third fallback in initial auth block | auth-expert nit | `install.sh` |
| CHANGELOG entry under `[Unreleased] ### Added` | fold-vs-defer rubric (mandatory) | `CHANGELOG.md` |

### Items deferred (follow-up issues recommended)

| Item | Reason |
|------|--------|
| `install.sh` supply-chain hardening (xtrace, GITHUB_URL, ASSET_URL) | Pre-existing patterns not introduced by this PR; cross-cutting security audit out of scope |
| `_get_github_token()` sync comment referencing TOKEN_PRECEDENCE | Cosmetic nit; not blocking |

Issue #548, opened 2026-04-02
[BUG] apm deps update skips transitive deps due to download_callback using locked SHA


sergio-sisternes-epam:
**Describe the bug**
`apm deps update` fails to re-download transitive dependencies even when
their remote SHA has changed. The BFS resolution phase (`download_callback()`
in install.py ~line 1268) downloads transitive deps using the OLD locked SHA
from the lockfile, ignoring `update_refs`. These packages are then marked as
`already_resolved`, causing the sequential download loop to skip them without
any SHA comparison.

**To Reproduce**
1. `apm install owner/package-a` (A depends on B transitively)
2. B's author pushes a new commit to B's main branch
3. `apm deps update` -- B is NOT re-downloaded
4. Only `apm install --force` + deleting `apm_modules/` works

**Expected behavior**
`apm deps update` should detect that B's remote SHA differs from the
lockfile and re-download the updated version.

**Environment (please complete the following information):**
- OS: All (code-level bug, not platform-specific)
- Python Version: 3.12
- APM Version: 0.8.9

**Logs**
N/A -- package is silently skipped.

**Additional context**
Root cause chain:
1. `download_callback()` (line ~1268) uses `locked_dep.resolved_commit` (old SHA)
   regardless of `update_refs` flag
2. Package downloaded with old SHA, stored in `callback_downloaded`
3. Pre-download phase (line ~1596) skips packages in `callback_downloaded`
4. Sequential loop (line ~1880): `already_resolved = True` -> `skip_download = True`
5. No SHA re-check happens for these packages

Fix options:
1. When `update_refs=True`, `download_callback()` should NOT override ref with locked SHA
2. OR: Do not mark packages as `already_resolved` during `--update`
3. OR: Re-check `already_resolved` packages in sequential loop when `update_refs=True`

Discovered during investigation of #515. The content hash fix in #546 addresses
a different scenario (same SHA, different content). This bug affects the more
common case where the SHA has actually changed.

Related: #515, #546

PR #1612, opened 2026-06-02
perf: skip re-download when callback SHA matches remote in --update mode (#551)

### LINKED ISSUE #551: [FEATURE] Optimize --update to skip re-download when callback-fetched SHA matches remote

During `apm deps update`, when `download_callback()` fetches a transitive dependency (e.g., in a clean checkout with empty `apm_modules/`), the sequential loop may re-download the same package because `lockfile_match=False` (stale lockfile SHA). The content is identical both times -- the second download is wasted network I/O.

Identified during Copilot review of #550 (fix for #548).

In the sequential loop, when `already_resolved=True` and `update_refs=True`, compare the SHA captured in `callback_downloaded[dep_key]` against the resolved remote SHA. If they match (and local HEAD matches), treat as up-to-date and skip the redundant download. The lockfile should still be refreshed with the new SHA.

- Accept the double-download as harmless (current behavior after #550). The second download overwrites identical content and only costs one extra network round-trip per transitive dep.
- Compare content hashes before/after callback download to detect no-ops.

**Additional context**

This is a performance optimization, not a correctness issue. The double-download only occurs when:
1. `apm_modules/` is empty or the package directory does not exist
2. The lockfile has a stale SHA (different from current remote HEAD)
3. `update_refs=True` (--update mode)

Related: #548, #550

danielmeppiel:
## TL;DR

In `--update` mode, when a transitive dep is already downloaded by the BFS resolver callback and the remote SHA still matches what was captured, skip the redundant sequential-loop re-download. One fewer network round-trip per affected transitive dep.

## Problem (WHY)

During `apm deps update`, the BFS resolver callback (`download_callback`) fetches transitive deps and records their resolved SHA in `ctx.callback_downloaded[dep_key]`. In the sequential integration loop, `skip_download` was computed as:

```python
skip_download = install_path.exists() and (
    (is_cacheable and not update_refs)
    or (already_resolved and not update_refs)   # False when update_refs=True
    or lockfile_match                            # False when lockfile SHA is stale
)
```

When `update_refs=True` and the lockfile has a stale SHA (common on the first `--update` after a remote push), `skip_download` evaluated to `False` even though the callback had already materialized the correct content. The dep was re-downloaded with identical bytes -- wasted I/O.

Identified during Copilot review of #550 (fix for #548). Reproduced when:
1. `apm_modules/` is empty or the package directory does not exist
2. The lockfile SHA is stale (differs from current remote HEAD)
3. `update_refs=True` (`--update` mode)

## Approach (WHAT)

Add a `_already_resolved_sha_match` short-circuit that fires when:
- `already_resolved=True` (dep is in `ctx.callback_downloaded`)
- `update_refs=True` (only relevant in `--update` mode)
- `resolved_ref` is set and carries a non-cached commit SHA
- `ctx.callback_downloaded[dep_key] == resolved_ref.resolved_commit`

When all four conditions hold, the callback bytes are confirmed correct and the loop sets `skip_download=True`.

## Implementation (HOW)

```mermaid
flowchart TD
    A[sequential loop: already_resolved=True, update_refs=True] --> B{resolved_ref available?}
    B -- No --> C[re-download: cannot confirm remote SHA]
    B -- Yes --> D{callback_sha == resolved_ref.resolved_commit?}
    D -- No --> E[re-download: content may have changed]
    D -- Yes --> F[skip_download=True: callback already fetched correct content]
    F --> G[lockfile refreshed via CachedDependencySource fetched_this_run=True]
```

**Surgical change** -- one new boolean `_already_resolved_sha_match` inserted before `skip_download` in `_resolve_download_strategy` (integrate.py). All existing conditions are preserved unchanged.

**Lockfile refresh**: when `skip_download=True` via this path, `_fetched_now = dep_key in ctx.callback_downloaded` is `True`, so `CachedDependencySource(fetched_this_run=True)` writes `callback_downloaded[dep_key]` as the lockfile SHA -- the stale entry is updated correctly.

## Trade-offs

| Decision | Rationale |
|---|---|
| Require `resolved_ref` | Without a confirmed remote SHA we cannot verify the callback content is still current |
| No local HEAD git check | The callback just wrote the dir; git HEAD is authoritative without an extra repo open |
| Negative gate tests included | SHA mismatch and missing resolved_ref must fall through to normal download |

## Validation evidence

- [+] 3 new targeted unit tests (match, mismatch, no-resolved-ref)
- [+] 53 tests in touched files all green (`uv run --extra dev pytest tests/unit/install/test_integrate_phase3w5.py tests/unit/test_install_update_refs.py -q`)
- [+] `ruff check` / `ruff format --check` clean
- [+] `pylint R0801` 10.00/10
- [+] `scripts/lint-auth-signals.sh` clean

## How to test

```bash
uv run --extra dev pytest tests/unit/install/test_integrate_phase3w5.py::TestResolveDownloadStrategy -v
```

Closes #551

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **BREAKING:** `apm pack --marketplace-output PATH` has been removed. This flag was deprecated in v0.14 with a stderr warning and auto-translated to `--marketplace-path claude=PATH`. Use `--marketplace-path claude=PATH` to override the Claude output path. (#1318)
 
+### Performance
+
+- `apm install --update` no longer re-downloads a dependency when the in-flight
+  resolution callback already fetched it at the correct SHA, eliminating a
+  redundant network round-trip per up-to-date dependency in update mode.
+  (closes #551)
+
 ### Added
 
 - Teams with existing `AGENTS.md` content can now adopt `apm compile` without
--- a/src/apm_cli/install/phases/integrate.py
+++ b/src/apm_cli/install/phases/integrate.py
@@ -213,9 +213,26 @@ def _resolve_download_strategy(
         ref_changed=ref_changed,
     )
 
+    # Issue #551: skip re-download when the BFS callback already fetched this
+    # dep during resolution AND the remote SHA still matches what was captured.
+    # This eliminates redundant network I/O in --update mode when apm_modules/
+    # is empty but the lockfile SHA is stale: the callback downloads the latest
+    # content (recording its SHA), and the sequential loop would otherwise
+    # re-download identical bytes because lockfile_match=False (stale SHA).
+    _callback_sha = ctx.callback_downloaded.get(dep_key)
+    _already_resolved_sha_match = (
+        already_resolved
+        and update_refs
+        and bool(resolved_ref)
+        and bool(_callback_sha)
+        and getattr(resolved_ref, "resolved_commit", None) not in (None, "cached")
+        and _callback_sha == resolved_ref.resolved_commit
+    )
+
     skip_download = install_path.exists() and (
         (is_cacheable and not update_refs)
         or (already_resolved and not update_refs)
+        or _already_resolved_sha_match
         or lockfile_match
     )
 
--- a/src/apm_cli/install/phases/integrate.py
+++ b/src/apm_cli/install/phases/integrate.py
@@ -229,6 +229,9 @@ def _resolve_download_strategy(
         and _callback_sha == resolved_ref.resolved_commit
     )
 
+    if _already_resolved_sha_match and logger:
+        logger.verbose_detail(f"  {dep_key}: callback SHA matches remote -- skipping re-download")
+
     skip_download = install_path.exists() and (
         (is_cacheable and not update_refs)
         or (already_resolved and not update_refs)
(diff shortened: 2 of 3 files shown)


danielmeppiel:
## APM Review Panel Advisory

**Stance: ship_with_followups** -- Clean perf win ships safe; add CHANGELOG entry and a comment on the content-hash interaction before merge.

### Arbitration

This is a straightforward, well-scoped performance optimization that eliminates redundant network round-trips in `--update` mode. The supply-chain-security specialist confirms the skip is sound (both sides derive from remote git, content-hash still gates, race fails safe), and the test-coverage specialist confirms the unit test suite proves the skip fires correctly. No specialist disagreed on correctness.

The two substantive follow-ups are (1) a CHANGELOG entry -- raised independently by doc-writer and oss-growth-hacker -- and (2) the python-architect observation that the content-hash verification block may negate the optimization in stale-lockfile scenarios. The latter is not a bug (it fails safe into a re-download), but it means the PR's perf claim has a narrower effective scope than the title implies. A brief inline comment documents this interaction for future maintainers.

### Principle Alignment

- **Pragmatic:** Eliminates a redundant network hop users never asked for -- exactly the kind of invisible-but-felt improvement that builds CLI loyalty.
- **Secure by default:** Content-hash verification still gates; the skip cannot bypass integrity checks. Fails safe into re-download on mismatch.
- **DevX:** No user-observable contract change; pure latency reduction on slow/rate-limited networks.

### Recommended Follow-ups (folded in this session)

| # | From | Summary | Blocking |
|---|------|---------|----------|
| 1 | doc-writer | Add CHANGELOG [Unreleased] Performance entry for #551 | No |
| 2 | python-architect | Add inline comment explaining content-hash verification interaction with `already_resolved_sha_match` | No |
| 3 | cli-logging-expert | Add `verbose_detail` trace when SHA-match skip fires | No |

> This is an advisory comment from the apm-review-panel. No verdict label is applied. The maintainer makes the merge decision.

PR #1615, opened 2026-06-02
feat: make apm self-update and version-checker respect air-gapped env vars (GITHUB_URL, APM_REPO, VERSION)

### LINKED ISSUE #1231: feat: make `apm update` respect air-gapped environment variables
## Context

Follow-up to #660, which made `install.sh` configurable for air-gapped / GHE environments via `GITHUB_URL`, `REPO`, and `VERSION` env vars.

## Problem

The `apm update` command (and the underlying version-checker) does not honour those same environment variables:

1. **Version check** -- the latest-version lookup always targets the public GitHub API, ignoring `GITHUB_URL` and `REPO`. In an air-gapped network this call will time out or fail.
2. **Installer download** -- the update command fetches the install script from a hardcoded public URL. Even though the downloaded `install.sh` itself now supports the env vars, the download step has no way to reach the script when the public internet is unavailable.

## Expected behaviour

`apm update` should respect the same `GITHUB_URL`, `REPO`, and `VERSION` env vars that `install.sh` supports, so that:

- The version check queries the correct API host.
- The installer script is fetched from the correct mirror / internal host.
- Setting `VERSION` skips the API call entirely (matching `install.sh` behaviour).

## Additional notes

- The `--check` flag should also work against the configured host.
- The version-checker utility used by the update-available notification on every CLI invocation has the same issue and should be fixed together.

danielmeppiel:
## TL;DR

Thread the `GITHUB_URL`, `APM_REPO`, and `VERSION` env vars that `install.sh` already honours (#660) into `apm self-update` (installer download URL) and the version-checker (GitHub API URL). Air-gapped and GHE environments can now run `apm self-update` without reaching the public internet.

## Problem (WHY)

`apm update` (now `apm self-update`) and the background version-checker had two hardcoded public endpoints:

1. **Version check** -- `get_latest_version_from_github` always queried `api.github.com`, ignoring `GITHUB_URL` and `APM_REPO`. In an air-gapped network this times out.
2. **Installer download** -- `_get_update_installer_url` always returned the `aka.ms` shortlink. A custom mirror cannot be reached when the public internet is unavailable.

## Approach (WHAT)

Mirror the semantics established by `install.sh` (see #660):
- `VERSION`: when set, skip the GitHub API entirely and return the pinned version.
- `GITHUB_URL`: override the base GitHub host. For GHE instances the API is addressed at `{GITHUB_URL}/api/v3` (standard GHE REST path).
- `APM_REPO`: override the repository (default `microsoft/apm`).

The installer subprocess inherits all three via `external_process_env()`, so the downloaded `install.sh`/`install.ps1` also honours them automatically.

## Implementation (HOW)

```
src/apm_cli/utils/version_checker.py
  + _get_air_gap_github_url() / _get_air_gap_repo() / _get_air_gap_version()
  + _build_releases_api_url(github_url, repo) -> public api.github.com or {host}/api/v3
  ~ get_latest_version_from_github: reads env vars; short-circuits on VERSION

src/apm_cli/commands/self_update.py
  + _DEFAULT_GITHUB_URL / _DEFAULT_APM_REPO / _INSTALL_SCRIPT_REF constants
  ~ _get_update_installer_url: returns aka.ms shortlink for public GitHub; derives
    {GITHUB_URL}/{APM_REPO}/raw/main/install.{sh,ps1} for any other host
```

### Architecture diagram

```
apm self-update
    |
    +-- VERSION set? --> return pinned version (no network call)
    |
    +-- version_checker.get_latest_version_from_github()
    |       GITHUB_URL == github.com --> api.github.com/repos/{APM_REPO}/releases/latest
    |       GITHUB_URL == ghe host   --> {GITHUB_URL}/api/v3/repos/{APM_REPO}/releases/latest
    |
    +-- _get_update_installer_url()
    |       GITHUB_URL == github.com --> https://aka.ms/apm-unix (unchanged)
    |       GITHUB_URL == custom     --> {GITHUB_URL}/{APM_REPO}/raw/main/install.sh
    |
    +-- subprocess (inherits VERSION, GITHUB_URL, APM_REPO via external_process_env)
```

## Trade-offs

- `VERSION` is consumed from `os.environ` inside `get_latest_version_from_github`; callers that pass `repo=` explicitly still get the env-var short-circuit. This is intentional: pinning a version is a deployment-level concern that should override everything.
- `install.sh` itself hard-codes `api.github.com` even when `GITHUB_URL` is set (pre-existing limitation of the shell script); the Python implementation does it correctly.

## Validation evidence

- 11 new TDD tests (`TestAirGappedEnvVars` in `test_version_checker.py`, `TestInstallerUrlAirGap` in `test_self_update_air_gapped.py`); all URL assertions use `urllib.parse` per test instructions.
- `uv run --extra dev pytest tests/unit/test_version_checker.py tests/integration/test_version_notification.py tests/unit/commands/test_self_update_air_gapped.py tests/unit/commands/test_update_command.py` -> 82 passed.
- `ruff check`, `ruff format --check`, `pylint R0801`, `bash scripts/lint-auth-signals.sh` -> all clean.

## How to test

```bash
# Air-gapped: skips API, uses pinned version, fetches installer from mirror
GITHUB_URL=https://gh.corp.com APM_REPO=corp/apm VERSION=v1.2.3 apm self-update

# GHE version check only (queries {GITHUB_URL}/api/v3)
GITHUB_URL=https://gh.corp.com APM_REPO=corp/apm apm self-update --check

# Regression: public default behavior is unchanged
apm self-update --check
```

## README.md drift note

`README.md` references the `aka.ms` shortlinks and `install.sh` usage but does not describe `apm self-update` env vars. `docs/src/content/docs/getting-started/installation.md` already documents `GITHUB_URL`/`APM_REPO`/`VERSION` for `install.sh`. The new `self-update.md` section now covers `apm self-update` specifically. README update would be additive and requires maintainer approval per docs rules.

Closes #1231


danielmeppiel:
## APM Review Panel -- PR #1615

**Headline:** Air-gapped self-update is a real enterprise unlock; supply-chain blockers are pre-existing risk, not regressions -- ship with follow-up issues.

**CEO stance:** `ship_with_followups`

Ship after folding the verbose-breadcrumb logging and trailing-slash test (done in this iteration). The supply-chain hardening items are real but pre-existing and out of scope -- filed as follow-up issues. This unblocks enterprise adoption with no regression to public-github users.

---

### Folded in this iteration

| Finding | Source | Action |
|---------|--------|--------|
| Docstring says `install.sh` but Windows uses `install.ps1` | Copilot inline | Fixed in docstring |
| Docs table GITHUB_URL row says `install.sh` only | Copilot inline | Fixed in self-update.md |
| No verbose breadcrumb when GITHUB_URL/VERSION overrides active | cli-logging-expert | Added progress logs in self_update command |
| Error message "from GitHub" inaccurate on GHE | cli-logging-expert | Fixed to "from remote" |
| No test for GITHUB_URL with trailing slash | test-coverage-expert | Added test |
| No test for default Windows URL (no env vars) | test-coverage-expert | Added test |

---

### Deferred (recommended follow-up issues)

| Finding | Source | Rationale |
|---------|--------|-----------|
| Installer downloaded without integrity check | supply-chain-security | Pre-existing behavior; hardening requires new checksum infrastructure. File dedicated issue. |
| Auth token sent to GITHUB_URL host | supply-chain-security | Intended behavior for GHE (operator-controlled env var, same threat model as git remote URL). Auth-expert rated nit. |
| APM_REPO not validated before URL interpolation | supply-chain-security | Low-cost check; file follow-up. |
| Single-source env-var table (docs duplication) | doc-writer | Cross-link + consolidation with installation.md; file follow-up. |
| AirGapConfig dataclass refactor | python-architect | Architectural refactor beyond stated scope. |
| Consolidate _DEFAULT_GITHUB_URL / _PUBLIC_GITHUB_URL | python-architect | Minor constant dedup; file follow-up. |
| VERSION pin as --version CLI flag | devx-ux | New UX feature; file follow-up. |
| Enterprise/air-gapped deployment docs page | oss-growth | New docs page; file follow-up. |
| _get_github_token() duplication | auth-expert | Refactor to use core/token_manager; file follow-up. |

---

*Panel: python-architect, cli-logging-expert, devx-ux-expert, supply-chain-security-expert, oss-growth-hacker, auth-expert, doc-writer, test-coverage-expert. CEO arbitration: apm-ceo.*

PR #1619, opened 2026-06-02
feat: apm install -g deploys copilot instructions to ~/.copilot/copilot-instructions.md

### LINKED ISSUE #650: [FEATURE] Copilot CLI user-scope instructions: can apm install -g write copilot-instructions.md
 
 We're trying to use `apm install -g` to bootstrap Copilot CLI for our team. Looking at the source, `targets.py` marks Claude as fully supported at user scope, but Copilot CLI has instructions listed in `unsupported_user_primitives`. Copilot CLI doesn't read `*.instructions.md` files from a subdirectory at user level. It only reads one specific file: `~/.copilot/copilot-instructions.md`. APM doesn't write to that file, so there's no way to get instructions deployed globally for Copilot through APM today.
 
 Support deploying instruction primitives to `~/.copilot/copilot-instructions.md` at user scope. Maybe by concatenating instruction primitives into that single file, or adding a special case mapping for Copilot.
 
 - Placing content in `~/.copilot/agents/*.agent.md`. Copilot picks up the agent, but doesn't load its content into every conversation. Agents are on demand, not 
always on like instructions.
 - Writing `copilot-instructions.md` via a script. Works but means maintaining config outside of APM.
 - Using project level `apm install` for `.github/instructions/`. Works but requires being inside a repo.

danielmeppiel:
Fixes #650.

## TL;DR

`apm install -g` now deploys instruction primitives for the Copilot target by concatenating all `*.instructions.md` files into `~/.copilot/copilot-instructions.md` -- the single file Copilot CLI reads at user scope.

## Problem (WHY)

Copilot CLI reads only `~/.copilot/copilot-instructions.md` at user scope, not individual files in a subdirectory. The copilot target had `"instructions"` in `unsupported_user_primitives`, so `apm install -g` silently skipped instruction primitives for Copilot users entirely.

## Approach (WHAT)

Rather than special-casing the copilot target in imperative code, the fix extends the data-driven `TargetProfile` architecture:

1. A new `user_primitive_overrides: dict[str, PrimitiveMapping] | None` field lets targets swap a primitive's mapping at user scope.
2. `for_scope(user_scope=True)` applies overrides after the `unsupported_user_primitives` filter.
3. A new `format_id = "copilot_user_instructions"` triggers a concat transform instead of per-file deploy.

## Changes

- `targets.py`: add `user_primitive_overrides` to `TargetProfile`, apply in `for_scope()`, update copilot entry
- `instruction_integrator.py`: new `_integrate_copilot_user_instructions()` concat method; `sync_for_target()` handles single output file cleanup
- Tests: 9 new acceptance tests + 4 existing tests updated
- Docs: `targets-matrix.md` updated

## Validation

All 15904 unit tests pass. Full lint chain clean (ruff + pylint R0801 + auth-signal).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `apm install -g` now deploys `instructions` primitives for the Copilot target by concatenating all `*.instructions.md` files from each installed package into `~/.copilot/copilot-instructions.md`, the single file Copilot CLI reads at user scope. Previously this primitive type was silently skipped for global installs. Each package's contribution is wrapped in an HTML provenance comment so the file is auditable and multi-package installs accumulate correctly. (closes #650)
 - `apm compile --target copilot` (and `agents`) no longer writes instructions into `AGENTS.md` when `apm install` has already deployed them to `.github/instructions/`, eliminating duplicate context that Copilot would read from both locations. Mirrors the equivalent dedup behaviour that was already in place for the Claude path (`.claude/rules/`). (closes #1550, refs #1445)
 
 ### Changed
--- a/docs/src/content/docs/reference/targets-matrix.md
+++ b/docs/src/content/docs/reference/targets-matrix.md
@@ -80,7 +80,7 @@ GitHub Copilot (CLI and IDE).
   - skills: `.agents/skills/<name>/SKILL.md`
   - hooks: `.github/hooks/<name>.json`
   - generated: `.github/copilot-instructions.md` (compile output)
-- **User scope.** Partial. `prompts` deploy under `~/.copilot/prompts/`; `instructions` are not supported at user scope. User-scope deploys land under `~/.copilot/`, not `~/.github/`.
+- **User scope.** Partial. `prompts` deploy under `~/.copilot/prompts/`; `instructions` from all packages are concatenated into `~/.copilot/copilot-instructions.md` (Copilot CLI reads only that single file at user scope). User-scope deploys land under `~/.copilot/`, not `~/.github/`.
 
 ## claude
 
--- a/src/apm_cli/integration/instruction_integrator.py
+++ b/src/apm_cli/integration/instruction_integrator.py
@@ -96,6 +96,17 @@ class InstructionIntegrator(BaseIntegrator):
         deploy_dir.mkdir(parents=True, exist_ok=True)
 
         fmt = mapping.format_id
+
+        if fmt == "copilot_user_instructions":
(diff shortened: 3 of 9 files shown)


danielmeppiel:
## apm-review-panel advisory

**PR:** feat: apm install -g deploys copilot instructions to ~/.copilot/copilot-instructions.md
**Panel:** python-architect | cli-logging-expert | devx-ux-expert | supply-chain-security-expert | oss-growth-hacker | doc-writer | test-coverage-expert | auth-expert (inactive)
**CEO verdict:** ship-with-changes -- all blocking and recommended items folded in commit fcdbfbb1

---

### Fixes applied in this session

**[blocking] Multi-package collision -- silent data loss (devx-ux-expert + 3 others)**
_instruction_integrator.py_ -- `_integrate_copilot_user_instructions` called `check_collision` which always blocked the second package in a session (file written by pkg-1 not in `managed_files` for pkg-2). Fixed: first write now stamps an APM-managed sentinel; subsequent writes detect it and append their section instead of colliding. All 11 acceptance tests pass.

**[recommended] Per-package provenance markers (supply-chain-security-expert, devx-ux-expert)**
Each package's contribution is now wrapped:
```
...body...
```
Enables attribution audit and surgical section replacement on re-install.

**[recommended] for_scope() dynamic-root path skips user_primitive_overrides (python-architect)**
_targets.py_ -- applied the override merge in the `user_root_resolver` branch before early return, matching the static-root path. Latent correctness bug for future targets that combine both fields.

**[nit] Windows CRLF frontmatter not stripped (python-architect, test-coverage-expert)**
_instruction_integrator.py_ -- regex changed from `\n` to `\r?\n`. New test: `test_strip_frontmatter_strips_crlf_line_endings`.

**[nit] ensure_path_within missing in concat path (supply-chain-security-expert)**
Added for defense-in-depth pattern consistency.

**[process] CHANGELOG entry (oss-growth-hacker)**
Added to [Unreleased] > Fixed framing the Copilot user-scope instructions narrative.

**[test] Multi-package regression test (test-coverage-expert)**
`test_multi_package_instructions_all_concatenated` -- installs two packages sequentially, asserts both packages' content is present in the output file.

---

### Deferred (out of scope)

- **User consent/preview before first write** (supply-chain-security-expert) -- requires an interactive confirmation flow; tracked separately.
- **Link resolution in concat path** (python-architect) -- relative links are meaningless at user scope; no resolution needed. Comment added in code for clarity.
- **resolve_links call** -- deferred with code comment; user-scope instructions have no meaningful link targets.

---

### Mergeability

All Copilot inline findings folded. Lint: clean. Tests: 11/11 new pass, 2346/2346 existing pass. CI watch in progress.

PR #1606, opened 2026-06-02
fix: surface file-missing error before marker search in managed_section mode

danielmeppiel:
## TL;DR

When `agents_md_mode: managed_section` is set but the target file does not exist, APM now immediately raises a clear, actionable error instead of the confusing 'markers not found' message.

## Problem (WHY)

When a user enables `managed_section` mode on a fresh workspace where `AGENTS.md` has not yet been created, the previous code path silently passed an empty string (`""`) to `apply_managed_section`. This caused a `ManagedSectionError` with the message 'markers not found' -- implying the file exists but is missing the markers. The real root cause is that **the file does not exist at all**, which is a distinct first-run UX failure that demands its own clear diagnosis.

Origin: apm-review-panel advisory on PR #1589 (devx-ux-expert finding).

## Approach (WHAT)

Detect a missing target file **before** calling `apply_managed_section` and raise `ManagedSectionError` immediately with:

```
<path> does not exist yet. Create it with the managed-section markers first,
or use mode: full for initial generation.
```

## Implementation (HOW)

```
src/apm_cli/compilation/agents_compiler.py
  _write_output_file_with_config(): replace the ternary
    existing = target.read_text(...) if target.exists() else ""
  with an explicit file-existence check that raises ManagedSectionError.

tests/unit/compilation/test_managed_section.py
  TestManagedSectionWriteIntegration: add
    test_write_output_file_managed_section_file_missing()
  -- TDD: written first (RED), then implementation made it GREEN.

docs/src/content/docs/producer/compile.md
  Constraints list: added bullet documenting the file-must-exist
  requirement with guidance to use mode: full for the first run.
```

## Trade-offs

- **No behavior change for the file-exists case** -- only the missing-file branch is affected.
- Keeps `ManagedSectionError` as the exception type (consistent with all other marker-related errors, so callers need no new catch clauses).
- Message is 100% ASCII (per encoding rules).

## Validation evidence

```
uv run --extra dev pytest tests/unit/compilation/test_managed_section.py -q
21 passed in 0.48s

uv run --extra dev ruff check src/ tests/  -- All checks passed
uv run --extra dev ruff format --check src/ tests/  -- 1165 files already formatted
uv run --extra dev python -m pylint --disable=all --enable=R0801   --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/  -- 10.00/10
bash scripts/lint-auth-signals.sh  -- [+] auth-signal lint clean
```

## How to test

[...]

PR #1627, opened 2026-06-02
fix(compilation): raise clear error when managed_section file is missing

yashpal-2004:
## Problem
When `agents_md_mode: managed_section` is enabled but the target output file (e.g., `AGENTS.md`) does not exist yet, the compiler falls back to checking for markers in an empty string. This triggers a confusing `'markers not found'` error, even though the actual issue is that the file itself is missing.

## Proposed Changes
- Added an explicit check for target file existence in `_write_output_file_with_config` when `managed_section` mode is enabled.
- If the file does not exist, raise a specific, descriptive `ManagedSectionError`:
  `AGENTS.md does not exist yet. Create it with markers first, or use mode: full for initial generation.`
- Added unit test `test_write_output_file_managed_section_missing_file` to `tests/unit/compilation/test_managed_section.py` to ensure correct behavior and error message generation.

## Verification
- Ran the unit tests locally under a Python 3.12 virtual environment:
  `pytest tests/unit/compilation/test_managed_section.py` (all tests passed, including the new test).
- Ran `ruff check` on the modified files (no linting/formatting errors).


yashpal-2004:

danielmeppiel:
Thanks for this fix, @yashpal-2004 -- the intent here is exactly right. Closing as superseded: the same "raise a clear error when a managed_section file is missing" guard already landed on main via #1606 (merged in 7032bc0), so the missing-file path is now covered. We really appreciate you flagging and fixing it; please keep the contributions coming.

Issue #1326, opened 2026-05-14
[BUG] Cross-repo bare repo on *.ghe.com marketplace silently resolves at github.com on validation success (dependency-confusion vector)


edenfunf:
## Context

Surfaced by the PR review panel for #1319 (supply-chain-security finding). PR #1319 fixes the visible **failure** path of #1305 -- when a cross-repo bare `repo` on a `*.ghe.com` marketplace fails validation, the install surfaces an actionable host-qualify hint. The panel correctly noted that #1319 deliberately does not address the silent **success** path of the same syntactic ambiguity.

## The gap

The two intents that share the bare cross-repo syntax are:

- **Intentional cross-host**: marketplace on `corp.ghe.com`, `repo: opensource-org/awesome-tool` -- author wants the github.com open-source dep.
- **Misconfigured same-host**: marketplace on `corp.ghe.com`, `repo: platform-team/shared-tool` -- author wants the corp.ghe.com enterprise dep.

Resolver-level disambiguation is impossible. PR #1319 covers the case where the misconfigured intent's resolution fails validation (because the same `owner/repo` is not on github.com), but **does not cover** the case where an attacker pre-stages `platform-team/shared-tool` on github.com with malicious content:

1. Operator publishes a `*.ghe.com` marketplace with `repo: platform-team/shared-tool` intending the enterprise repo.
2. Attacker registers `platform-team/shared-tool` on github.com first (the enterprise org's namespace is observable from public marketplace.json files in many setups).
3. `apm install shared-tool@my-marketplace` resolves canonical to bare `platform-team/shared-tool/plugins/shared`; `DependencyReference.parse` defaults host to `github.com`; `_validate_package_exists` returns True against the attacker-staged repo.
4. Install succeeds, lockfile records `https://github.com/platform-team/shared-tool` as the origin, attacker content executes under the operator's APM context.
5. The `CrossRepoMisconfigRisk` sentinel #1319 attached **is consumed only on validation failure**; the success path never reaches the hint emission branch.

This is a dependency-confusion-class vector (cf. the npm internal-package-name confusion attacks). Mitigations span resolver + install layers and cannot fit inside #1305's scope without changing the resolver routing PR #1292 deliberately preserved.

## Why this is not part of #1319

[...]

Issue #1, opened 2025-09-22
Why do we need a GitHub token?


svallory:
Please clarify in the README why APM requires a github access token

svallory:
Note: I just found why in this file: https://github.com/danielmeppiel/apm/blob/main/docs/getting-started.md

And is not _actually_ required. So I would suggest clarifying that in the README as people may feel uncomfortable giving an access token

PR #1628, opened 2026-06-02
feat(install, compile): `--root DIR` to redirect writes while sources resolve from $PWD

### LINKED ISSUE #888: enhancement: `apm install --root <dir>` to target a directory other than $PWD

danielmeppiel:
Closes #888. Supersedes #928 (re-ported onto current `main`).

## Why this supersedes #928
#928 was accepted (issue #888) but went 187 commits stale with 10
conflicts (CONFLICTING/DIRTY). I re-applied @srid's feature **semantically**
onto current `main` rather than textually merging. The single commit on
this PR preserves **Sridhar Ratnakumar** as author; I could not force-push
to the original `juspay/apm` branch because GitHub disables maintainer-edit
pushes for **organization-owned forks**, hence this superseding PR.

## What it does
Adds `--root DIR` to both `apm install` and `apm compile`: all generated
writes (apm_modules, lockfile, integrated primitives, AGENTS.md/CLAUDE.md)
land under `DIR`, while primitives and the constitution continue to
resolve from `$PWD`. This lets you deploy compiled context into a
scratch/deploy tree without polluting the source checkout.

## Design (folded from an APM review-panel pass)
- Source-root override lives in a `contextvars.ContextVar` (thread-safe),
  not a module global; the redirect context manager hardens its `finally`
  to always restore cwd and clear the override.
- `<!-- Source: ... -->` provenance comments render relative to the source
  dir (not the deploy root) in both the single-file and distributed
  paths, keeping output **byte-identical** under `--root`.
- The install resolver anchors its manifest to the source root **only**
  when `--root` is active, so non-`--root` installs are unaffected (zero
  regression).
- Extracted a shared `resolve_base_and_source_dirs` helper to keep the
  two compiler formatters duplication-free (pylint R0801 clean).

## Guards
- `compile --root` + `--watch` -> exits non-zero with a teaching message.
- `install --root` + `--global` (user scope) -> exits non-zero likewise.

## Validation
- New: 8 unit mechanism tests + 3 e2e tests (compile byte-identical &
  sequential-leakage; install resolver source_root != project_root).
- `tests/unit/install/` + `tests/unit/compilation/`: 2616 passed.
- Full lint chain green (ruff check/format, pylint R0801, auth-signals,
  file-length, relpath guards).
- Manual byte-identical smoke verified for both `compile --root` and
  `install --root`.
- Docs updated: `cli/install.md`, `cli/compile.md`, apm-usage commands,
  CHANGELOG (credits @srid / juspay).


danielmeppiel:
## APM Review Panel: ship-with-followups

> Well-motivated --root DIR feature ships with one architectural anti-pattern (dual-truth ContextVar) and two missing conflict-guard tests that should fold before merge.

All eight panelists converged on ship-readiness; no dissent on severity classification. The Python Architect's PA-1 (dual-truth between ContextVar and ctx fields) and the Test Coverage Expert's TC-2 (untested manifest_anchor divergence) describe the same underlying bug surface from different angles -- the ContextVar query in resolve.py creates a path where ctx.source_root and the global override can disagree, and no test pins the behavior when they do. Supply Chain and Python Architect independently flagged the misleading resolve() symlink comment (PA-4 / SC-1), confirming it is a real documentation-accuracy gap, not opinion. The DevX UX expert floated --prefix as an alternative name; the consensus is --root -- it matches the PR's semantics (root of the output tree) and avoids conflation with npm's prefix which also affects bin resolution. The OSS Growth Hacker's CI recipe suggestion (GH-1) is strategically sound but is post-merge scope.

**Aligned with:** portability-by-manifest (manifest resolution remains anchored to $PWD regardless of --root, preserving the source-of-truth contract); secure-by-default (no new attack surface; user opts in explicitly; symlink comment fix needed to avoid false confidence); governed-by-policy (conflict guards enforce policy but lack test assertions to survive refactoring).

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 2 | 2 | Well-documented source-root/deploy-root separation using ContextVar + os.chdir. One dual-truth anti-pattern in resolve.py (ContextVar check instead of ctx fields) and a misleading symlink comment; the AgentsCompiler source_dir threading is well-structured. |
| CLI Logging Expert | 0 | 2 | 2 | Error messages are well-crafted. However, the --root redirect is completely silent in default and verbose mode -- no progress/info line confirming where artifacts will land. |
| DevX UX Expert | 0 | 1 | 3 | Well-motivated flag with clear precedents (pip --target, npm --prefix). Flag naming, conflict guards, and exit codes are consistent. Help text is too dense for a single --help line. |
| Supply Chain Security Expert | 0 | 1 | 2 | No new supply chain attack surface. Misleading resolve() symlink comment gives false security confidence; overall risk is moderate since user explicitly provides --root. |
| OSS Growth Hacker | 0 | 3 | 3 | Strong feature for CI/deploy adoption -- pip install --target analogy is exactly right. Contributor attribution is exemplary community practice. Main gaps: discoverability and a missed launch beat opportunity. |
| Auth Expert | 0 | 0 | 0 | APM's auth stack resolves credentials exclusively via environment variables and protocol-keyed git credential helpers -- none depend on cwd for credential lookup. No auth risk identified. |
| Doc Writer | 0 | 3 | 2 | Documentation is well-written and matches house style. CHANGELOG entry omits (#1628) PR backref; attribution phrasing is non-standard. Reference pages should cross-link and document mutual exclusions. |
| Test Coverage Expert | 0 | 2 | 0 | PR ships solid coverage for the happy path. Two conflict-guard paths (--root+--global, --root+--watch) lack assertions. The manifest_anchor divergence when source_root is set without override is untested. |

### Top 5 follow-ups

1. **[Python Architect]** Replace the ContextVar query in resolve.py:720 with a ctx.source_root comparison -- the dual-truth anti-pattern is the only architectural defect; fix is a one-line guard change that eliminates a class of future bugs.
2. **[Test Coverage Expert]** Add tests asserting exit-code 2 for --root+--global and --root+--watch conflicts -- conflict guards are user-facing promises; without tests they silently regress on next refactor.
3. **[Test Coverage Expert]** Add a test pinning manifest_anchor behavior when source_root diverges without the override -- validates the same surface PA-1 fixes; the test survives even if the fix approach changes.
4. **[Supply Chain Security Expert]** Correct the resolve() symlink comment to state it follows symlinks -- factually wrong security comment misleads future maintainers; 10-second fix.
5. **[CLI Logging Expert]** Add one info-level log line confirming redirect (source -> deploy root) -- silent redirects frustrate debugging in CI; one line closes the observability gap.

### Architecture

```mermaid
classDiagram
    direction LR
    class InstallContext {
        <<Dataclass>>
        +project_root Path
        +apm_dir Path
        +source_root Path
        +__post_init__()
    }
    class InstallScope {
        <<Enum>>
        PROJECT
        USER
    }
    class install_root_redirect {
        <<ContextManager>>
        +root Path
        +dry_run bool
    }
    class SourceRootOverride {
        <<ContextVar>>
        +set(path)
        +get() Path|None
    }
    class get_source_root {
        <<Pure>>
        +scope InstallScope
    }
    class AgentsCompiler {
        +base_dir Path
        +source_dir Path
        +compile()
    }
    install_root_redirect ..> SourceRootOverride : sets/clears
    get_source_root ..> SourceRootOverride : reads
    InstallContext ..> get_source_root : populated from
    note for install_root_redirect "os.chdir(root) + ContextVar pin\nProcess-global side effect"
```

```mermaid
flowchart TD
    A["CLI: apm install --root DIR"] --> B{"--root and --global?"}
    B -- yes --> ERR["click.UsageError exit 2"]
    B -- no --> C["install_root_redirect(root)\n[FS] mkdir(root)"]
    C --> D["[FS] os.chdir(root)"]
    D --> E["set_source_root_override($PWD)"]
    E --> F["run_install_pipeline()"]
    F --> G["get_deploy_root(scope)\nreturns Path.cwd() == root"]
    F --> H["get_source_root(scope)\nreads SourceRootOverride"]
    H --> I["ctx.source_root = $PWD"]
    G --> J["ctx.project_root = root"]
    I --> K["resolve phase"]
    J --> K
    K --> L{"ContextVar override\nis not None?"}
    L -- yes --> M["manifest_anchor = ctx.source_root"]
    L -- no --> N["manifest_anchor = ctx.apm_dir"]
    M --> O["resolver.resolve_dependencies(anchor)\n[FS] reads apm.yml from $PWD"]
    N --> O
    O --> P["integrate phase\n[FS] writes to root/apm_modules/"]
    P --> Q["finally: os.chdir(original)\nset_source_root_override(None)"]
```

### Recommendation

The feature is well-designed, correctly scoped, and closes a real user need (CI deploy roots). Priorities 1-4 are small, bounded fixes that should fold into this PR before merge -- they total approximately 30 lines of code and one comment correction. Priorities 5-8 can land as fast-follows without blocking. No security blocker, no breaking change, strong contributor attribution.

---

<details>

#### Python Architect

- **[recommended]** resolve.py uses global ContextVar check instead of ctx fields (dual-truth anti-pattern) at `src/apm_cli/install/phases/resolve.py:720`
  Line 720 reads `manifest_anchor = ctx.source_root if get_source_root_override() is not None else ctx.apm_dir`. This queries the process-global ContextVar to decide behavior, bypassing the explicit ctx fields. The correct guard is `ctx.source_root != ctx.project_root` since `__post_init__` defaults source_root to project_root. Querying the ContextVar leaks implementation detail into a phase that should only read ctx and creates a dual-truth bug surface where ctx.source_root and the ContextVar could disagree in tests or embedded callers.
- **[recommended]** os.chdir() is process-global -- fragile under concurrency and test parallelism at `src/apm_cli/install/root_redirect.py:80`
  chdir is process-global state not isolated by contextvars, threads, or asyncio tasks. Parallel test runners sharing the same process will race on cwd. Recommend adding a doc comment in pipeline.py stating this mechanism is safe ONLY for single-threaded, single-invocation CLI entry points.
- **[nit]** source_root field uses `Path = None # type: ignore` instead of proper Optional typing at `src/apm_cli/install/context.py:44`
  Should use `source_root: Path | None = None`.
- **[nit]** root_redirect.py comment incorrectly claims Path.resolve() avoids symlinks at `src/apm_cli/install/root_redirect.py:72`
  Path.resolve() actually FOLLOWS symlinks to produce a canonical path -- it does not avoid them.

#### CLI Logging Expert

- **[recommended]** Missing info line when --root redirects writes at `src/apm_cli/install/root_redirect.py:75`
  No logger.info or logger.start line confirms the redirect. A single `[i] Writing to /deploy/root (sources from /src/path)` right after the context manager enters would address the newspaper test.
- **[recommended]** No --verbose breadcrumb for source/deploy root split at `src/apm_cli/install/pipeline.py:397`
  In verbose mode no output shows which directories resolved for source_root vs project_root. Something like `[i] source_root={source_root}, deploy_root={project_root}` would aid debugging when paths go wrong.
- **[nit]** UsageError for dry-run + missing root could be tighter at `src/apm_cli/install/root_redirect.py:64`
  The message is solid but slightly long. Consider leading with the thing that is wrong.
- **[nit]** Install summary does not mention deploy root at `src/apm_cli/commands/install.py:1622`
  When --root is active the summary should include the target path so the output is self-contained. Follow-up enhancement.

#### DevX UX Expert

- **[recommended]** Help text for --root (install) is too dense for a single --help line at `src/apm_cli/commands/install.py:1127`
  The help string is 67 words and names 6 specific output directories. Consider shortening to one sentence and letting the reference doc carry the full enumeration.
- **[nit]** Consider --prefix instead of --root for npm familiarity at `src/apm_cli/commands/install.py:1119`
  npm's equivalent is `--prefix`. --root is defensible and matches the PR's semantics; flagging for awareness.
- **[nit]** Dry-run error wording could be slightly clearer at `src/apm_cli/install/root_redirect.py:62`
  The second half "or drop --dry-run to let install/compile create it" is slightly confusing.
- **[nit]** Compile docs say "Mirrors pip install --target" but compile is not installing at `docs/src/content/docs/reference/cli/compile.md:77`
  A more intuitive framing for compile would be "Mirrors make DESTDIR=..." or simply "Redirects output files to DIR."

- **[recommended]** Misleading comment: resolve() follows symlinks, does not defend against them at `src/apm_cli/install/root_redirect.py`
  The comment states "resolve() collapses any symlinks... so we never silently follow a planted symlink." This is incorrect: Path.resolve() resolves THROUGH symlinks. The security claim in the comment is misleading and should be corrected.
- **[nit]** TOCTOU between resolve()+mkdir and os.chdir() at `src/apm_cli/install/root_redirect.py`
  Small race window between resolve/mkdir and chdir. Defense-in-depth nit; low likelihood for a CLI tool.
- **[nit]** ContextVar source-root override is process-global in async contexts at `src/apm_cli/install/root_redirect.py`
  Worth a doc comment noting the single-threaded/single-task invariant.

#### OSS Growth Hacker

- **[recommended]** Add a CI/build pipelines recipe page
  Users searching "apm CI", "GitHub Actions apm", or "read-only source tree" won't land on the cli/install reference page. A short recipe guide with a GitHub Actions snippet using `--root $RUNNER_TEMP/artifacts` would convert searchers with existing intent.
- **[recommended]** Surface --root in the quickstart "next steps" or templates
  A single tip like "use `--root build/` to keep generated files out of your source tree" bridges first-use to production-use.
- **[recommended]** CHANGELOG entry is a launch beat -- rewrite lede for release announcement
  For the release announcement, rewrite as: "APM now supports out-of-tree builds (--root DIR) -- deploy artifacts anywhere without touching your source tree. Perfect for CI staging, monorepo scratch builds, and read-only containers."
- **[nit]** Amplify contributor shout-out in release notes
  Promote @srid attribution to a visible "Community contributions" callout with a PR link.
- **[nit]** --help text should lead with the "why", not the mechanic
  Lead with the use-case: "Redirect all generated files to DIR (useful for CI staging and read-only source trees)."
- **[nit]** Add keyword anchors for search engines
  Ensure "out-of-tree build", "CI artifact staging", and "read-only source tree" appear verbatim in the reference docs.

#### Auth Expert

No findings.

#### Doc Writer

- **[recommended]** CHANGELOG entry missing (#1628) PR backref
  Other entries pair issue close with PR number (e.g. "(closes #1326) -- by @handle (#1459)"). This entry ends with "(closes #888)" only. Add "(#1628)" for traceability.
- **[recommended]** install.md and compile.md --root sections should cross-link
  --root is a symmetric concept across both commands. Add a cross-reference in each --root section so readers discover the related command.
- **[recommended]** Document --root mutual exclusions in the flag tables
  install.md already documents flag exclusivity inline (e.g. --update/--frozen). The --root rows should mirror this: "Rejected with --global" (install) and "Rejected with --watch" (compile).
- **[nit]** Attribution phrasing diverges from "-- by @handle" convention
  Every other credited entry uses "-- by @handle" style. Normalize "Thanks to @srid (juspay)" to "-- by @srid (#1628)".
- **[nit]** Keep skill commands.md --root a one-line summary
  Verify the skill doc addition did not copy the full multi-sentence explanation; it should defer to the reference docs for full behavior.

#### Test Coverage Expert

- **[recommended]** --root + --global / --root + --watch conflict guards untested
  Neither conflict path has a test asserting exit code 2 and error message. Suggest two parametrized integration tests: subprocess with conflicting flags, assert returncode==2 and stderr contains the message.
- **[recommended]** manifest_anchor divergence path untested when source_root set without override
  In resolve.py, if source_root != project_root but the ContextVar override is not set, manifest_anchor falls back to ctx.apm_dir rather than ctx.source_root. A test should pin this behavior so it cannot silently drift.

</details>

PR #1288, opened 2026-05-12
fix: add support for custom github endpoints (host.tld/owner/repo) as apm.yml entrypoint for command apm pack (marketplace)

abi-jey:
## Description

Currently apm pack (on apm yaml for marketplace block) does validate only github.com or local file. This PR extends the validation and package entries to support the items hosted in github enterprise for example with including the host (optional if exists).

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass
- [x] Added tests for new functionality (if applicable)

Issue #1010, opened 2026-04-27
feat: ADO marketplace support (marketplace.yml with Azure DevOps repos)


sergio-sisternes-epam:
## Summary

`apm marketplace build` currently assumes `owner/repo` as the package source format (enforced by `SOURCE_RE` in `yml_schema.py`). Azure DevOps repositories use a 3-part path (`org/project/_git/repo`) which doesn't fit this schema.

## Context

This came up during the investigation for #1008 (GHE support). While `RefResolver` and `build_https_clone_url()` now work for any `GITHUB_HOST` value, ADO has structural differences that require dedicated support:

1. **URL structure**: ADO repos use `https://dev.azure.com/{org}/{project}/_git/{repo}` -- the 3-part path doesn't fit `owner/repo`
2. **Schema validation**: `SOURCE_RE = r"^[^/]+/[^/]+$"` in `yml_schema.py` only allows 2-segment paths
3. **Metadata API**: ADO uses a different REST API (`/_apis/git/repositories/`) than GitHub
4. **Authentication**: ADO uses `ADO_APM_PAT` or `az` CLI, not `GITHUB_APM_PAT`

## Scope

- Extend `SOURCE_RE` to accept ADO-style 3-part paths (or add `host` field to `marketplace.yml` package entries)
- Add ADO URL construction for `git ls-remote`
- Add ADO metadata fetch via ADO REST API
- Wire ADO auth (`ADO_APM_PAT` / `az` CLI) into the marketplace builder
- Tests and documentation

## Related

- #1008 -- GHE marketplace support (landed)
- `src/apm_cli/utils/github_host.py` -- `build_ado_https_clone_url()` already exists
- `src/apm_cli/core/auth.py` -- `AuthResolver` already supports ADO hosts

leocamello:
## Same host-stripping blocks self-managed GitLab on the consumer install path

We hit this while building the authoring side of #1519 (`sourceBase` for nested GitLab groups). #1010 reads as primarily an Azure DevOps authoring concern today, but the **consumer install path has the same root cause** and it blocks self-managed GitLab too — so here is a precise, fully generic reproducer and a fix direction in case it helps scope this.

### Symptom

`apm install <pkg>@<marketplace>` fails for any marketplace whose packages live on a non-default host. Discovery (`apm marketplace add` / `list` / `browse` / `search`) works; only install fails — it routes to `github.com` and suggests the GitHub token, even though the package is on GitLab and the GitLab token is configured and verified.

[...]

PR #1731, opened 2026-06-10
feat(marketplace): add sourceBase for nested package sources (#1519)

### LINKED ISSUE #1519: marketplace.packages[].source: generalize beyond `host/owner/repo` (deeply-nested GitLab subgroups, arbitrary git hosts)
## Summary

PR #1288 (merged) extended `marketplace.packages[].source` to accept `<host.tld>/<owner>/<repo>` and `https://<host.tld>/<owner>/<repo>[.git]`. That unblocks GitHub Enterprise and shallow GitLab layouts, but the regex pins exactly **three** path segments (host + owner + repo), and the test suite explicitly rejects 4+ segments (`test_four_segment_path_rejected`).

Self-hosted GitLab deployments commonly nest projects under multiple subgroups, e.g.:

```
gitlab.example.com/group/sub-group/team/projects/my-package
```

That's host + 4 path segments + repo. With the post-#1288 schema there is still no `source` form that names this repo — the schema rejects every URL form, the `git: { url: ... }` object form is rejected as "must be a non-empty string", and the bare `<owner>/<repo>` shorthand silently routes to **github.com** regardless of `marketplace.owner.url`. The `apm view` resolver, by contrast, already handles arbitrary-depth host-prefixed paths correctly — so this is a marketplace-schema gap, not a resolver-capability gap.

## Repro

`apm.yml`:

```yaml
name: test-mp
version: 0.1.0
description: repro
marketplace:
  owner:
    name: my-group
    url: https://gitlab.example.com/group/sub-group/team/projects
  outputs:
    claude: {}
  build:
    tagPattern: "v{version}"
  packages:
    - name: my-package
      description: deeply nested repo
      source: gitlab.example.com/group/sub-group/team/projects/my-package
      version: "^0.1.0"
```

```
$ apm marketplace check
[x] marketplace config error: 'packages[0].source' must be one of
    '<owner>/<repo>', '<host.tld>/<owner>/<repo>',
    'https://<host.tld>/<owner>/<repo>[.git]', or './<path>',
    got 'gitlab.example.com/group/sub-group/team/projects/my-package'
```

The object form sometimes recommended in authoring guides also fails:

```yaml
      source:
        git: https://gitlab.example.com/group/sub-group/team/projects/my-package.git
```

```
[x] marketplace config error: 'packages[0].source' must be a non-empty string
```

## Evidence the resolver itself isn't the blocker

```
$ apm view gitlab.example.com/group/sub-group/team/projects/my-package versions
Available versions:
  v0.1.0    tag      5d908cad
  main      branch   5d908cad
```

## Evidence `marketplace.owner.url` is unused at resolution time

With:

```yaml
marketplace:
  owner:
    url: https://gitlab.example.com/group/sub-group/team/projects
  packages:
    - source: my-group/my-package          # the only schema-accepted shorthand
```

I wrapped `git` on `PATH` to log every invocation. `apm marketplace check` ran:

```
git ls-remote --tags --heads https://github.com/my-group/my-package.git
```

The shorthand routed to **github.com**, ignoring `owner.url`. That's consistent with the schema docs (§7.3): `owner.url` is documented as *"Owner homepage"* — descriptive metadata, not a resolution input.

## Why this matters

Self-hosted GitLab with subgroup nesting is the canonical enterprise pattern in many orgs. Until the schema accepts a form that names these repos, APM marketplaces can't list them — and the only schema-accepted shorthand silently exfiltrates the lookup to github.com under whatever auth happens to be in scope, which is also a small footgun for anyone publishing a marketplace from an enterprise tenant.

## Suggested directions

Two options, not mutually exclusive:

### Option A (preferred): a marketplace-level git base, packages name themselves relative to it

This is the shape every other package manager has converged on — `npm`'s `registry`, `cargo`'s `[registries]`, Maven's `<repositories>`. Declare the host/path once, then entries don't repeat it. It sidesteps the "how many slashes" parser problem entirely.

A new field is cleaner than repurposing `owner.url` (which is documented as "Owner homepage" and the official example, `https://github.com/contoso`, is a profile page with no git semantics):

```yaml
marketplace:
  sourceBase: https://gitlab.example.com/group/sub-group/team/projects
  owner:
    name: my-group
    url: https://gitlab.example.com/groups/group/sub-group/team/projects   # still a homepage
  packages:
    - name: my-package
      source: my-package                             # → sourceBase + my-package

    - name: external
      source: github.com/some-org/external           # #1288's host-prefixed form
                                                      # overrides sourceBase
```

Properties:
- Mixed-host marketplaces still work — #1288's `host.tld/owner/repo` form is the per-package escape hatch.
- The marketplace stays portable: forking/relocating means changing one line, not every entry.
- `sourceBase` is a strict prefix concatenation (no segment counting), so it composes with arbitrary GitLab subgroup depth, Bitbucket workspaces, Gitea orgs, etc.
- Validation: require `https://` scheme, reject userinfo / port / query / fragment (same security posture as #1288).

(Field name `sourceBase` is illustrative — `host`, `defaultHost`, `registry`, etc. are all reasonable.)

### Option B (mechanical fallback): widen the source regex to accept `>3` path segments

Treat the last two slash-separated segments as `owner/repo`, everything before as `host + group path`. Same security guards from #1288 (no userinfo, no port, no query, https-only for URL form). Less elegant than Option A but a smaller diff if `sourceBase` is too big a change to land soon.

These compose well — Option A is the principled fix, Option B is available for one-off packages that don't fit the marketplace's base.

## Decision needed from maintainers

Before anyone invests in implementation, it would help to know which direction lands cleanly with your design intent for the marketplace block:

- **Option A only** — introduce a marketplace-level base field; keep the source regex as-is.
- **Option B only** — widen the source regex to arbitrary depth; don't introduce a new field.
- **Both** — Option A as the primary path, Option B as the per-package escape hatch (my read of the trade-offs, but happy to be redirected).
- **Neither / different shape** — there's a direction we haven't considered.

Once a direction is chosen I'm happy to take a stab at the implementation PR (schema + tests + docs), tested against an actual deeply-nested self-hosted GitLab on our side. Just need the maintainer steer first to avoid building the wrong thing.

## Out of scope (suggested for separate issues)

- SSH SCP form (`git@host:path`) — explicitly rejected by #1288 for confused-deputy reasons; not asking to revisit here.
- Non-`https` URL schemes — same.
- Per-host auth wiring beyond what #1288 already added.

## References

- #1288 (merged 2026-05-26) — adds 3-segment `host.tld/owner/repo` form; pins the limit via `test_four_segment_path_rejected`.
- #1472 — handles arbitrary GitLab subgroup depth on the *install* side via Artifactory probe. Different surface, but informs how the marketplace schema could think about depth.
- #1195 (closed) — earlier GitLab host support pass.
- #676, #692 — adjacent source-shape parity work.

leocamello:
## Description

Adds an optional `marketplace.sourceBase` field so marketplace authors on
self-managed hosts (canonical case: GitLab nested subgroups) can declare a git
base once and write each package `source` relative to it. The per-entry `source`
shorthand tops out at `host.tld/owner/repo` (3 segments); enterprise GitLab
nesting is routinely deeper (`.../group/sub-group/team/project/<pkg>`), which the
shorthand cannot express.

```yaml
marketplace:
  sourceBase: https://gitlab.example.com/group/sub-group/team/projects
  packages:
    - name: my-package
      source: my-package                  # -> .../projects/my-package
      version: "^1.0.0"
    - name: from-github
      source: github.com/acme/helper      # host-prefixed: overrides the base
      ref: v1.0.0
```

Semantics ("relative to base, always"): a host-less `source` composes onto the
base; a `source` carrying its own host (`host.tld/owner/repo` or full URL) or a
local `./` source overrides the base; with no `sourceBase` declared, behaviour is
byte-for-byte unchanged and a host-less single-segment source stays rejected
(fail-closed — no silent routing to the default host).

**Implementation:** composition happens at parse time (`yml_schema.py`) — a
relative entry is rewritten to the equivalent host-prefixed form, so the
resolver, builder, and output mappers treat it identically to a #1288
host-prefixed entry with no changes to those modules. `sourceBase` is validated
with #1288's posture (https-only, FQDN host, no userinfo/port/query/fragment/
`.git`) plus `validate_path_segments`. `apm marketplace plugin add` accepts
relative sources when a base is declared.

**Also fixes `apm marketplace check`:** it was building a single default-host
resolver with no token and ignoring `entry.host`, so it failed (`git ls-remote`
exit 128) for any non-default host — every `sourceBase` entry, and the
pre-existing #1288 host-prefixed form. It now uses the same per-host resolver +
per-host token pattern as `apm pack` (extracted into a shared
`marketplace/auth_helpers.resolve_token_for_host`) and short-circuits local
packages, so `check` and `pack` agree.

Validated end-to-end against a real self-managed GitLab (4-segment nested group):
`apm marketplace check` exits 0, and `apm pack` emits a `source: url` entry with
the composed URL and a real resolved `sha`.

The consumer-side install gap for non-default-host marketplaces
(`apm install <pkg>@<marketplace>` strips the host) is independent, reproduces on
stock 0.18.0, and is tracked in #1010 — out of scope here.

Fixes #1519

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

`tests/unit/marketplace/test_source_base.py` (validation, composition for 1 and N
segments, override precedence, base-less rejection, composed marketplace.json
URL) and `tests/unit/commands/test_marketplace_check.py` (per-host resolution:
composed→GitLab+token, host-prefixed override→github, local→zero ls-remote).
`uv run pytest tests/unit tests/test_console.py` is green except one pre-existing,
environment-dependent `test_auth_scoping` failure unrelated to this change.
`ruff check` / `ruff format --check` clean.

(`src/apm_cli/marketplace/` is not a Mode-B critical path and
`CONFORMANCE.{json,md}` regenerate with no diff.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `apm.yml` marketplace blocks accept an optional `sourceBase` field: a git base
+  (`https://host/path`, arbitrary depth) that host-less relative package `source`
+  values compose onto, enabling deeply nested enterprise GitLab group paths that
+  the `host.tld/owner/repo` shorthand cannot express. Per-entry host-prefixed and
+  full-URL sources override the base; local and base-less configs are unchanged.
+  (by @leocamello, #1519)
+
 ## [0.18.0] - 2026-06-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   full-URL sources override the base; local and base-less configs are unchanged.
   (by @leocamello, #1519)
 
+### Fixed
+
+- `apm marketplace check` now resolves each package against its effective host
+  with that host's token (matching `apm pack`), instead of always probing the
+  default host with no token. Entries on a non-default host -- a
+  `host.tld/owner/repo` shorthand or a relative source composed onto
+  `marketplace.sourceBase` -- now validate correctly, and local `./` packages
+  skip the network entirely. (by @leocamello, #1519)
+
 ## [0.18.0] - 2026-06-04
 
 ### Added
--- a/docs/src/content/docs/producer/publish-to-a-marketplace.md
+++ b/docs/src/content/docs/producer/publish-to-a-marketplace.md
@@ -117,6 +117,33 @@ compiled `marketplace.json` -- that rename is the only structural
 transform `apm pack` performs. Strict schema: unknown keys raise an
 error, never silently ignored.
 
+### Deeply nested hosts: `sourceBase`
+
+If your packages live on a self-managed host whose group nesting is
+deeper than the `host.tld/owner/repo` shorthand can express (a common
+self-managed GitLab shape), declare a `sourceBase` once and write each
+`source` relative to it:
+
+```yaml
+marketplace:
+  owner:
+    name: acme-org
+  sourceBase: https://gitlab.acme.example.com/group/sub-group/team/projects
+  packages:
+    - name: my-package
+      source: my-package                  # -> .../team/projects/my-package
+      version: "^1.0.0"
+    - name: from-github
+      source: github.com/acme-org/helper  # host-prefixed: overrides the base
+      ref: v1.0.0
+```
+
+A host-less `source` composes onto `sourceBase`; a `source` that names
+its own host (the `host.tld/owner/repo` or full `https://` URL form) or
+a local `./` path overrides it and the base is ignored. Marketplaces
+without a `sourceBase` resolve exactly as before. See the
+[manifest schema](../reference/manifest-schema/) for the full rules.
+
 Add and edit packages without leaving the shell:
 
 ```bash
--- a/docs/src/content/docs/reference/manifest-schema.md
+++ b/docs/src/content/docs/reference/manifest-schema.md
@@ -757,6 +757,7 @@ Overrides exist for the rare case where the published marketplace identity diffe
 | `output` | `string` | OPTIONAL | `.claude-plugin/marketplace.json` | Output path for the generated marketplace JSON. |
 | `metadata` | `object` | OPTIONAL | `{}` | Free-form metadata forwarded verbatim to `marketplace.json` (e.g. `homepage`, `support`). |
 | `build` | `Build` | OPTIONAL | `tagPattern: "v{version}"` | Build configuration for resolving package refs. See Section 7.4. |
+| `sourceBase` | `string` | OPTIONAL | -- | Git base (`https://<host.tld>/<path>`, arbitrary depth) that host-less relative package `source` values compose onto. See Section 7.5. |
 | `packages` | `list<Package>` | OPTIONAL | `[]` | Packages exposed in the marketplace. See Section 7.5. |
 
 Unknown keys inside `marketplace:` are rejected at parse time.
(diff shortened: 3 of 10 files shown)


leocamello:
Closing this — #1736 landed the ratified `sourceBase` design and supersedes the authoring side of this draft. Thanks @danielmeppiel for carrying it to merge.

The one piece not covered by #1736 is the `apm marketplace check` routing gap I flagged there: a `sourceBase` (or #1288 host-prefixed) entry still resolves against github.com in `check` and fails with exit 128. I've opened #1763 (issue #1762) with a focused fix (adapted to the merged build-time composition) and regression tests, so that follow-up carries it rather than this branch.

PR #1754, opened 2026-06-11
fix(docs): correct installation redirect to avoid circular reference

Ikalus1988:
## Background

The Quickstart page links to "Installation" for alternative install methods (Homebrew, Scoop, pip, air-gapped):
`/getting-started/installation` redirects to `/apm/quickstart`, creating a self-referencing loop — users who click the link land back on the same Quickstart page with no installation content.

## Fix

Change the redirect destination from `/apm/quickstart` to `/apm/getting-started/installation/`. The `getting-started/installation.md` content already exists and contains all alternative installation methods. This matches the existing pattern used by other restored redirects in the same block.

See [getting-started/installation.md](https://github.com/microsoft/apm/blob/main/docs/src/content/docs/getting-started/installation.md) for the existing content.

> This redirect issue has a pattern similar to documentation reorganization challenges we encountered in MisakaNet (a federated knowledge network project) — where legacy URL paths pointing to restructured content needed careful remapping rather than circular fallbacks. The same approach of verifying the target page actually exists before setting the redirect applies here.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -36,7 +36,7 @@ export default defineConfig({
 		'/introduction/anatomy-of-an-apm-package': '/apm/concepts/package-anatomy',
 		// Legacy getting-started -> persona ramps
 		'/getting-started/quick-start': '/apm/quickstart',
-		'/getting-started/installation': '/apm/quickstart',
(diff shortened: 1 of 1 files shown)


danielmeppiel:
already fixed, thanks! we will be deploying the fix in a few mins

PR #1581, opened 2026-06-02
feat(marketplace): inherit description/version from local-path apm.yml

imbabamba:
## TL;DR

`apm pack` now auto-fills `description` and `version` in
`marketplace.json` from each local-path package's own `apm.yml` --
the same fallback the remote source path has used since #1061.
A curator-side value under `marketplace.packages` still wins when
set, and the local read is path-traversal-guarded against the
project root.  Validated against the full unit suite (15837 tests)
with 13 new tests covering the local fallback path.

## Problem (WHY)

The remote branch in `ClaudeMarketplaceMapper.compose` already reads
`description` and `version` from each resolved package's own
`apm.yml` when the curator entry omits them; `_fetch_remote_metadata`
populates that lookup over HTTPS.

The local branch had no such fallback.  It emitted only what the
curator wrote in root `apm.yml`'s `marketplace.packages[]`.  A stale
comment in `_prefetch_metadata` even claimed "Local-path packages are
skipped (they carry their own metadata)" -- but the local code path
never read any metadata at all.

For a monorepo marketplace using `source: ./plugins/<name>` entries,
that gap leaves producers with two options:

- Duplicate the description between root `apm.yml`'s
  `marketplace.packages[]` and each `plugins/<name>/apm.yml`.  The
  two copies drift over time -- the producer edits one and forgets
  the other.
- Run custom sync tooling outside `apm pack` to keep them aligned.

Remote sources do not have this gap.  This PR closes it for local
sources by mirroring the existing remote behavior exactly.

## Approach (WHAT)

| Layer | Before | After |
|---|---|---|
| `builder._prefetch_metadata` | Filtered out local packages.  Returned `{}` under `--offline`. | Reads local packages from disk first (always), then remote concurrently (still skipped under `--offline`).  Returns one dict the mapper consumes the same way regardless of source kind. |
| `builder._fetch_local_metadata` | Did not exist. | New method.  Reads `<project_root>/<subdir>/apm.yml`, extracts `description` and `version`.  Path-traversal-guarded via `ensure_path_within`.  Skips a source that resolves to the project root itself. |
| `output_mappers.ClaudeMarketplaceMapper.compose` (`is_local` branch) | Emitted curator-side fields only.  No fallback. | Same hierarchy as the remote branch: curator wins; otherwise fall back to package-manifest meta.  Verbose diagnostic on divergence. |
[...]

PR #1755, opened 2026-06-11
fix(marketplace): inherit description/version from local-path apm.yml

### LINKED ISSUE #1725: [BUG] marketplace.json omits description/version for local virtual-subdirectory packages (only remote packages are enriched)
**Describe the bug**

For a monorepo that is **its own marketplace** and references its packages by **local virtual-subdirectory** source (`source: ./packages/<name>`), `apm pack` produces a `marketplace.json` whose plugin entries have **no `description` and no `version`** — even though every package's own `apm.yml` declares both. `apm marketplace browse <name>` therefore renders `--` in the **Description** and **Version** columns for every plugin.

The metadata enrichment that fills these fields runs for **remote** packages only: the builder fetches `description`/`version` from a package's `apm.yml` over the network, but for a local-path package it never reads the `apm.yml` sitting on disk next to the manifest. The Claude output mapper has no local fallback, so unless the curator **duplicates** `description`/`version` onto each marketplace entry by hand, local entries are emitted with only `name`/`source`/`tags`. Note the asymmetry: `tags` declared on the entry pass through fine, so the manifest is valid — only the package-owned `description`/`version` are silently lost.

Verified against apm-cli 0.19.0 by reading the source and with live `apm pack` runs.

**To Reproduce**

1. A monorepo marketplace whose entries reference packages by local subpath, with the per-package metadata owned by each package's manifest:

   ```yaml
   # apm.yml (repo root)
   marketplace:
     owner: { name: tester, url: https://github.com/tester }
     outputs: { claude: {} }
     packages:
       - name: foo
         source: ./packages/foo
         tags: [demo]
   ```

   ```yaml
   # packages/foo/apm.yml
   name: foo
   version: 1.2.3
   description: A local foo package with real metadata.
   type: skill
   target: all
   ```

2. Run `apm pack --offline`.
3. Inspect `.claude-plugin/marketplace.json`.

Observed — `description` and `version` are dropped, even though `packages/foo/apm.yml` declares them:

```json
{
  "name": "foo",
  "tags": ["demo"],
  "source": "./packages/foo"
}
```

Duplicating the same fields onto the marketplace **entry** makes them appear, which confirms the serialization path itself works — only the local-source-of-truth read is missing:

```yaml
    - name: foo
      source: ./packages/foo
      tags: [demo]
      description: A local foo package with real metadata.
      version: 1.2.3
```

```json
{
  "name": "foo",
  "description": "A local foo package with real metadata.",
  "version": "1.2.3",
  "tags": ["demo"],
  "source": "./packages/foo"
}
```

A real-world instance: a 96-package monorepo marketplace, all local subpath sources, every package `apm.yml` carrying `version` + `description` → all 96 browse entries show empty Description/Version.

**Expected behavior**

For a local-path (virtual-subdirectory) package, the builder should read `<source>/apm.yml` from disk and inherit `description`/`version` (and the other Anthropic pass-through metadata) the same way it fetches them from the remote `apm.yml` for remote packages — a local-filesystem mirror of the existing remote-metadata fetch. Curator-supplied values on the marketplace entry should still take precedence. This matches the intent already stated in the code (`_prefetch_metadata` documents that *"Local-path packages are skipped (they carry their own metadata)"*) — but that local metadata is never actually read, so the field ends up empty instead of inherited.

 - OS: Linux (WSL2)
 - Python Version: 3.14
 - APM Version: 0.19.0
 - VSCode Version (if relevant): n/a

**Logs**

`apm marketplace browse` against such a marketplace (Description / Version empty for every plugin):

```
                         Plugins in 'example-marketplace'
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Plugin                       ┃ Description ┃ Version ┃ Install               ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━┩
│ foo                          │ --          │   --    │ foo@example-marketpl… │
└──────────────────────────────┴─────────────┴─────────┴───────────────────────┘
```

`apm pack --offline` reports success while emitting the metadata-less entries:

```
[*] Built marketplace.json [claude] (1 package(s)) -> .claude-plugin/marketplace.json
```

**Additional context**

Source-level root cause (apm-cli 0.19.0):

1. `marketplace/builder.py` (`_resolve_entry`) resolves a local entry with `source_repo=""` (the `if entry.is_local:` branch). The `source_repo` field is what later marks a package as "remote".

2. `marketplace/builder.py` (`_prefetch_metadata`) builds the `{name: {description, version}}` map but first filters to remote entries only:
   ```python
   # Local-path packages are skipped (they carry their own metadata).
   remote = [pkg for pkg in resolved if pkg.source_repo]
   ```
   Because local packages have `source_repo == ""`, they are excluded from enrichment entirely. The comment asserts local packages "carry their own metadata", but nothing downstream reads it.

3. `marketplace/builder.py` (`_fetch_remote_metadata`) is documented as *"fetch `description` and `version` from the package's **remote** `apm.yml`"* and reaches over the network (GitHub host / token aware). There is no local-filesystem counterpart that reads `<source>/apm.yml`.

4. `marketplace/output_mappers.py` (`ClaudeMarketplaceMapper`, `uses_remote_metadata = True`) emits `description`/`version` differently for local vs remote:
   ```python
   if is_local:
       if entry.description:
           plugin["description"] = entry.description
       if entry.version:
           plugin["version"] = entry.version
   else:
       meta = remote_metadata.get(pkg.name, {})
       ...
       elif meta.get("description"):
           plugin["description"] = meta["description"]
       ...
       elif meta.get("version"):
           plugin["version"] = meta["version"]
   ```
   The remote branch falls back to the fetched `remote_metadata`; the local branch only emits curator-supplied `entry.description`/`entry.version` and has **no fallback** to the package's on-disk `apm.yml`. (The `CodexMarketplaceMapper` is unaffected — that format carries `category`, not `description`/`version`.)

Net: for a monorepo-as-marketplace using local subpath sources, every plugin's `apm.yml` (with `version` + `description`) is on disk right beside its manifest, but the builder never reads it, so the generated `marketplace.json` omits both fields.

Workaround in use: a CI step discovers every `packages/*/apm.yml` and writes `description`/`version` onto each marketplace entry in the root `apm.yml` before `apm pack`, so the fields are present on the entry by the time the Claude mapper runs.

danielmeppiel:
## TL;DR

`apm pack` now auto-fills `description` and `version` in `marketplace.json` from each local-path package's own `apm.yml` -- the same fallback the remote source path has used since #1061. A curator-side value under `marketplace.packages` still wins when set, and the local read is path-traversal-guarded against the project root.

Closes #1725.

## Problem (WHY)

The remote branch in `ClaudeMarketplaceMapper.compose` already reads `description`/`version` from each resolved package's own `apm.yml` when the curator entry omits them (`_fetch_remote_metadata`, over HTTPS). The local branch had no such fallback -- it emitted only what the curator wrote in the root `apm.yml`'s `marketplace.packages[]`, and a stale comment in `_prefetch_metadata` even claimed local packages "carry their own metadata" while never reading any.

For a monorepo marketplace using `source: ./plugins/<name>` entries, this left producers duplicating the description between the root manifest and each package manifest (which drift), or running custom sync tooling. This PR closes the gap for local sources by mirroring the existing remote behavior exactly.

## Approach (WHAT)

| Layer | Before | After |
|---|---|---|
| `builder._prefetch_metadata` | Filtered out local packages; returned `{}` under `--offline`. | Reads local packages from disk first (always), then remote concurrently (still skipped under `--offline`). Returns one dict the mapper consumes the same way regardless of source kind. |
| `builder._fetch_local_metadata` | Did not exist. | New method. Reads `<project_root>/<subdir>/apm.yml`, extracts `description`/`version`. Path-traversal-guarded via `ensure_path_within`. Skips a source that resolves to the project root itself. |
| `output_mappers.compose` | Local + remote branches each had bespoke precedence logic. | Both branches route through one `_apply_field_with_precedence` helper: curator wins, else fall back to manifest metadata, with a verbose divergence diagnostic. |
| `docs/.../manifest-schema.md` | `description`/`version` rows had no mention of the fallback. | Rows updated + a one-paragraph note covering the fallback for both source kinds. |

## Edge cases handled

- Missing per-package `apm.yml` -> `None`; no key emitted.
- Malformed YAML -> `None`; logged at debug; build continues.
- Source escapes project root -> `ensure_path_within` raises; caught, returns `None`.
- Source resolves to the project root itself -> explicit skip so the marketplace's own `apm.yml` is never read as a package manifest.
- Empty `subdir` -> defensive early return.
- `--offline` -> local reads still run (filesystem); remote reads skipped.

## Validation evidence

```
$ uv run --extra dev ruff check src/ tests/            # All checks passed!
$ uv run --extra dev ruff format --check src/ tests/   # already formatted
$ uv run --extra dev pytest tests/unit/marketplace -q  # 1551 passed
    --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/   # 10.00/10
$ bash scripts/lint-auth-signals.sh                    # auth-signal lint clean
```

19 new tests: 8 unit tests for `_fetch_local_metadata` and 6 compose-level inheritance/precedence tests (plus the existing suite re-greened for the unified override-summary wording).

## How to test

1. Create a marketplace `apm.yml` with a `source: ./packages/<name>` entry that omits `description`/`version`.
2. Give `packages/<name>/apm.yml` a `description` and `version`.
3. Run `apm pack` and confirm the generated `marketplace.json` plugin entry carries both fields.

---

### Provenance

This re-implements the fix originally proposed in #1581 by @imbabamba, which could not be merged because the CLA was not signed. Thanks to @srobroek for the original report (#1725). Opened as an equivalent maintainer-authored change so the fix can land; #1581 is being closed in favor of this PR.


danielmeppiel:

> Local-path marketplace packages now inherit description and version from their own apm.yml, reducing monorepo authoring boilerplate while preserving curator precedence and path safety.

The panel converged on ship_now after the shepherd pass folded all in-scope Copilot and panel follow-ups. The implementation now normalizes the project-root comparison through the same containment helper as package roots, validates the package apm.yml file path before reading, caps local metadata reads, and keeps diagnostics/doc wording aligned with the actual field-count and remote-host semantics. Test coverage and CI both confirm the final state.

**Aligned with:** portable_by_manifest: package apm.yml remains the source of portable package identity; secure_by_default: local reads stay inside the project root and symlink escapes are skipped; pragmatic_as_npm: monorepo authors get DRY metadata without extra config.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 0 | Clean, bounded feature addition with shared precedence logic and guarded local metadata reads. |
| CLI Logging Expert | 0 | 0 | 0 | Diagnostic summary now describes field overrides truthfully and remains verbose-level. |
| DevX UX Expert | 0 | 0 | 0 | Local inheritance, curator precedence, and docs now match package-manager expectations. |
| Supply Chain Security Expert | 0 | 0 | 0 | Path traversal, project-root, oversized-file, and symlink escape cases are handled as cosmetic skips. |
| OSS Growth Hacker | 0 | 0 | 0 | Strong monorepo authoring DX improvement with changelog/docs coverage. |
| Doc Writer | 0 | 0 | 0 | Manifest docs now describe display-version and GitHub-class remote fallback accurately. |
| Test Coverage Expert | 0 | 0 | 0 | Regression traps cover local reads, project-root normalization, symlink escape, compose inheritance, precedence, and summary wording. |

> B = high-signal correctness/security findings, R = recommended follow-ups, N = nits.

### Recommendation

Ready for maintainer review. No deferred follow-ups remain from this shepherd pass.

### Folded in this run

- (copilot) Normalize the local metadata project-root equality check with the same path guard used for package roots -- resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879.
- (copilot) Make the override summary describe curator-supplied field overrides instead of package entries -- resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879.
- (copilot) Update the logging test assertion to match the corrected override-summary semantics -- resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879.
- (panel) Clarify manifest docs for remote display-version fallback and semver ranges -- resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879.
- (panel) Cap local apm.yml metadata reads before yaml.safe_load -- resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879.
- (panel) Re-validate the package apm.yml path so symlinks cannot escape the project root -- resolved in abc25f683b43ad78c7ec11244b33ca07f55b78b9.
- (panel) Scope the docs claim to GitHub-class remote metadata fallback -- resolved in abc25f683b43ad78c7ec11244b33ca07f55b78b9.

### Copilot signals reviewed

- `src/apm_cli/marketplace/builder.py` -- LEGIT: project-root comparison used a different path shape on Windows extended-prefix paths (resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879).
- `src/apm_cli/marketplace/output_mappers.py` -- LEGIT: override_count tracked fields while summary said entries (resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879).
- `tests/unit/marketplace/test_builder_logging.py` -- LEGIT: logging assertion needed to track the corrected summary wording (resolved in 0f28baeafac5905d41a0d7c26bf1c8322a780879).

- `tests/unit/marketplace/test_builder.py::TestFetchLocalMetadata::test_project_root_skip_uses_normalized_root` -- restored the old `self._project_root.resolve()` comparison; test FAILED as expected; guard restored.
- `tests/unit/marketplace/test_builder.py::TestFetchLocalMetadata::test_apm_yml_symlink_escape_returns_none` -- replaced file-path containment with a direct file read; test FAILED as expected; guard restored.

### Lint contract

`uv run --extra dev ruff check src/ tests/`, `uv run --extra dev ruff format --check src/ tests/`, pylint R0801, and `bash scripts/lint-auth-signals.sh` passed locally before the final push.

### CI

`gh pr checks 1755 --repo microsoft/apm` reports all checks successful: 14 successful, 1 skipped, 0 pending, 0 failing (after 0 CI fix iterations).

### Mergeability status

Captured from `gh pr view 1755 --json mergeable,mergeStateStatus,statusCheckRollup` immediately after the last push of this run.

| #1755 | `abc25f6` | ship_now | 2 | 7 | 0 | 2 | green | MERGEABLE | BLOCKED | pending required review |

### Convergence

2 outer iteration(s); 2 Copilot round(s). Final panel verdict: `ship_now`.

Ready for maintainer review.

---

<details>

Final pass findings were folded or determined to be polish outside the CEO follow-up set. No in-scope follow-up remains open.

</details>

PR #1812, opened 2026-06-16
fix: correct stale tar.gz/tarball refs and publish UX gaps after zip default

nadav-y:
## Summary

Follow-up to #1779 (the zip-default cleanup). Fixes five issues surfaced by the post-merge panel review, plus completes the stale-string sweep the first pass missed.

- **`--package` missing from all examples** (`_PUBLISH_HELP` in publish.py and Custom layouts in package-authoring.md): `--package` is `required=True` but every example omitted it -- copy-paste crash. Three panelists flagged this.
- **Four stale `registry tarball for` strings in resolver.py** (lines 98, 104, 363, 369): same class of fix as extractor.py in the original PR, missed in the first pass. User-visible on every package validation failure.
- **Em dash in `Forbidden` ClickException** (publish.py line 272): user-visible on 403 responses; prior commit fixed logger.info() calls in the same file but not this one.
- **Missing zip-build guidance in Custom layouts** (registries.md): removed the tar czf example without adding a zip equivalent; adds a cross-platform `python -m zipfile` step.
- **No `--tarball` migration note** (publish.md `--zip` row): CI pipelines using `apm publish --tarball` get `No such option: --tarball` with no recovery path.

## What was omitted and why

- `--package` optional (infer from `apm.yml name:`): design change, tracked as a separate issue per panel recommendation.
- Pre-existing em dashes and section signs in internal comments (resolver.py, publish.py): outside this PR's scope; those are in non-user-visible code comments predating the cleanup.
- `test_mismatch_raises` adding `match=` (nit): behavioral coverage confirmed; message-text trap deferred.

## Test plan

- [ ] `uv run pytest tests/unit/registry/` passes (213 passed, 2 xfailed)
- [ ] `ruff check` and `ruff format --check` pass on changed files
- [ ] `apm publish --help` shows `--package` in all four examples
- [ ] `apm publish` (no args) gives `Missing option '--package'` (not a crash from a different missing option)

Generated with [Claude Code](https://claude.ai/claude-code)


nadav-y:
Superseded by the squash push to PR #1779.

PR #1901, opened 2026-06-25
fix(install): persist --target into auto-bootstrapped apm.yml (#1743)

### LINKED ISSUE #1743: Suggestion: persist install-time --target in apm.yml so bare apm update works
### Summary

When we run `apm install <package> --target <harnesses>`, the chosen targets are not written into the auto-generated `apm.yml`. Because of that, `apm update` currently asks us to pass the same `--target` flag again.

It would be more convenient if `apm update` could simply refresh what was already installed, without having to remember and re-type the target list from the initial install.

### Motivation

- **On install**, specifying `--target` feels natural—we are choosing which agent harnesses should receive primitives.
- **On update**, we usually just want to pull in newer package versions for the same setup. Re-specifying `--target` each time adds a small but recurring step, and it is easy to mismatch the value used at install time.

APM already documents a target resolution chain (`--target` > `apm.yml` `targets:` > auto-detect > fallback). It looks like the missing piece is persisting `targets:` into `apm.yml` when `--target` is used during `apm install`.

Docs reference: [Install packages — Where files land](https://microsoft.github.io/apm/consumer/install-packages/)

### Current behavior

1. Run install with an explicit target:

       apm install example.com/org/my-package#main --target cursor,copilot,claude

2. `apm.yml` is auto-created with dependencies, but without a `targets:` field.

3. Later, run update without `--target`:

       apm update

4. Update may fail or ask for target disambiguation (e.g. `Multiple harnesses detected`), so we need to repeat the same `--target` value from install.

### Suggested behavior

**On `apm install`:** when `--target` is passed, persist the resolved target list into `apm.yml`:

       targets:
         - cursor
         - copilot
         - claude

This would be a one-time manifest update at install time, alongside the existing dependency entries.

**On `apm update`:** read `targets:` from `apm.yml` and deploy to those harnesses. `apm update` would not need to modify `apm.yml`—only `apm.lock.yaml` (and deployed primitives) would change, which matches how we think about update vs install.

After that, bare `apm update` would be enough for the common refresh workflow.

`--target` on update could still override `apm.yml` for one-off runs, consistent with the documented resolution priority.

### Scenario overview

| Scenario | Suggested behavior |
| --- | --- |
| `apm install <pkg> --target a,b` | Write `targets:` into `apm.yml` |
| `apm update` when `targets:` is already in `apm.yml` | Use persisted targets; no `--target` flag needed |
| `apm update` | Update `apm.lock.yaml` and redeploy; leave `apm.yml` unchanged |
| `apm update --target x` | Override persisted targets for that run only |

### Workaround (today)

We can manually add `targets:` to `apm.yml` after the first install, or start from a template manifest that already includes `targets:`. That works because `apm.yml` `targets:` is already honored by the resolution chain—it is just not set automatically by the recommended one-liner install flow.

### Environment

- APM CLI: `0.19.0`
- OS: macOS / Windows
- Use case: installing skills/rules into multiple harnesses (`cursor`, `copilot`, `claude`) from a private Git host, then periodically running `apm update` in the same project directory

### References

- Target resolution priority: `apm update --help` (`--target` > `apm.yml` `targets:` > auto-detect)
- Consumer docs: https://microsoft.github.io/apm/consumer/install-packages/
- Update docs: https://microsoft.github.io/apm/consumer/update-and-refresh/

danielmeppiel:
## TL;DR

`apm install <pkg> --target X` in a directory with no `apm.yml` now persists the selected harness(es) into the auto-created manifest's `targets:` field, so a later bare `apm update` redeploys to the same targets without re-specifying `--target`. Closes #1743.

## Problem (WHY)

When `apm install` auto-bootstraps a manifest (no `apm.yml` present, packages given), it built the minimal config via `_get_default_config(project_name)` — which returns only `name`/`version`/`description`/`author`. The `--target` CLI value was used for the one-shot deploy but **never written to disk**. The next bare `apm update` had no `targets:` to read and re-prompted / re-required `--target`.

Note: the resolution chain (`apm.yml targets:`) and `apm init`'s target persistence already worked — only this install-time auto-bootstrap path dropped the selection.

## Approach (WHAT)

Thread the resolved `--target` into the config dict before `_create_minimal_apm_yml`, exactly mirroring how `apm init` does it (`init.py`: `config["targets"] = sorted(...)`). `_create_minimal_apm_yml` already supports writing `targets:` from config, so the change is a single populate step at the call site — no new writer logic.

```python
config = _get_default_config(project_name)
if target is not None:
    config["targets"] = sorted([target] if isinstance(target, str) else list(target))
_create_minimal_apm_yml(config, target_path=manifest_path)
```

`target` is already normalized by `TargetParamType` to `str | list[str] | None`, and the persisted plural list round-trips through `parse_targets_field` — the same reader `apm update` uses.

## Implementation (HOW)

- `src/apm_cli/commands/install.py`: populate `config["targets"]` from the resolved `--target` before the auto-bootstrap write.
- `tests/unit/test_install_command.py`: two regression tests — (1) `--target claude,cursor` persists `targets: [claude, cursor]` and round-trips through `parse_targets_field`; (2) no `--target` flag still omits `targets:` (preserves auto-detect fallthrough).
- Docs: `reference/cli/install.md` auto-bootstrap note + `apm-usage/commands.md` `--target` clause.

## Trade-offs

- Mirrors `apm init` behavior verbatim for consistency, including its existing plural-form normalization. Out of scope: the pre-existing latent quirk where a multi-target list containing `copilot` normalizes to `vscode` (and `all` stays literal) — that is shared with `apm init` and tracked separately; this PR does not widen or fix it.

## Validation evidence

- New tests fail before the fix (red: `targets` is `None`), pass after (green).
- `tests/unit/test_install_command.py` + `tests/unit/commands/`: 1193 passed.
- CI-mirror lint chain clean: `ruff check`, `ruff format --check`, `pylint R0801` (10.00/10), `lint-auth-signals.sh`.

## How to test

```bash
cd "$(mktemp -d)"
apm install some-org/some-pkg --target claude,cursor
grep -A2 '^targets:' apm.yml   # now lists claude + cursor
apm update                     # reuses persisted targets, no --target needed
```


danielmeppiel:

> Clean ship: `install --target` now persists recognized targets into an auto-bootstrapped `apm.yml`, and the PR is converged after shepherd follow-ups.

All active panel lenses converged after the shepherd fold. The implementation centralizes manifest-safe target normalization in `core.target_detection`, keeps `install.py` under the CI file-length guardrail, writes only values that round-trip through `parse_targets_field`, and confirms persistence in CLI output. The docs now scope the behavior to recognized manifest targets, and coverage spans helper unit tests, install command regression tests, and an integration-with-fixtures bootstrap test.

**Aligned with:** portable by manifest: target intent is written to `apm.yml`; pragmatic as npm: users do not need to repeat `--target` on the next update; multi-harness support: alias and `all` expansion paths persist manifest-safe target names.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 0 | Helper extraction keeps target normalization in the target-detection layer and `install.py` below the guardrail. |
| CLI Logging Expert | 0 | 0 | 0 | Target persistence is reported through CommandLogger with exact target names. |
| DevX UX Expert | 0 | 0 | 0 | Bootstrap now confirms durable target persistence while preserving no-target auto-detect. |
| Supply Chain Security Expert | 0 | 0 | 0 | No supply-chain surface added; only allowlisted manifest target names are written. |
| OSS Growth Hacker | 0 | 0 | 0 | The PR removes second-run friction for first-time target-specific installs. |
| Doc Writer | 0 | 0 | 0 | Docs accurately scope persistence to recognized manifest targets. |
| Test Coverage Expert | 0 | 0 | 0 | Unit and integration evidence covers the critical install-pipeline behavior. |

### Recommendation

Ship now. Remaining gate is normal maintainer review; PR checks are green and deploy is skipped as expected for PRs.

### Folded in this run

- (panel) Persist only manifest-safe target names for CLI aliases and `--target all` -- resolved in `965ec75`.
- (panel) Add user-visible confirmation that bootstrap target persistence happened -- resolved in `965ec75`.
- (panel) Split and narrow install docs so the behavior is discoverable and accurate -- resolved in `965ec75`.
- (panel) Add integration and helper edge-case coverage for target persistence -- resolved in `965ec75`.
- (panel) Recover CI Lint file-length guardrail by moving target normalization out of `install.py` -- resolved in `965ec75`.

### Copilot signals reviewed

No Copilot inline comments were present in this run.

- `tests/unit/test_install_command.py::TestInstallCommandAutoBootstrap::test_install_no_apm_yml_persists_target_flag` -- deleted `manifest_targets_from_target_option` behavior by returning `None`; test FAILED as expected; guard restored.
- `tests/integration/test_wave7_marketplace_install_coverage.py::TestInstallAutoBootstrap::test_no_apm_yml_with_target_flag_persists_targets` -- deleted `manifest_targets_from_target_option` behavior by returning `None`; test FAILED as expected; guard restored.

### Lint contract

`uv run --extra dev ruff check src/ tests/` and `uv run --extra dev ruff format --check src/ tests/` both silent. Pylint R0801 reported 10.00/10, file-length guardrail passes with `install.py` at 2098 lines, and `scripts/lint-auth-signals.sh` is clean.

### CI

All PR checks pass on `965ec7523b64aa775020c2f48ab0cbb0708ce7c0`; `deploy` is skipped as expected for PRs (after 1 CI fix iteration).

### Mergeability status

Captured from `gh pr view 1901 --json mergeable,mergeStateStatus,statusCheckRollup` immediately after the last push of this run.

| #1901 | `965ec75` | ship_now | 2 | 5 | 0 | 1 | green | MERGEABLE | BLOCKED | pending required review |

### Convergence

2 outer iteration(s); 1 Copilot round(s). Final panel verdict: `ship_now`.

Ready for maintainer review.

---

<details>

No active persona has remaining findings. Auth and performance personas were inactive because the final diff does not touch auth, credential, transport, cache, materialization, or performance-sensitive paths.

</details>

PR #1958, opened 2026-06-29
feat(install): allow dynamic override of package versions via env vars or version.env

yohi:
This PR introduces a mechanism to dynamically override package versions via environment variables prefixed with `APM_` (e.g. `APM_OH_MY_OPENAGENT=v2.0.0`) or by placing them inside a local `version.env` file.

### Changes
- Add `_load_version_env()` to `apm_cli/cli.py` to parse environment variables from `version.env` if it exists in the current directory on start.
- Add a `__post_init__` method to `DependencyReference` to override `self.reference` when a matching environment variable `APM_<LIBRARY_NAME>` is found.
- Add unit tests for both environment variable override and `version.env` parsing.

### DIFF
--- a/src/apm_cli/cli.py
+++ b/src/apm_cli/cli.py
@@ -329,8 +329,31 @@ def _configure_encoding() -> None:
             _warn_encoding_issue(current_cp)
 
 
+def _load_version_env() -> None:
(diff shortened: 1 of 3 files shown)


yohi:
mistake

PR #2015, opened 2026-07-04
feat(copilot): support local MCP config in .mcp.json and global via -g flag

Skolim007:
feat(copilot): support local MCP config in .mcp.json and global via -g flag

## TL;DR

This PR resolves outdated MCP configuration handling for GitHub Copilot by aligning with the official migration guide. It introduces support for project-local `.mcp.json` configurations and scopes the previous default to a global context accessible via the `-g` flag.

## Problem (WHY)

- The GitHub Copilot CLI adapter did not support the newly recommended `.mcp.json` local configuration file.
- We must provide deterministic behavior across environments. As noted in Agent Skills: ["Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/).
- Configuration needed to correctly separate project-local settings from user-global settings.

## Approach (WHAT)

- Update `CopilotClientAdapter` path resolution to return `.mcp.json` for the project scope and `~/.copilot/mcp-config.json` for the user scope.
- Propagate the scope (`InstallScope.PROJECT` vs `InstallScope.USER`) through the `MCPIntegrator`.
- Update relevant docs and add test coverage for path resolution.

## Implementation (HOW)

| File | Change |
| --- | --- |
| `src/apm_cli/adapters/client/copilot.py` | Overhauled configuration pathing based on `user_scope` boolean flag. |
| `src/apm_cli/integration/mcp_integrator.py` | Added scope propagation. |
| `CHANGELOG.md` & `docs/**/*.md` | Documented the new `-g` flag and local `.mcp.json` standard. |
| `tests/**/*.py` | Validated that both scopes return correct paths and do not overwrite each other. |

## Diagrams

The updated configuration resolution path:

```mermaid
flowchart LR
    A["CopilotClientAdapter"] -->|Project Scope| B[".mcp.json"]
    A -->|User Scope (-g)| C["~/.copilot/mcp-config.json"]
```

## Trade-offs

- **Breaking Change for Default Behavior**: Users who relied on the previous behavior for global configuration must now explicitly pass the `-g` flag if they are outside a project directory. 

## Benefits

1. Enables native, standard-compliant project-level MCP configurations.
2. Directly aligns with the VS Code `mcp.json` migration guide.
3. Cleanly separates workspace config from user config, preventing accidental bleed.

## Validation

<details><summary>Test Output</summary>

```
============================= test session starts =============================
platform win32 -- Python 3.12.3, pytest-9.0.3, pluggy-1.6.0
collected 378 items

tests\unit\test_copilot_adapter.py ..................................... [  9%]
..................                                                       [ 14%]
tests\unit\test_copilot_adapter_compatibility.py ....................... [ 20%]
........................................................................ [ 39%]
.................                                                        [ 44%]
tests\unit\test_copilot_adapter_phase3.py .............................. [ 52%]
........................................................................ [ 71%]
..........                                                               [ 73%]
tests\unit\integration\test_mcp_integrator.py .......................... [ 80%]
........................................................................ [ 99%]
.                                                                        [100%]

============================ 378 passed in 15.91s =============================
```

</details>

## How to test

- [ ] Run `apm install` within a project directory and verify `.mcp.json` is updated.
- [ ] Run `apm install -g` outside a project directory and verify `~/.copilot/mcp-config.json` is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `apm install --mcp NAME --target copilot` now writes project-scoped Copilot
+  CLI MCP configuration to `.mcp.json`; use `-g` / `--global` to keep writing
+  `~/.copilot/mcp-config.json`. Stale MCP cleanup now follows the same scope.
 - `apm pack` now produces a `marketplace.json` the Copilot App accepts even
   when your manifest `name` contains dots, underscores, or other non-kebab-case
   characters (e.g. `my.marketplace`): the emitted `name` field is normalized to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `apm install --mcp NAME --target copilot` now writes project-scoped Copilot
   CLI MCP configuration to `.mcp.json`; use `-g` / `--global` to keep writing
-  `~/.copilot/mcp-config.json`. Stale MCP cleanup now follows the same scope.
+  `~/.copilot/mcp-config.json`. Stale MCP cleanup now follows the same scope. (#2015)
 - `apm pack` now produces a `marketplace.json` the Copilot App accepts even
   when your manifest `name` contains dots, underscores, or other non-kebab-case
   characters (e.g. `my.marketplace`): the emitted `name` field is normalized to
--- a/docs/src/content/docs/consumer/install-mcp-servers.md
+++ b/docs/src/content/docs/consumer/install-mcp-servers.md
@@ -100,7 +100,7 @@ for the full required-vs-optional runtime config rule.
 
 | Harness | File | Scope | Format |
 |---|---|---|---|
-| GitHub Copilot CLI | `~/.copilot/mcp-config.json` | global | JSON `mcpServers` |
+| GitHub Copilot CLI | `.mcp.json` (project) or `~/.copilot/mcp-config.json` (`-g`) | both | JSON `mcpServers` |
 | VS Code (Copilot) | `.vscode/mcp.json` | project | JSON `servers` |
 | Claude Code | `.mcp.json` (project) or `~/.claude.json` (`-g`) | both | JSON `mcpServers` |
 | Cursor | `.cursor/mcp.json` | project (only if `.cursor/` exists) | JSON `mcpServers` |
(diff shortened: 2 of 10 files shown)


danielmeppiel:
## Maintainer decision: declining (NO-GO)

Thanks for the contribution, @Skolim007, and for the clean scope propagation and test work. After a multi-lens review (DevX, architecture, CEO/strategy), we're going to **decline this PR**. The reason is a single load-bearing premise that does not hold up.

### The premise this PR rests on

The change flips the default install path so `apm install --target copilot` writes MCP config to a project-local `.mcp.json`. That only works if the **standalone GitHub Copilot CLI actually reads `.mcp.json` at the project root.** It does not.

### What the official docs say

Per GitHub's own Copilot CLI documentation ([Using GitHub Copilot CLI -> Add an MCP server](https://docs.github.com/en/copilot/how-tos/copilot-cli/use-copilot-cli/overview#add-an-mcp-server)):

> "Details of your configured MCP servers are stored in the `mcp-config.json` file, which is located, by default, in the `~/.copilot` directory. This location can be changed by setting the `COPILOT_HOME` environment variable."

So the Copilot CLI reads MCP config **only** from `~/.copilot/mcp-config.json`, and the **only** supported way to relocate it is the `COPILOT_HOME` environment variable. There is no documented support for a project-root `.mcp.json`.

The two sources cited in the PR body do not establish the premise:
- The PROSE link (`awesome-ai-native/.../prose/`) is about deterministic tool execution, not MCP config discovery.
- The "VS Code `mcp.json` migration guide" applies to the **VS Code Copilot extension** (`.vscode/mcp.json`), which is a separate target with its own adapter (`src/apm_cli/adapters/client/vscode.py`). It does not govern the standalone `copilot` CLI.

### Why this is a correctness break, not a nit

Even setting the premise aside, the PR only changes the **install-write** path. The **runtime-read** path is unchanged: `src/apm_cli/runtime/copilot_runtime.py` (get_mcp_config_path / get_runtime_info) hardcodes `~/.copilot/mcp-config.json`. After this PR, the default `apm install --target copilot` would write a file that neither Copilot CLI nor APM's own runtime ever reads -- MCP servers would **silently fail to load**. That is a worse user outcome than today.

### If you want per-project MCP config

There is a real, valid need here, but the documented mechanism is `COPILOT_HOME`, not `.mcp.json`. A future PR could set/point `COPILOT_HOME` per project and teach both the install-write and runtime-read paths to agree on that location. That is a design change rather than a fold, so it belongs in a fresh proposal (ideally with a linked issue).

Happy to reopen or review a follow-up if you can cite an official GitHub Copilot CLI source showing the standalone CLI reads project-root `.mcp.json`. Absent that, we can't take this change. Thanks again for digging into this area.

Skolim007:
Thanks for the feedback! My team uses AMP and I think it is a great tool to get some level of consistency for AI-assisted coding for larger teams. You refference to docs is correct, but somehow "it works on my machine" :) When working on multiple projects, having a workspace-specific MCP setup is essencial, otherwise, global MCP gets huge and generates conflicts.

Is it possible that there is some undocumented behaviour of Copilot CLI? My `~/.copilot/mcp-config.json` has no MCP configured. When I open Copilot CLI in a project folder that has "old" `.vscode/mcp.json` but no `.mcp.json` in the project root I get the following warning message:
```
Copilot CLI's incomplete support for .vscode/mcp.json has been removed. See https://gh.io/copilotcli-mcpmigrate to migrate to .mcp.json or .github/mcp.json.
```
See the screenshot.
image

After creating `.mcp.json` as the manual suggests, Copilot CLI does not display the warning and correctly detects MCP locally (it is listed as "workspace-level" MCP). 
image
 I was also able to use the "patched" APM to get this local setup to work.

Skolim007:
PS. I found the documentation that confirms taht workspace-level MCP is supported, though I'd say it is well hidden

> MCP servers from different sources are merged in priority order (highest first). When servers share a name, the higher-priority source takes precedence.
> 
> --additional-mcp-config flag (highest)
> Plugin-provided servers
> Workspace servers—.mcp.json and .github/mcp.json loaded from the working directory upward to the Git root; requires the folder to be trusted
> ~/.copilot/mcp-config.json (lowest)

https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-command-reference#mcp-server-loading-priorityi

PR #2005, opened 2026-07-03
feat(tls): verify against the OS trust store by default (closes #2004)

fangkangmi:
## Summary

By default, `apm` now verifies HTTPS against the **operating-system trust store** (via [`truststore`](https://pypi.org/project/truststore/)) — the same source `git` and `curl` use — instead of the bundled `certifi` set. This makes `apm` work out-of-the-box behind a corporate CA or a TLS-inspecting proxy (closes #2004).

Coverage is the Python-based paths: `apm install` (in-process) and the Python `llm` child runtime spawned by `apm run` (a self-contained `.pth` bootstrap + `truststore` are installed into its venv at setup time). The frozen binary honours the system store as well, with `certifi` as a genuine fallback. An explicit `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` still wins, and `APM_DISABLE_TRUSTSTORE=1` restores the previous certifi-only behaviour. Node (Copilot) and Rust (Codex) child runtimes are not yet covered — tracked in #2034.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass
- [x] Added tests for new functionality (if applicable)

Added unit tests (`tests/unit/core/test_tls_trust.py`, every branch of `configure_tls_trust()`), an integration test that verifies against a freshly-minted private CA over a loopback HTTPS server (`tests/integration/test_tls_custom_ca.py`), and a regression test that drives the real frozen SSL runtime hook (`tests/integration/test_tls_frozen_hook.py`). The child-delivery chain is covered end-to-end: wheel/sdist content guards (`tests/integration/test_tls_wheel_content.py`), foreign-venv `.pth` injection (`tests/integration/test_tls_child_runtime.py`), and the round-2/round-3 verification suites. 69 TLS tests green.

Also validated the end-to-end effect by mimicking the enterprise condition (corporate CA present in the OS trust store but absent from certifi):

| Scenario | Result |
|---|---|
| Before (no truststore), CA in OS store only | TLS failure — reproduces the reported bug |
| After (truststore injected), CA in OS store | Success (HTTP 200) — fixed via OS store |
| Control: injected, CA **not** in OS store | TLS failure — verification remains enforced |

Not yet exercised on a real PyInstaller build; a `make build` smoke against a custom CA on Linux is recommended before release (the code degrades safely to certifi regardless).

## Hardening

[...]

PR #2022, opened 2026-07-04
fix(tls): use system trust store by default

### LINKED ISSUE #2004: [FEATURE] Verify TLS using the system trust store by default (breaks in enterprise/proxy environments)

`apm` fails TLS verification against internal Git hosts that use a corporate CA or a TLS-inspecting proxy, even when `git` and `curl` to the **same host** succeed. This makes the first-run experience in enterprise environments confusing and requires per-shell env-var setup.

What the user sees:

```
apm marketplace add https://<internal-ghe-host>/Org/repo.git --name foo

[x] Failed to register marketplace: Failed to fetch marketplace 'foo':
HTTPSConnectionPool(host='<internal-ghe-host>', port=443): Max retries exceeded with
url: /api/v3/repos/Org/repo/contents/marketplace.json?ref=main
(Caused by SSLError(SSLCertVerificationError(1,
'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
unable to get local issuer certificate (_ssl.c:1010)')))
```

`git clone` of the exact same URL succeeds, which is what makes this so confusing.

**Why it happens:** APM classifies GitHub/GHES/GitLab hosts and fetches `marketplace.json` (and package metadata) through the REST Contents API using **`requests`**, via a plain module-level session:

```python
# src/apm_cli/marketplace/client.py
_HTTP_SESSION = requests.Session()
```

A default `requests.Session()` verifies against the bundled **`certifi`** CA set (`certifi.where()`), which does not include internal/corporate root CAs or the certs injected by a TLS-inspecting proxy. The PyInstaller build bundles this set explicitly (`build/apm.spec`: `'certifi'  # CA certificate bundle for SSL verification in frozen binary`). Repository operations, by contrast, shell out to `git`, which reads the OS trust store (e.g. `/etc/ssl/certs/ca-certificates.crt`) -- so `git clone` succeeds while the `requests`-based Contents API call fails against the exact same certificate chain.

Use the OS trust store by default via [`truststore`](https://pypi.org/project/truststore/) (Python 3.10+; APM already requires `>=3.10`), e.g. call `truststore.inject_into_ssl()` once at CLI startup. This automatically picks up corporate CAs on Linux/macOS/Windows with **no user configuration**, is the modern recommended approach for exactly this problem, and makes `git` and `apm` behave consistently against the same host.

Acceptance criteria:

- A fresh install of APM, in an environment where only the **OS** trust store contains the corporate CA, can `apm marketplace add` / `apm install` from an internal HTTPS Git host **without setting any env vars**.
- `git`-based and `requests`-based operations use a consistent trust source.
- Behaviour documented for air-gapped / custom-CA setups (extend the existing [`troubleshooting/ssl-issues`](https://microsoft.github.io/apm/troubleshooting/ssl-issues/) page).

Please **do not** offer a "disable verification" flag as the primary remedy -- trusting the OS store is the secure fix.

- **`REQUESTS_CA_BUNDLE` env var (current workaround).** This already works and is documented, and APM already prints a hint pointing at it on TLS failure. But it must be re-set per shell / per container and is easy to miss:
  ```bash
  export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
  ```
  Note a sharp edge worth fixing regardless: **`SSL_CERT_FILE` alone does not repair the APM `requests` path.** `requests` pins `certifi.where()` and only consults `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` -- it ignores `SSL_CERT_FILE`. So the fix documented for the older #428 (setting only `SSL_CERT_FILE`) does **not** help the marketplace / install code path; users additionally need `REQUESTS_CA_BUNDLE`. This split between "what the OS trusts", "what `git` uses", and "what `requests` uses" is exactly the confusion `truststore` removes.

- **A `--ca-bundle` flag / `apm config set ca-bundle <path>`.** Lets users pin a custom CA file without relying on env vars. Useful as a secondary option, but still not zero-config.

- **Disabling verification.** Rejected -- insecure, and not an acceptable default or primary remedy.

**Additional context**

Related to #428, but distinct: #428 was about APM's bundled OpenSSL pointing at a non-existent python.org cert path. This issue is about internal/corporate root CAs not being present in the bundled `certifi` set **at all**, which is the common enterprise failure mode.

Environment where this was observed:

- APM CLI `0.23.1`
- Bundled Python (PyInstaller, `--onedir`)
- Linux (dev container, Ubuntu 24.04), behind a corporate CA / TLS-inspecting proxy
- `git clone` of the same URL succeeds; `apm marketplace add` / `apm install` fail

WilliamK112:
﻿## Summary

Closes #2004.

- add `truststore` as a runtime dependency so APM uses the operating system trust store by default for Python HTTPS clients
- bootstrap truststore at the CLI entry point before command modules import network libraries, while preserving explicit CA overrides (`REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, `SSL_CERT_FILE`, `SSL_CERT_DIR`)
- update the PyInstaller TLS runtime hook to prefer truststore/system roots and fall back to bundled certifi only when truststore is unavailable
- update TLS troubleshooting docs and the install TLS warning to explain OS trust-store defaults and the correct `REQUESTS_CA_BUNDLE` override path

## Validation

- `uv run --extra dev pytest tests/unit/test_tls_truststore.py tests/unit/test_ssl_cert_hook.py tests/unit/install/test_validation_tls.py -q` -> 27 passed
- `uv run --extra dev pytest tests/unit/test_cli_consistency.py tests/unit/test_install_command.py tests/unit/test_registry_client.py tests/unit/test_registry_client_http_cache.py tests/unit/test_ssl_cert_hook.py tests/unit/test_tls_truststore.py tests/unit/install/test_install_cmd_auth_rendering.py tests/unit/commands/test_install_error_handling.py tests/unit/marketplace/test_marketplace_commands.py tests/unit/marketplace/test_marketplace_resolver.py tests/unit/marketplace/test_client_git.py tests/unit/marketplace/test_marketplace_client.py tests/unit/registry/test_client.py tests/unit/registry/test_resolver_e2e_http.py tests/integration/test_install_validation_rules.py tests/integration/test_marketplace_add_generic_git.py -q` -> 496 passed, 10 subtests passed
- `uv run --extra dev ruff check src tests` -> passed
- `uv run --extra dev ruff format --check src tests` -> 1388 files already formatted
- `uv lock --check` -> passed
- `uv run apm --help` -> passed
- `npm run build --prefix docs` -> passed; internal links valid; only existing Astro/Vite warnings were reported
- `git diff --check` -> passed

Not counted as passing: `uv run --extra dev pytest tests/unit -q` exceeded a 5 minute local timeout without useful progress output, so I stopped the leftover pytest processes and used the targeted TLS/CLI/install/marketplace/registry validation above instead.

### DIFF
--- a/build/apm.spec
+++ b/build/apm.spec
@@ -187,7 +187,8 @@ hiddenimports = [
     'pathlib',
     'frontmatter',
     'requests',
-    'certifi',  # CA certificate bundle for SSL verification in frozen binary
+    'truststore',  # System trust store support for enterprise TLS roots
+    'certifi',  # Fallback CA bundle for frozen binaries
     # Rich modules (lazily imported, must be explicitly included)
     'rich',
     'rich.console',
--- a/build/hooks/runtime_hook_ssl_certs.py
+++ b/build/hooks/runtime_hook_ssl_certs.py
@@ -1,15 +1,14 @@
-# PyInstaller runtime hook -- configures SSL certificate paths for the
-# frozen binary so that HTTPS connections work on every platform without
-# requiring the user to install Python or set environment variables.
+# PyInstaller runtime hook -- configures TLS trust for the frozen binary so
+# HTTPS connections work on every platform without requiring the user to
+# install Python or set environment variables.
 #
-# Problem: PyInstaller bundles OpenSSL, but the compiled-in certificate
-# search path points at the *build machine's* Python framework directory
-# (e.g. /Library/Frameworks/Python.framework/...).  On end-user machines
-# that path rarely exists, causing SSL verification failures.
+# Problem: requests defaults to certifi, while enterprise machines typically
+# install corporate or internal PKI roots into the OS trust store used by git,
+# curl, and browsers. Frozen binaries also have brittle OpenSSL default paths.
 #
-# Solution: Point ``SSL_CERT_FILE`` at the certifi CA bundle shipped
-# inside the frozen binary.  ``requests``, ``urllib3``, and the stdlib
-# ``ssl`` module all honour this variable.
+# Solution: Use truststore first so urllib3/requests/stdlib HTTPS share the OS
+# trust store. If truststore is unavailable, fall back to the bundled certifi
+# CA bundle so frozen binaries still have a working public-web baseline.
 #
 # This hook executes before any application code so the variables are
 # visible to every subsequent import.
@@ -17,14 +16,40 @@
 import os
 import sys
 
+_CA_OVERRIDE_ENV_VARS = (
+    "REQUESTS_CA_BUNDLE",
+    "CURL_CA_BUNDLE",
+    "SSL_CERT_FILE",
+    "SSL_CERT_DIR",
+)
+
+
+def _has_explicit_ca_override() -> bool:
+    """Return True when the user explicitly selected a CA bundle/path."""
+    return any((os.environ.get(name) or "").strip() for name in _CA_OVERRIDE_ENV_VARS)
+
+
+def _inject_system_trust_store() -> bool:
+    """Inject truststore into stdlib SSL if available."""
+    try:
+        import truststore
+
+        truststore.inject_into_ssl()
+        return True
+    except Exception:
+        return False
+
 
 def _configure_ssl_certs() -> None:
-    """Set SSL_CERT_FILE to the bundled certifi CA bundle when frozen."""
+    """Configure TLS trust for frozen binaries."""
     if not getattr(sys, "frozen", False):
         return
 
     # Honour explicit user overrides -- never clobber them.
-    if os.environ.get("SSL_CERT_FILE") or os.environ.get("REQUESTS_CA_BUNDLE"):
+    if _has_explicit_ca_override():
+        return
+
+    if _inject_system_trust_store():
         return
 
     try:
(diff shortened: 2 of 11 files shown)


danielmeppiel:
Thank you for this, @WilliamK112 -- it was a genuinely strong implementation of the #2004 fix.

We ran a comparative review-panel pass across both open PRs for this issue (this one and #2005) with a CEO arbitration on which to drive to merge. The decision was to land #2005 as the base, primarily on three points: it ships a real private-CA end-to-end integration test that *proves* verification is not weakened (untrusted CA still rejected after injection), it keeps the CLI starting on a broken/incompatible `truststore` (broad `except` at startup), and it deliberately excludes `SSL_CERT_FILE`/`SSL_CERT_DIR` from the override set so the feature does not silently no-op in the frozen binary / on Linux boxes that export `SSL_CERT_FILE`.

That said, your PR was better on several axes, and we are folding those ideas into #2005 with attribution on the commits:

- The updated in-CLI `TLS verification failed` warning in `install/validation.py` (leading with "APM uses the system trust store by default").
- The `common-errors.md` troubleshooting update, so the CLI message and both docs pages tell one consistent story.
- The `env: Mapping` injectable-parameter pattern for override detection (nicer to unit-test).
- The `truststore>=0.10.0` version floor.
- The subprocess import-timing test approach (asserting injection fires before `requests` is imported).

Closing in favor of #2005. Your contribution is credited via `Co-authored-by` trailers on the folded commits. If any of the folded changes drift from what you intended, please do weigh in on #2005 -- we would value your sign-off there too. Thanks again for pushing on the enterprise TLS story.

PR #2025, opened 2026-07-04
fix(hooks): deploy self-contained hook scripts with siblings and package module type (closes #2023)

danielmeppiel:
fix(hooks): deploy hook bundles self-contained

## TL;DR

Hook installs now deploy the hook script bundle, not only the entry script, for Claude-family targets, Copilot, and Kiro. JavaScript and TypeScript bundles get a minimal module-type sidecar so consumer repos with `type: module` cannot hijack CommonJS hooks. The fold also excludes hook descriptor manifests, symlinks, and `.apm-pin` markers, and adds regression traps for bundle parity, stale cleanup, and shell-only behavior.

> [!NOTE]
> Closes #2023. Direction A (self-contained deploy) was chosen over runtime references into `apm_modules/` because installed hook configs must stay portable after clone and in user-scope installs.

apm-spec-waiver: Hook bundle copying fixes existing hook primitive deployment semantics; it adds no new OpenAPM manifest, lockfile, registry, or policy field.

## Problem (WHY)

- [x] Hook entry scripts could deploy without sibling modules, so a package that worked in its source tree failed after `apm install` when it used `require('./helper')` or `import './helper'`.
- [x] CommonJS hook scripts could run under a consumer repo's `type: module` unless APM wrote a nearer module-type boundary beside the deployed scripts.
- [x] Kiro still used the old single-file copy path, so the same hook package could work for Claude/Copilot and fail for Kiro.
- [x] Root hook descriptor JSON files and symlinks were eligible for deployment because the bundle copy loop only excluded `package.json`.
- [!] Tests proved two happy paths, but not the lockfile/stale-cleanup contract, shell-only sidecar skip, descriptor exclusion, symlink exclusion, nested subdirs, or Kiro parity.

Why these matter: the fix needs deterministic evidence because PROSE says, ["Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) The implementation stays small and shared because PROSE says, ["Favor small, chainable primitives over monolithic frameworks."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) The PR body uses a concrete scenario table because Agent Skills says, ["agents pattern-match well against concrete structures"](https://agentskills.io/skill-creation/best-practices).

## Approach (WHAT)

| # | Fix | Principle | Source |
|---|-----|-----------|--------|
[...]

PR #2028, opened 2026-07-04
fix(hooks): deploy hook script runtime context

### LINKED ISSUE #2023: [BUG] Deployed hook scripts crash at runtime: required sibling modules are not copied, and the consumer's package.json module type is applied to them
**Describe the bug**

Hook integration copies only the script files referenced directly by hook commands. In
[`hook_integrator.py`](https://github.com/microsoft/apm/blob/main/src/apm_cli/integration/hook_integrator.py),
`_rewrite_hooks_data` extracts the script paths named in each command and the deploy path `shutil.copy2`s exactly
those files into the target tree (`.claude/hooks/<pkg>/`, `.github/hooks/scripts/<pkg>/`). Two problems follow:

1. **Sibling modules are not copied.** DietrichGebert/ponytail's entry scripts begin with
   `require('./ponytail-config')`, `require('./ponytail-instructions')`, `require('./ponytail-runtime')` -- none
   of which are referenced by a hook command, so none are deployed. Every deployed hook crashes with
   `Cannot find module './ponytail-config'`, in every project, for every target.

2. **The copies inherit the consumer project's module type.** Because the deployed scripts live inside the
   consumer's tree, Node resolves their module system from the *consumer's* `package.json`. In any project with
   `"type": "module"` (a very common modern Node setup), the package's CommonJS scripts are parsed as ESM and die
   even earlier: `ReferenceError: require is not defined in ES module scope`. The package's own `package.json`
   (which correctly declares its module semantics) is left behind in `apm_modules/`.

The same scripts run correctly from `apm_modules/<owner>/<pkg>/hooks/`, where all siblings exist and the package's
own `package.json` governs module resolution. So the deployment step is what breaks them.

**To Reproduce**

1. `mkdir /tmp/repro && cd /tmp/repro`
2. `apm install DietrichGebert/ponytail -t copilot`
3. Run a deployed hook the way the harness would:

```
$ echo '{}' | node .github/hooks/scripts/ponytail/hooks/ponytail-activate.js
Error: Cannot find module './ponytail-config'
```

4. Add a typical consumer `package.json` (`{"type": "module"}`) and run again:

```
$ echo '{}' | node .github/hooks/scripts/ponytail/hooks/ponytail-activate.js
ReferenceError: require is not defined in ES module scope
```

5. Contrast with the undeployed original, which works:

```
$ echo '{}' | node apm_modules/DietrichGebert/ponytail/hooks/ponytail-activate.js
PONYTAIL MODE ACTIVE -- level: full   (exit 0)
```

**Expected behavior**

Deployed hook scripts are runnable: their module dependencies are present and they keep the module semantics their
package declares, regardless of the consumer project's `package.json`.

Two possible directions:

- Copy the package's entire hooks directory (not just command-referenced entries) and preserve the package's
  `package.json` module `"type"` alongside the copies, keeping the current self-contained deploy model; or
- Do not copy at all: rewrite hook commands to execute from `apm_modules/<owner>/<pkg>/...`, matching the native
  Claude Code plugin model (`CLAUDE_PLUGIN_ROOT` with the whole plugin directory present). Trade-off:
  `apm_modules/` is gitignored, so deployed hooks would require `apm install` after clone.

**Environment**

- OS: Linux
- Python Version: 3.13
- APM Version: 0.23.1 (also reproduced on main @ `14a87475`)

**Logs**

```
file:///.../.claude/hooks/ponytail/hooks/ponytail-activate.js:9
const fs = require('fs');
           ^
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and
'/.../package.json' contains "type": "module". To treat it as a CommonJS script, rename it to
use the '.cjs' file extension.
```

**Additional context**

Found while validating #2021 (refs #2020): once hook manifests route correctly to the claude target, the deployed
scripts crash at session start with the errors above. The copilot target hits this today with no other bug
involved -- the repro above is stock v0.23.1. Package authors can partially defend with `.cjs` extensions (fixes
the module-type issue, not the missing siblings). Ponytail is authored to the native Claude Code plugin spec,
where the whole plugin directory is present at runtime, so its multi-file scripts are conventional rather than
unusual.

WilliamK112:
## Summary

- Expands deployed hook script copies from only the command entrypoint to the entrypoint runtime subtree, so required sibling modules like `./ponytail-config` are present after install.
- Copies the nearest source `package.json` into the deployed hook runtime directory so hook scripts keep the package module semantics instead of inheriting the consumer project's `package.json` `type`.
- Adds a Node-backed regression for both Copilot and Claude deployments in a consumer project with `type=module`, proving the deployed CommonJS hook can run and load its sibling module.

Closes #2023.

## Preflight

- #2023 was open, unassigned, and had no comments before implementation and again immediately before push.
- Duplicate PR search for #2023 / hook script crash / sibling modules / ponytail runtime / package.json module type terms returned no open PRs immediately before push.

## Validation

- `uv run --extra dev pytest tests/integration/test_hook_script_runtime_deployment_e2e.py -q` (2 passed)
- `uv run --extra dev pytest tests/unit/integration/test_hook_integrator.py tests/integration/test_hook_script_runtime_deployment_e2e.py tests/integration/test_hook_routing_combined_manifest_e2e.py tests/integration/test_hook_integrator_copilot_casing_e2e.py tests/integration/test_integrators_hooks_execution.py tests/integration/test_integrators_hooks_phase3c.py -q` (444 passed)
- `uv run --extra dev ruff check src tests`
- `uv run --extra dev ruff format --check src tests`
- `git diff --check`
- `uv run --extra dev pytest -q` was also attempted; this repo's default marker configuration skipped 29592 tests and deselected 173, so the targeted hook suite above is the relevant behavior evidence.

### DIFF
--- a/src/apm_cli/integration/hook_integrator.py
+++ b/src/apm_cli/integration/hook_integrator.py
@@ -49,7 +49,7 @@ import logging
 import re
 import shutil
 from dataclasses import dataclass, field
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any
 
 import yaml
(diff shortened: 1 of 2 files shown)


sergio-sisternes-epam:
## APM Review Panel -- PR #2028

**Ship recommendation: APPROVE**

---

### What this PR does

Closes a runtime crash: hook scripts that `require('./sibling-module')` failed in consumer projects
with `"type": "module"` in their `package.json` because the deployer only copied the entrypoint.
This PR copies the entrypoint's directory subtree and the nearest source `package.json` alongside
it, so the deployed CommonJS hook runs with its original module semantics.

---

### Correctness

The fix is implemented across 6 focused private helpers feeding two call sites
(`integrate_package_hooks` line 1399, `_integrate_merged_hooks` line 1625). Both sites now
expand script lists before copying. The fallback contract is conservative at every layer -- if
anything fails (path traversal, rglob error, missing package.json), the function returns
the original single-entrypoint pair, preserving pre-PR behaviour rather than crashing.

---

### Security

Path traversal is guarded at every materialisation point via `ensure_path_within`. Symlinks are
skipped before `is_file()` checks throughout `_iter_hook_runtime_files` and
`_nearest_package_json_for_script`. No concerns.

---

### Test evidence

Two new integration tests run `node` as a subprocess and assert `stdout == "PONYTAIL MODE ACTIVE"`,
proving the deployed script can load its sibling module inside a `"type": "module"` consumer project.
`scripts_copied >= 3` asserts expansion happened. 170 pre-existing hook tests continue to pass.

---

### Minor suggestions (non-blocking)

1. `_hook_runtime_target_root`: add a one-line comment on the `len(target_path.parts) <= len(child_parts)` guard explaining the invariant.
2. `_expand_hook_runtime_scripts`: add a brief comment on the `{target: source}` dict inversion before the list comprehension to make the swap explicit.
3. Consider a Cursor or Codex runtime-subprocess test to match the Copilot/Claude pair -- not required, the shared code path is covered, but it would complete the cross-target runtime proof.

---

*Panel: python-architect, supply-chain-security, test-coverage, performance, devx-ux, cli-logging, ceo-synthesizer*

sergio-sisternes-epam (approved):
@WilliamK112  Panel review complete. All findings are non-blocking nits. The PR needs rebase and fix CI and potential conflicts

danielmeppiel:
fixed by #2025; #2028 is stale

PR #2142, opened 2026-07-10
[BUG] Make dependency resolution transactional so a corrected cyclic graph resumes without manual apm_modules deletion

### LINKED ISSUE #2140: [BUG] Cyclic-dependency rejection leaves un-resumable apm_modules snapshot; corrected source still fails until manual rm -rf
**Describe the bug**

When `apm install` rejects a cyclic local dependency graph (exit 1), it leaves a partial `apm_modules/` snapshot on disk. After the user fixes the source to break the cycle, `apm install` **still** exits 1 -- because the stale `apm_modules/` snapshot from the failed run is treated as authoritative. The only way to recover is to manually `rm -rf apm_modules`, after which `apm install` succeeds (exit 0, lockfile written). The failed resolution is not transactional: it neither rolls back its staging directory nor treats a corrected source as a reason to rebuild.

**To Reproduce**

Baseline SHA: `26d8cee8a4079bb3d31956f1b57d343654835fc8`.

1. Two local packages A and B where A depends on B and B depends on A (2-node cycle), consumed by the root project.
2. `apm install` -> **exit 1** (cycle correctly rejected), but `apm_modules/` snapshot is left behind.
3. Edit source to break the cycle (e.g. remove B's back-dependency on A).
4. `apm install` -> **still exit 1** (stale `apm_modules/` blocks the corrected resolution).
5. `rm -rf apm_modules` then `apm install` -> exit 0, lockfile created.

**Expected behavior**

A failed dependency resolution must be transactional: either roll back the partial `apm_modules/` staging on failure, or invalidate/rebuild it when the source graph changes, so that fixing the source and re-running `install` recovers without a manual `rm -rf apm_modules`. A rejected install should not leave a snapshot that poisons the next run.

**Environment**
- OS: macOS
- APM Version: 0.24.1 (`26d8cee8a`)

**Root-cause (architecture)**

Architecture cluster `r2-resolution-staging-transactionality` (recommendation: canonical-owner refactor). Resolution writes `apm_modules/` incrementally with no rollback boundary and no source-graph fingerprint to detect staleness. Recommended: a resolution-staging session owner (e.g. `src/apm_cli/install/resolution_staging.py::ResolutionStagingSession`) that stages to a temp location and commits on success / discards on failure, and/or fingerprints the resolved source graph so a changed source forces a rebuild. Vendor-neutral, no primitive-schema change.

**Additional context**

Authorized first-party reliability/regression validation (red-team Round 2). Independently reproduced twice (finder != confirmer) with control (single non-cyclic install succeeds) and cleanup verification.

danielmeppiel:
# fix(install): make dependency resolution transactional

## TL;DR

A failed cyclic dependency resolution now rolls back only the `apm_modules/` paths staged or replaced by that attempt. After the user fixes the source graph, a normal `apm install` succeeds without manually deleting `apm_modules/`, while valid pre-existing cache entries remain protected.

> [!NOTE]
> Fixes #2140. This deliberately does not use a blanket delete-all recovery path.

## Problem (WHY)

- [x] A two-node local cycle correctly exited 1, but left both package snapshots in `apm_modules/`.
- [x] After the source cycle was fixed, the stale snapshots remained authoritative and the next install still exited 1.
- [x] Recovery required `rm -rf apm_modules`, discarding unrelated valid cache entries.
- [!] This violated P6: ["Behavior must be predictable, auditable, and explainable in plain English."](https://github.com/microsoft/apm/blob/main/PRINCIPLES.md#p6----reliability-over-magic)
- [!] A delete-all fix would violate P4: ["No bug fix, hardening, security patch, or refactor lands if it makes those commands harder, slower, more verbose, or more confusing for a new user."](https://github.com/microsoft/apm/blob/main/PRINCIPLES.md#p4----ux-is-the-floor-not-a-trade)

## Approach (WHAT)

| # | Fix | Principle |
|---|---|---|
| 1 | Register every package path before resolution mutates it. | P6 - Reliability over magic |
| 2 | Move replaced content to a session-unique backup and record newly created paths. | Non-destructive, rollback-scoped staging |
| 3 | Commit only after graph validation; on any resolution exception, remove only session-created content and restore backups. | P4 - UX is the floor, not a trade |
| 4 | Keep the staging layer below target adapters and independent of package source type. | P3 - Vendor neutral by construction |

## Implementation (HOW)

| File | Change |
|---|---|
| `src/apm_cli/install/resolution_staging.py` | Adds the thread-safe `ResolutionStagingSession` and transaction decorator. Existing paths are atomically moved to session backups; rollback restores them in reverse order. |
| `src/apm_cli/install/phases/resolve.py` | Wraps dependency resolution in the staging transaction and registers a path only after cache reuse has been ruled out. |
| `tests/unit/install/test_resolution_transactionality.py` | Reproduces #2140 through the real CLI, proves failed resolution is clean, then fixes the source and proves retry succeeds. Also verifies replaced cache content is restored. |
| `docs/src/content/docs/troubleshooting/install-failures.md` | Documents that cyclic-graph retries no longer require manual `apm_modules/` deletion. |

## Diagram

Legend: dashed nodes are the new transaction boundary around existing resolution and validation.

```mermaid
flowchart LR
    S[Source manifests] --> P[Prepare touched path]
    P --> M[Materialize package snapshot]
    M --> V{Graph valid}
    V -->|yes| C[Discard session backups]
    V -->|no| R[Remove new snapshots and restore backups]
    R --> X[Correct source and retry]
    class P,C,R new;
```

## Trade-offs

- Chose per-path backup and rollback over deleting all of `apm_modules/`; unrelated valid cache entries are preserved.
- Chose same-filesystem sibling staging under `apm_modules/` over copying the full cache; path replacement stays atomic without cache-sized copy cost.
- Kept package materialization unchanged; authenticated-source resolution, content-hash checks, and valid cache short-circuits retain their existing behavior.
- The transaction covers resolution writes, not later deployment phases; this is the narrow owner locus for #2140.

## Benefits

1. A rejected two-node cycle leaves neither newly staged package snapshot behind.
2. Fixing the source graph allows the next plain install to exit 0 and write the lockfile.
3. Copilot instructions deploy on the retry without manual cache deletion.
4. A pre-existing path replaced during a failed attempt is restored in the regression test.

### Scenario evidence

| # | Scenario (user promise) | Principle(s) | Test proving it | Type |
|---|---|---|---|---|
| 1 | Fix a rejected local dependency cycle and rerun install without deleting `apm_modules/`; install succeeds. | DevX, Governed by policy | `tests/unit/install/test_resolution_transactionality.py::test_corrected_local_cycle_resumes_without_manual_cache_deletion` (regression-trap for #2140) | integration |
| 2 | A failed resolution that replaces a valid cache path restores the original content. | Secure by default, Governed by policy | `tests/unit/install/test_resolution_transactionality.py::test_resolution_rollback_restores_replaced_cache_path` | unit |

<details><summary>Test and lint evidence</summary>

Base regression run before production changes:

```text
1 failed in 5.33s
```

Regression after the fix:

```text
2 passed
```

Targeted install suite:

```text
1787 passed in 28.21s
```

CI lint mirror:

```text
All checks passed!
1424 files already formatted
[+] auth-signal lint clean
YAML, file-length, and relative_to guards passed
```

</details>

## How to test

- [ ] Create local packages A and B where A depends on B and B depends on A; run `apm install --target copilot --verbose` and expect exit 1 with no A/B snapshots retained.
- [ ] Remove B's dependency on A without deleting `apm_modules/`.
- [ ] Run the same install command and expect exit 0.
- [ ] Confirm `apm.lock.yaml` and `.github/instructions/package-a.instructions.md` exist.

Fixes #2140

### DIFF
--- a/docs/src/content/docs/troubleshooting/install-failures.md
+++ b/docs/src/content/docs/troubleshooting/install-failures.md
@@ -220,6 +220,8 @@ apm install
 
 The cache short-circuits already-downloaded packages and the integrate phase overwrites partially-deployed files.
 
+If resolution rejects a cyclic dependency graph, fix the package manifests and run `apm install` again. APM rolls back only the package snapshots staged by the rejected resolution, so no manual `apm_modules/` deletion is required.
+
 If files in `apm_modules/` or under target harness directories look corrupt, force a fresh deploy by combining cache bypass with overwrite:
 
 ```bash
--- a/src/apm_cli/install/phases/resolve.py
+++ b/src/apm_cli/install/phases/resolve.py
@@ -24,11 +24,13 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from apm_cli.install.helpers.ref_seed import seed_ref_resolver_from_lockfile
+from apm_cli.install.resolution_staging import transactional_resolution
 from apm_cli.models.apm_package import GitReferenceType, ResolvedReference
 from apm_cli.utils.short_sha import format_short_sha
 
 if TYPE_CHECKING:
     from apm_cli.install.context import InstallContext
+    from apm_cli.install.resolution_staging import ResolutionStagingSession
     from apm_cli.models.dependency.reference import DependencyReference
 
 _logger = logging.getLogger(__name__)
@@ -374,13 +376,9 @@ def _fail_on_resolution_errors(ctx: InstallContext, dependency_graph) -> None:
     raise RuntimeError(f"Dependency resolution failed: {joined_errors}")
 
 
-def _resolve_dependencies(ctx: InstallContext) -> None:
-    """Run ``APMDependencyResolver``, handle errors; populate ``ctx.deps_to_install`` and ``ctx.dependency_graph``.
-
-    Also wires the download callback (which handles transitive package fetching),
-    builds ``ctx.dep_base_dirs``, writes ancillary state to ``ctx``, and cleans up
-    the shared clone cache.
-    """
+@transactional_resolution
+def _resolve_dependencies(ctx: InstallContext, staging_session: ResolutionStagingSession) -> None:
+    """Resolve dependencies and populate the resolution fields on ``ctx``."""
     import threading as _threading
 
     from apm_cli.core.scope import InstallScope
@@ -499,6 +497,7 @@ def _resolve_dependencies(ctx: InstallContext) -> None:
             )
             if not _force_semver_resolve:
                 return install_path
+        staging_session.prepare_path(install_path)
         # F1 (#1116): surface a heartbeat BEFORE the network/copy work so
         # users see the install advancing past silent transitive lookups.
         # Under F7's parallel BFS this callback may run on a worker
--- a/src/apm_cli/install/resolution_staging.py
+++ b/src/apm_cli/install/resolution_staging.py
@@ -1,0 +1,80 @@
+"""Rollback-scoped staging for dependency resolution writes."""
+
+from __future__ import annotations
+
+import threading
+import uuid
+from collections.abc import Callable
+from functools import wraps
+from pathlib import Path
+from typing import Any
+
+from apm_cli.utils.path_security import ensure_path_within, safe_rmtree
+
+
+class ResolutionStagingSession:
+    """Track paths mutated during resolution and restore them on failure."""
+
+    def __init__(self, apm_modules_dir: Path) -> None:
+        """Create an empty staging session rooted below ``apm_modules``."""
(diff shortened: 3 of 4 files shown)


danielmeppiel:
Superseded by #2155. The reliability campaign consolidated all 13 point fixes into a single architectural cure PR that re-homes each fix under its canonical owner module and adds owner-invariant guards. The fix for #2140 is folded into #2155 and re-verified there (full CI green). Closing this point PR in favor of the consolidated cure; the branch is preserved and this can be reopened if #2155 is not merged.

PR #2153, opened 2026-07-11
fix(mcp): purge APM-managed MCP servers from dropped targets on contraction (#2149)

### LINKED ISSUE #2149: [BUG] Contracting the target set leaves APM-managed MCP servers stranded in the removed harness's native config
## Summary
When the active target set is contracted (a target that previously received MCP configuration is dropped and the project is recompiled/reinstalled), APM's MCP cleanup only visits the *currently active* targets and removes servers by identity (`stale_names`). The dropped target's native MCP config is never revisited, so APM-managed MCP server entries previously written there (in `.mcp.json` / `config.toml` / the Claude config, etc.) remain behind.

## Impact
- Removed Codex / Kiro / Windsurf (and other) targets retain APM-managed MCP entries after they are dropped from the active set; the harness keeps advertising MCP servers APM no longer manages for it.
- Native schemas are otherwise intact (JSON/TOML and Codex literal path keys are preserved) -- this is a cleanup / lifecycle gap, not a schema-format defect.

## Reproduction (Round 3 reliability sweep; finder + independent confirmer both reproduced)
1. Install a package that contributes an MCP server across a broad target set.
2. Contract to a single target (e.g. `copilot`) and reinstall twice.
3. Add `claude`, then replace it with `cursor`.
4. Observe the MCP server still present in the removed targets' native configs.
Control: removing the MCP *dependency* while the broad target set is still active cleans all entries correctly.

## Root cause (canonical-owner gap)
The MCP write path deploys to whatever targets are active at write time, but the cleanup path (`src/apm_cli/integration/mcp_integrator.py`: `_clean_json_mcp_config` / `_clean_toml_mcp_config` / `_clean_claude_config`) only iterates active targets and diffs *server identity*. There is no per-target deployment provenance recording "APM wrote server X into target Y", so when Y leaves the active set it is never revisited to purge X. Two sibling paths (write vs cleanup) act on the same fact (which targets carry an APM-managed server) without a shared owner.

## Expected
Dropping a target from the active set should purge the APM-managed MCP entries APM previously wrote into that target's native config (or the cleanup pass should consult per-target provenance rather than only the currently active targets).

## Strategy compliance
No schema translation required -- each harness's native MCP format is read/written as-is; this is lifecycle cleanup of entries APM itself owns. Vendor-neutral.

danielmeppiel:
## Problem

When an install contracts its active target set, MCP cleanup only visits active runtimes. APM-managed servers therefore remain stranded in native configs for targets that were removed.

## Fix

- Record successful MCP writes as per-target ownership in `apm.lock.yaml`.
- Reconcile that ownership against the active runtime set on reinstall.
- Dispatch existing native JSON, TOML, and Claude cleanup helpers for removed targets.
- Remove only server names recorded as APM-managed, preserving user-authored servers, Codex literal path keys, and unrelated native settings.
- Document the lockfile ownership field and contraction behavior.

## Test

- Added an end-to-end regression that installs across Copilot and Codex, contracts to Copilot, and verifies the Codex APM-managed entry is removed while a user-authored server and literal Windows path key remain.
- `uv run --extra dev python -m pytest tests/integration/test_mcp_install_flow.py -q` - 28 passed.
- `uv run --extra dev python -m pytest tests/unit -k mcp -q` - 1430 passed, 1 skipped.
- Ruff check/format, pylint R0801, source guards, and `scripts/lint-auth-signals.sh` passed.

Closes #2149

apm-spec-waiver: lifecycle cleanup of APM-owned MCP entries on target contraction; native schemas unchanged, no new OpenAPM artifact or wire behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   
 ### Fixed
 
+- Contracting the active target set now removes APM-managed MCP server
+  entries from dropped targets while preserving user-authored servers and
+  unrelated native config. (#2149)
 - Fresh checkouts with declared consumer targets no longer remain
   `apm audit --ci`-red for files those targets cannot restore: `apm install`
   now removes stale `deployed_files` entries outside the legitimate target
--- a/docs/src/content/docs/consumer/install-mcp-servers.md
+++ b/docs/src/content/docs/consumer/install-mcp-servers.md
@@ -128,6 +128,10 @@ the gate took effect:
 [i] Skipped MCP config for claude, codex  (active targets: copilot)
 ```
 
+On reinstall, removing a previously configured target also removes the
+APM-managed server entries from that target's native config. User-authored
+servers and unrelated JSON or TOML settings remain unchanged.
+
 This single rule replaces two older ones that used to coexist:
 
 - A "directory opt-in" carve-out for Cursor / Gemini / OpenCode -- now
--- a/docs/src/content/docs/reference/lockfile-spec.md
+++ b/docs/src/content/docs/reference/lockfile-spec.md
@@ -99,6 +99,12 @@ mcp_configs:
   transitive-server:
     type: stdio
     command: local-server
+mcp_target_servers:
+  codex:
+    - github
+  copilot:
+    - github
+    - transitive-server
 mcp_config_provenance:
   transitive-server: local-package
 lsp_servers:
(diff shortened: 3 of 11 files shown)


danielmeppiel:
Superseded by #2155. The reliability campaign consolidated all 13 point fixes into a single architectural cure PR that re-homes each fix under its canonical owner module and adds owner-invariant guards. The fix for #2149 is folded into #2155 and re-verified there (full CI green). Closing this point PR in favor of the consolidated cure; the branch is preserved and this can be reopened if #2155 is not merged.

PR #2173, opened 2026-07-12
fix(deps): render logical local dep keys in deps list/tree; refresh stale v0.25.0 release tests

danielmeppiel:
fix(deps): render logical local dep keys in deps list/tree; refresh stale v0.25.0 release tests

## TL;DR

The v0.25.0 release integration suite ([run 29184076146](https://github.com/microsoft/apm/actions/runs/29184076146)) went red with 13 failures across four independent root causes. Exactly one is a production bug: `apm deps list` and `apm deps tree` leak physical, parent-scoped install slots for transitive **local** dependencies instead of their logical lockfile identity. This PR fixes that display regression and regrounds three stale test fixtures that drifted behind intentional prior changes (#2058, #2155, and upstream `awesome-copilot`). No production behavior other than local-dep display changes; physical hashed storage, remote behavior, and lockfile keys are untouched.

## Problem (WHY)

Thirteen failures, one production bug, three stale tests:

- **Production regression (`deps list`/`deps tree`)**: for a local transitive chain, `apm deps list` printed `_local/<12-char-hash>/pkg-depth-N` and `apm deps tree` printed `local:/abs/host/path/pkg-depth-N`. Both leak the parent-scoped physical slot (introduced by #2155 to avoid sibling collisions) and, in the tree case, the **absolute host filesystem path** into user-facing output. The logical identity the user typed is `../pkg-depth-N` (lockfile `repo_url` `_local/pkg-depth-N`).
- **Stale policy test**: `test_emu_ssh_remote_routes_to_correct_org_policy_repo` asserted the first policy candidate is `contoso/.github`, but #2058 intentionally made `.github-private` first in the cascade.
- **Stale pack fixture**: two `test_pack_apm_provenance_e2e` cases built a lockfile shape no real install produces -- a skill **directory** in `deployed_files` with per-**child** hashes. #2155's deployment ledger rebuilds hashes only for paths present in `deployed_files`, so the fixture never exercised the intended path.
- **Drifted plugin fixture**: `test_plugin_e2e` installed `github/awesome-copilot/plugins/context-engineering` from default `main`, which upstream has since reduced to a thin stub whose manifest declares absent component paths. APM correctly fails closed (#2103/#2155); the assembled plugin now lives on the upstream `marketplace` branch.

[...]

PR #2174, opened 2026-07-12
fix(deps): preserve local identity display boundaries

danielmeppiel:
# fix(deps): preserve local identity display boundaries

## TL;DR

This follow-up closes two display edge cases found after #2173 merged: Windows absolute local paths parsed on a POSIX host could still reach `deps tree`, and a declared remote identity shaped like a local hash slot could be shortened incorrectly. The patch derives absolute-local labels from the declared path's cross-platform basename and limits hash redaction to confirmed orphan slots.

> [!IMPORTANT]
> This is a narrow release-recovery follow-up on top of merged PR #2173; it does not change local dependency storage, resolution, or lockfile identity.

## Problem (WHY)

- [x] A real `DependencyReference.parse(r"C:\Users\alice\pkg")` to `LockedDependency` flow rendered `_local/C:\Users\alice\pkg@latest` on POSIX instead of `_local/pkg@latest`.
- [x] `_logical_local_display()` ran before orphan status was known, so a declared remote identity such as `_local/abcdef123456/pkg` was displayed as `_local/pkg`.
- [!] Both cases sit on user-visible `apm deps list/tree` output added in #2173, so they must preserve logical identity without exposing host-specific paths.

The two regression tests were added first and failed with the exact outputs above. That keeps the follow-up grounded in ["deterministic tool execution transforms probabilistic generation into verifiable action"](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/).

## Approach (WHAT)

| # | Fix |
|---:|---|
| 1 | For absolute local paths, derive `_local/<basename>` with `PureWindowsPath` or `PurePosixPath`; keep relative paths and missing-path fallbacks unchanged. |
| 2 | Determine orphan status from the raw logical identity before display redaction. |
| 3 | Strip `_local/<12-hex>/` only for confirmed unmapped orphan slots; preserve declared remote identities byte-for-byte. |
| 4 | Lock both boundaries with real parser, lockfile, and filesystem regression cases. |

## Implementation (HOW)

- **`src/apm_cli/commands/deps/cli.py`** - adds a small absolute-local display helper and moves the orphan predicate ahead of display-name selection. Physical identity remains the authority for orphan detection; only the rendered orphan label is sanitized.
- **`tests/unit/commands/test_deps_cli_helpers.py`** - adds a parser-to-lock-to-display Windows regression and a real lockfile/filesystem case proving that a declared remote hash-shaped identity is neither orphaned nor shortened.

## Diagrams

[...]

PR #2175, opened 2026-07-12
fix(deps): harden local-dep display against Windows repo_url leak and over-eager hash stripping

danielmeppiel:
## TL;DR

Two verified pre-merge follow-ups to the v0.25.0 local-dependency display hardening (PR #2173, now merged). Both leaked or corrupted the user-facing local-dependency identity in `apm deps list`/`tree` output. Fixed TDD-first; no behavior change for the common relative-path case.

## Problem (WHY)

PR #2173 hardened local-dep display but two edge cases still leak/corrupt identity:

1. **Windows-path `repo_url` pollution.** A Windows-form declared path parsed on POSIX (`C:\packages\pkg`) leaves the drive letter and backslashes embedded in the parsed `repo_url` (`_local/C:\packages\pkg`), because `PurePosixPath("C:\\...").name` treats backslashes as ordinary characters. `_dep_display_name` fell back to that polluted `repo_url` for absolute paths, leaking host filesystem structure into user-facing output.

2. **Over-eager orphan hash stripping.** `_resolve_scope_deps` stripped the hash segment from *every* scanned `_local/<12hex>/pkg` slot name. A legitimately declared/mapped slot whose logical key happens to look hashed was corrupted into `_local/pkg` and masked, instead of being reported under its true logical identity.

## Approach (WHAT)

1. Derive a hash-free `_local/<basename>` from `local_path` for absolute declared paths, via a new `_local_basename` helper that routes drive/UNC/backslash paths through `PureWindowsPath` and everything else through `PurePosixPath`. Relative `local_path` stays verbatim; a missing `local_path` still falls back to `repo_url`. `get_unique_key` is still never called for local deps.

2. Gate the hash-strip in the scan loop on a **confirmed orphan** (a slot with no lockfile mapping to a logical key). Declared/mapped slots keep their logical name verbatim. Orphan *detection* still runs on the raw physical identity, so detection accuracy is unchanged.

## Implementation (HOW)

- `src/apm_cli/commands/deps/cli.py`: add `_local_basename`; rework `_dep_display_name` local branch; compute `is_orphaned` before the scan `try` and apply `_logical_local_display` only when orphaned.
- `tests/unit/commands/test_deps_cli_helpers.py`: real `DependencyReference.parse(r"C:\packages\pkg") -> LockedDependency -> _dep_display_name` regression (asserts no drive/backslashes); declared-hashed-slot preservation test (asserts `_local/<12hex>/pkg` is preserved, not orphaned). Both were RED before the fix.

## Trade-offs

Detection intentionally uses `PureWindowsPath` heuristics (backslash or drive) rather than importing the heavier path resolver -- keeps this a bounded release patch with no new cross-cutting abstraction.

## Validation evidence

| Check | Result |
| --- | --- |
| New RED->GREEN unit tests | 2 new, RED confirmed then GREEN |
| `test_deps_cli_helpers.py` | 25 passed |
| Related deps modules (local_deps, deps_utils, list_tree_info, wave6, deps_cli_coverage) | 282 passed |
| `test_transitive_chain_e2e.py` (binary shim) | 4 passed |
| ruff check / ruff format --check | clean |
| auth-signals lint | clean |
| pylint R0801 | 10.00/10 |
| ASCII-only additions | verified |

## How to test

```
uv run --extra dev pytest tests/unit/commands/test_deps_cli_helpers.py -q
```

Does not merge or move any tags.

### DIFF
--- a/src/apm_cli/commands/deps/cli.py
+++ b/src/apm_cli/commands/deps/cli.py
@@ -93,6 +93,20 @@ def _logical_local_display(physical_name: str) -> str:
     return physical_name
 
 
+def _local_basename(path: str) -> str:
+    """Cross-platform basename of a local dependency path.
+
+    A Windows-form path parsed on POSIX (``C:\\packages\\pkg``) is opaque to
+    ``PurePosixPath`` -- the backslashes are ordinary characters, so ``.name``
+    returns the whole string and the drive leaks. Route drive/UNC/backslash
+    paths through ``PureWindowsPath`` and everything else through
+    ``PurePosixPath`` so the leaf name is recovered on either host.
+    """
+    if "\\" in path or PureWindowsPath(path).drive:
+        return PureWindowsPath(path).name
+    return PurePosixPath(path).name
+
+
 def _dep_display_name(dep: LockedDependency) -> str:
     """Get display name for a locked dependency (key@version).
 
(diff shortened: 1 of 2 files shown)


danielmeppiel:
Superseded by #2174, which already landed both round-3 fixes in d73e6ac364. v0.25.0 has been published from that commit; no additional patch is needed.

PR #2239, opened 2026-07-16
test(auth): add Windows SSH passphrase evidence fixture

danielmeppiel:
# test(auth): add Windows SSH passphrase evidence fixture

## TL;DR

This draft adds a hosted Windows evidence contract for issue #1976 without changing APM product behavior. It builds the packaged `apm.exe`, serves an ephemeral package through loopback Windows OpenSSH, and runs encrypted-correct, unencrypted, wrong-passphrase, and missing-passphrase cases through real Git processes. Hosted run [`29469702438`](https://github.com/microsoft/apm/actions/runs/29469702438) passed all four expectations: the reported failure did not reproduce when Windows OpenSSH received the correct response through `SSH_ASKPASS`.

> [!IMPORTANT]
> Evidence baseline: `796e229805ade19244683f0a1e70ec534ce4fb85`; exact evidence head: `f7c9d0e23e9800f06ce2792306ac316fca6970ba`. This is a fixture-only evidence draft. Never merge it from this lane.

## Problem (WHY)

- [x] Issue #1976 reports that Windows installs work with an unencrypted SSH key but fail when the key has a passphrase.
- [x] The issue's original log ended in a port-22 timeout, so it did not isolate passphrase prompting from network reachability.
- [x] Existing coverage does not cross packaged APM, Git for Windows, Windows OpenSSH, an encrypted key, and passphrase handling in one boundary.
- [!] A source-only or mocked test could report the expected environment while missing frozen-process or Windows child-process behavior.

Why this matters: ["Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) The fixture uses concrete process and artifact receipts because ["agents pattern-match well against concrete structures"](https://agentskills.io/skill-creation/best-practices).

## Approach (WHAT)

| # | Fixture decision | Evidence produced |
|---:|---|---|
| 1 | Build the repository's packaged Windows CLI | Binary path, SHA-256, version, and source head |
| 2 | Bind Windows `sshd` to `127.0.0.1:22222` | Server logs, host key, forced `git upload-pack` command |
| 3 | Generate encrypted and unencrypted Ed25519 keys at runtime | Public fingerprints only; private material is deleted |
| 4 | Prove the server with direct unencrypted `git ls-remote` | Independent Git/OpenSSH receipt before packaged APM runs |
[...]

Issue #1976, opened 2026-07-01
[BUG] SSH connection with passphrase fails to clone


bugra-derre:
**Describe the bug**
Users that have a git connection via SSH having a passphrase are failing to install APM dependencies

**To Reproduce**
1. Add SSH key with passphrase to a gh repo
2. Run apm install / update

**Expected behavior**
APM requests user to prompt the passphrase for git clone /checkout operation

**Environment (please complete the following information):**
 - OS: Windows PowerShell
 - Python 3.14.0
 - APM Version: [e.g. 0.21.0]

**Logs**
```
Failed to download dependency
sandbox/arcade-pack: Failed to clone repository sandbox/arcade-pack via
SSH. To allow cross-protocol fallback (not recommended), pass --allow-protocol-fallback, set
APM_ALLOW_PROTOCOL_FALLBACK=1, or run: apm config set allow-protocol-fallback true Please check
repository access permissions and authentication setup. Last error: Cmd('git') failed due to: exit
code(128)
  cmdline: git clone -v --depth=1 --
org@org.ghe.com:sandbox/arcade-pack.git
C:\Users\USER\AppData\Local\Temp\tmpddtpqv9y
  stderr: 'Cloning into 'C:\Users\USER\AppData\Local\Temp\tmpddtpqv9y'...
ssh: connect to host org.ghe.com port 22: Connection timed out
fatal: Could not read from remote repository.
 
Please make sure you have the correct access rights
and the repository exists.
```

**Additional context**
none

danielmeppiel:
@bugra-derre thanks for the report. Can you please clarify: (1) Can you reach org.ghe.com:22 from your machine? (ssh -vT org@org.ghe.com or nc -zv org.ghe.com 22). (2) Is ssh-agent running with your key loaded? (ssh-add -l). If the answer to (1) is no, the root cause is network/GHE firewall, not APM. If ssh-agent is not running, APM could improve its error message to suggest ssh-add before apm install.

bugra-derre:
@danielmeppiel thanks for the clarifiying question.

Tbh it was not on my machine. But on my colleague's one.

I observed the behaviour like this:
- Yes. org.ghe.com:22 was reachable from his machine.
- He was able to check it out via ssh having no passphrase - but failed to check it out when having a ssh key with passphrase (so the passphrase was the only difference)
- My assumption is that the required prompting for the ssh passphrase is not passed through to the user - might that cause an issue within apm?

[...]

PR #2242, opened 2026-07-16
[DRAFT / DO NOT MERGE] Evidence: interactive ConPTY SSH passphrase repro for #1976

danielmeppiel:
## This is a DRAFT, EVIDENCE-ONLY PR. It must NOT be merged.

Its sole purpose is to run on a hosted `windows-latest` GitHub Actions runner and produce artifact evidence for issue #1976's **remaining, previously-untested boundary**. Once the evidence is captured and reported, this PR should be closed unmerged.

### Context

Closed PR #2239 proved that the **forced-`SSH_ASKPASS`, non-interactive** path works: when a script-based askpass helper supplies the passphrase, packaged `apm.exe` -> Git for Windows -> Windows OpenSSH successfully authenticates against an encrypted Ed25519 key. That PR explicitly flagged one gap it never exercised: a **real interactive console passphrase prompt**, as an actual human typing into a PowerShell window would experience it -- with no askpass helper and no agent.

This PR is an independent, from-scratch fixture (branched fresh from `main` @ `796e229805`, not built on #2239's branch or files) that closes that specific gap. It does not retest the already-proven forced-askpass path.

### What it does

1. Builds the real packaged `apm.exe` via the repo's existing `scripts/windows/build-binary.ps1`.
2. Stands up an ephemeral bare git repo served over the pre-installed Windows OpenSSH `sshd` service, bound to `127.0.0.1` only, with a forced `git upload-pack` command in `authorized_keys` (no shell access).
3. Generates ephemeral Ed25519 keys (one encrypted, one plain) -- generated fresh per run, never committed, never reused.
4. Strips `SSH_ASKPASS`/`GIT_ASKPASS`/`SSH_ASKPASS_REQUIRE`/`DISPLAY`/`SSH_AUTH_SOCK`/`SSH_AGENT_PID` **and** ambient CI markers (`CI`, `GITHUB_ACTIONS`, etc. -- confirmed via `src/apm_cli/registry/operations.py` and three other call sites to gate several of apm_cli's own non-interactive code paths) from the child environment, so the child process sees an environment indistinguishable from a real human's PowerShell terminal.
5. Drives `apm install` through a genuine Windows ConPTY pseudo-console (`CreatePseudoConsole`/`CreateProcessW`, kernel32 P/Invoke only, no external dependency) across 4 hermetic scenarios:
   - `unencrypted-positive` -- control, no prompt expected
   - `encrypted-correct` -- types the correct passphrase into the real prompt
   - `encrypted-wrong` -- types an incorrect passphrase
   - `encrypted-cancel` -- supplies no input; bounded timeout/cancel
6. Captures per-scenario evidence (prompt text observed, console input metadata, process/env tree, `GIT_TRACE2` events, ssh/git/apm exit codes, binary SHA/versions, materialization outcome) into a sanitized `summary.json`, scans it for secret/key-header leakage before upload, and computes a RED (prompt never observed -> reproduces the #1976 boundary) or GREEN (interactive path works) verdict.
7. Restores the runner's `sshd` service to its original state in a `finally` block (including on partial setup failure, via a state snapshot captured before any destructive action).

### Constraints honored

- No mocks, no external git network access, no stored/committed secrets.
- No product code changes (verified: this diff touches only new files under `scripts/windows/` and `.github/workflows/`).
- No credential-abstraction refactor, no taxonomy edits, no fix attempted here -- pure evidence collection.
- Workflow: narrow trigger (`workflow_dispatch` + path-filtered `pull_request` scoped to this fixture's own files only), least-privilege (`permissions: contents: read`), SHA-pinned actions, evidence upload gated by a secret-scan step, `if: always()` so evidence uploads even on failure.

### How this will be used

Once the hosted run completes, the `summary.json` verdict will be read and reported back verbatim (RED with the exact boundary detail, or GREEN with a non-reproduction note and recommendation) to the coordinating session. **This PR will then be closed without merging.**


danielmeppiel:
## Result: technical blocker in the evidence-collection harness (not a verdict on #1976)

After several hosted-run iterations, this fixture cannot currently produce a
trustworthy RED/GREEN verdict on the real-interactive-passphrase-prompt
boundary. The blocker is isolated to the ConPTY evidence-collection
mechanism itself, not to apm's behavior.

### What was proven trustworthy
- Packaged `apm.exe` genuinely runs on the hosted Windows runner and invokes
  `git.exe`/`ssh.exe` (confirmed via trace2 in an earlier run).
- A real dependency-URL bug was found and fixed in the fixture: apm's SSH
  dependency parser requires an `owner/repo`-shaped path
  (`src/apm_cli/models/dependency/reference.py::_validate_final_repo_fields`);
  the fixture's original single-segment bare-repo path
  (`.../conpty-fixture.git`) was rejected by apm during manifest parsing
  before ssh.exe was ever invoked. Fixed by using a cosmetic two-segment URL
  path only (the forced-command sshd config always serves the same real
  bare repo regardless of the requested path).

### The unresolved blocker
Added a cheap ConPTY self-test (`Test-ConPtyCapture`) that runs a trivial
`cmd.exe /c "echo <marker> & exit /b 7"` through the exact same
`ConPtyProcess.Start`/`ReadLoop` pipeline the real scenarios use, before any
expensive sshd/origin setup. On every hosted run this self-test fails:

- The marker string appears **directly in the raw GitHub Actions job log**
  (unprefixed, not through any `Write-*` helper in this script), proving the
  child process's actual stdout is reaching some ambient console/log stream.
- The transcript captured through my ConPTY output pipe shows only
  conhost's own initial VT100 mode-set negotiation bytes
  (`\x1b[?9001h\x1b[?1004h`, 16 bytes) -- emitted automatically when
  `CreatePseudoConsole` succeeds, regardless of any child ever attaching.
- Added byte-count-only reader diagnostics (`GetDiagnostics()` /
  `Get-ConPtyDiagnostics`) confirm: `CreatePseudoConsole` and `CreateProcessW`
  both succeed (a real HPCON handle and child PID are obtained), the reader
  thread starts, does exactly **one** successful `Read()` (16 bytes), and
  then **blocks indefinitely** -- no EOF, no exception, no further reads --
  for the entire 10s self-test window and beyond, until force-terminated.

This means the child process is never actually attached to my pseudo
console's output side, despite `PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE` being
set correctly on the attribute list passed to `CreateProcessW`.

### Why this isn't a quick fix
`ConPtyNative.cs` was reviewed line-by-line against Microsoft's official
ConPTY reference sample (struct layouts, P/Invoke signatures, creation
flags, attribute values) three separate times this investigation, plus
verified to compile cleanly via a real `dotnet build` (not just PS-AST
string parsing). No structural defect was found: `STARTUPINFOEX`/
`STARTUPINFO` field layout matches the native ABI size (104/112 bytes on
x64) exactly; `bInheritHandles=false`; no `STARTF_USESTDHANDLES`;
`UpdateProcThreadAttribute` value/size arguments match known-good samples.
The defect appears specific to how the hosted GitHub Actions Windows runner
environment handles console attachment for a ConPTY-created child -- a
plausible but unconfirmed hypothesis given no interactive/debugger access
to the runner itself.

### Conclusion
No reliable evidence -- RED or GREEN -- can be produced about the real
interactive-passphrase-prompt boundary until this harness-level ConPTY
child-attachment defect is understood and fixed. Reporting this as a
precise technical blocker rather than fabricating or trusting a verdict
from an unproven capture mechanism. **This PR must not be merged**; it
exists solely to host the evidence trail (workflow runs, logs, this
comment) for whoever continues this investigation.

danielmeppiel:
## Update: process-tree diagnostic + DSR(6) hypothesis ruled out (reactivated lane)

Continuing from the prior blocker report, this session (reactivated as candidate
authoritative lane after the parallel "original ConPTY" lane failed a second time
with zero recorded diff) made two further attempts to root-cause the harness-level
ConPTY capture defect, before concluding it warrants reporting as a precise
technical blocker again rather than continued open-ended guessing:

### 1. DSR(6) cursor-position handshake hypothesis -- tested and disproven here

Found an outwardly-identical symptom in an unrelated project
(`sudoprivacy/pty-expect` issue #1, closed via commit `814287bce326`) on the same
`windows-latest` runner image, root-caused there to ConPTY's internal
`CSI 6n` (Device Status Report) handshake not being answered by the reader.
Implemented the equivalent auto-responder in `ConPtyNative.cs`
(`RespondToDeviceStatusReportRequests`, shared `_inputWriteLock`), verified it
compiles cleanly (`dotnet build`, 0 warnings/errors), and re-ran hosted
(run `29473769848`).

**Result: identical failure, and no `CSI 6n` ever appears in the captured stream.**
This rules out DSR(6) as the cause in this specific environment -- the failure
mode here is different from pty-expect's (there, `CSI 6n` was the literal first
and only chunk read; here, only conhost's own VT100 mode-set bytes ever arrive,
and nothing else, DSR or otherwise).

### 2. Live process-tree diagnostic -- pseudo-console creation confirmed structurally correct

Added a `Get-CimInstance Win32_Process` snapshot (~700ms after self-test launch,
extended the child's lifetime with a loopback-only `ping -n 3 127.0.0.1` delay)
to directly answer: is the child process actually attached to the expected
pseudo-console, or something else? Re-ran hosted (run `29473912528`). Result:

```
pid=9976 parentPid=3596 name=conhost.exe  commandLine=...conhost.exe --headless --width 120 --height 32 --signal 0x84c --server 0x844
pid=3300 parentPid=3596 name=cmd.exe      commandLine="C:\Windows\System32\cmd.exe" /c "echo <marker> & ping -n 3 127.0.0.1 >nul & exit /b 7"
(this script's own pid for reference: 3596)
```

A genuine headless conhost host process (matching ConPTY's expected
`--headless --server` invocation pattern) was spun up as a sibling of the child
`cmd.exe`, both parented directly by the fixture's own PowerShell process (which
is the expected/correct Win32 parentage regardless of console attachment --
`CreateProcessW`'s caller is always the reported parent). `cmd.exe`'s command
line, PID, and lifetime all match expectations exactly; it is genuinely running
(not suspended, not exited early).

Combined with re-verifying `CreatePseudoConsole`/`CreateProcessW` call sites a
5th time (pipe handle wiring `hInput=inputReadSide`/`hOutput=outputWriteSide`
matches the API contract; `STARTUPINFOEX` is zero-initialized via `new` with no
stray `hStdInput/Output/Error` handles; `creationFlags=0x00080400` is exactly
`EXTENDED_STARTUPINFO_PRESENT | CREATE_UNICODE_ENVIRONMENT` with no
`CREATE_SUSPENDED`/`CREATE_NEW_CONSOLE` bits), there is no structural defect
visible in the pseudo-console setup itself.

### Net conclusion

The child process is created correctly and attached to a genuine ConPTY host
process per the Win32 API contract, yet its real console output never reaches
the private ConPTY read pipe -- while leaking directly into the raw (un-redirected)
GH Actions job log instead. Both the DSR(6) protocol-handshake theory and a
"child not attached to the pseudo console" theory are now ruled out by direct
evidence. A brief cross-check of another project's Windows ConPTY CI
(`postmeridiem/clide`) suggests hosted `windows-latest` runners CAN support
working ConPTY sessions in general, so this is not necessarily an inherent
runner limitation either -- it may be a subtler defect specific to this
fixture's setup that requires interactive/live debugging on the runner (not
available to this agent) to isolate further.

**This remains an unresolved harness-level blocker, not a RED or GREEN verdict
on the actual #1976 interactive-passphrase-prompt question** -- the fixture
cannot yet reach the real 4 scenarios (unencrypted success, encrypted
correct/wrong/cancel) because the self-test that gates them has not yet passed.
No product code has been touched; PR remains draft and must not be merged.

danielmeppiel:
## Update 2: ResizePseudoConsole kick tried; sharper characterization of the defect

One more concrete, cheap hypothesis tried before reporting further: several
third-party ConPTY wrapper libraries document that some ConPTY host builds
buffer/withhold child output until the console is explicitly resized at least
once after creation (even to its own initial size). Added a
`ResizePseudoConsole(pseudoConsole, size)` call immediately after
`CreateProcessW` succeeds (`ConPtyNative.cs`, commit `cf66d8bf6f`). Verified
compile-clean, re-ran hosted (run `29474242466`).

### Result: partial change, and it sharpens the picture

The captured transcript grew from 16 bytes to 135 bytes across two `Read()`
calls:

```
Read() returned 16 bytes   -> \x1b[?9001h\x1b[?1004h   (win32-input-mode, focus-tracking)
Read() returned 119 bytes  -> \x1b[?25l\x1b[2J\x1b[m\x1b[H ... \x1b[H\x1b]0;C:\Windows\System32\cmd.exe\x07\x1b[?25h
```

That second chunk is: hide cursor, clear screen, reset attributes, home
cursor, home cursor again, **set console title to `C:\Windows\System32\cmd.exe`**,
BEL, show cursor. This is standard Windows-console-subsystem boilerplate
synthesized automatically whenever a new process attaches to a console
(the title auto-set to the child's own image path is a well-known default
OS behavior, not something the child process explicitly writes) -- it is
not literal output the child chose to produce.

**The marker text itself (`echo CONPTY_SELFTEST_...`) still never appears in
the captured transcript**, even though it is confirmed running (`cmd.exe`
appears in the live process-tree snapshot with the correct command line,
correct PID, correctly parented) and confirmed producing output (the same
marker text leaks directly into the raw, un-redirected GH Actions job log,
completely bypassing this fixture's private ConPTY pipe, on every single run
so far).

### Sharper conclusion

ConPTY/conhost correctly relays its **own OS-synthesized console bookkeeping**
(mode-set sequences, cursor sequences, auto title-set, screen clear) to this
fixture's reader -- proving the pipe, the reader thread, and the resize kick
all function. What specifically never arrives is the **child process's own
command output** (the literal `echo` text), even though that exact text is
demonstrably produced by the running child and lands somewhere GH Actions'
own log capture can see it directly.

This is a materially sharper characterization than the prior report ("nothing
but 16 bytes ever arrives") and rules out three hypotheses now: DSR(6)
handshake (disproven), non-attachment to the pseudo-console (disproven via
process tree), and missing resize-kick (tried; helped surface more OS
boilerplate but not the child's literal output).

**Still an unresolved harness-level blocker, not RED or GREEN on the actual
#1976 interactive-passphrase-prompt question.** The remaining defect appears
specific to how this GH Actions Windows runner's console subsystem routes a
child's own `WriteConsoleW`/`WriteFile` output when a second, independently
created pseudo-console session exists alongside the runner's own (redirected,
non-console) PowerShell process -- a scenario that would need interactive/live
debugging on the runner itself (not available to this agent) to isolate
further. No product code touched; PR remains draft, unmerged.

danielmeppiel:
## Update 3: decisive proof -- the pseudo-console's screen buffer is genuinely empty

Final probe before reporting: since a same-size `ResizePseudoConsole` only
flushed host bookkeeping (title-set), tried a resize to a **genuinely
different** size (120x32 -> 100x24 -> back to 120x32, ~700ms after launch, by
which point `cmd.exe` should already have executed its `echo`) to force a real
reflow/full-repaint, on the theory that a real repaint must emit whatever text
actually exists in the buffer cells. Added `ConPtyProcess.Resize()` +
`Resize-ConPtySession` wrapper (commit `4a99b74dcb`). Verified compile-clean +
PS-AST clean, re-ran hosted (run `29474499601`).

### Result: conclusive

The forced full-repaint transcript is:

```
\x1b[?25l\x1b[8;24;100t\x1b[K \x1b[K \x1b[K ... (24 identical erase-line sequences) ...
\x1b[H\x1b[?25h\x1b[?25l\x1b[K \x1b[K ... (another ~24 identical erase-line sequences) ...
\x1b[K\x1b[H\x1b[?25h
```

Every single row in the repaint is an **empty** `CSI K` (erase line, no
preceding text) -- this is XTerm's window-resize-report (`CSI 8;24;100t`)
followed by a full-screen repaint of an entirely blank buffer. There is no
partial/garbled/truncated marker text anywhere in either repaint pass.

**This proves, conclusively and not as a timing/flush artifact, that
`cmd.exe`'s `echo <marker>` output was never written into this pseudo
console's screen buffer at all** -- a full repaint of real buffer content
would have included it if it were there, exactly as it correctly showed the
host's own title-set bookkeeping in the prior probe. Combined with the
process-tree evidence (the child is definitely running, correctly parented,
correct command line, correct PID) and the fact the same marker text
reliably leaks into the raw GH Actions job log on every run, this rules out
every remaining "capture/flush/timing" theory: the child process is
structurally attached to the pseudo console per the Win32 API contract, but
its actual `WriteConsoleW`/`WriteFile` output is being rendered somewhere
else entirely, not into this ConPTY session's buffer.

### Conclusion for this evidence lane

Given:
- 5+ full code reviews of `ConPtyNative.cs` against Microsoft's reference
  ConPTY sample (struct sizes, P/Invoke signatures, pipe-handle wiring,
  creation flags) with no defect found, verified via clean `dotnet build`
  every time;
- Two protocol/attachment hypotheses (DSR(6) handshake, non-attachment to
  the pseudo console) tested and disproven with direct evidence;
- A same-size resize kick and a genuine-reflow resize kick both tested,
  proving the reader/pipe mechanism itself works (host bookkeeping does
  arrive) while conclusively proving the buffer never contains the child's
  actual output;
- A brief cross-reference against another project's Windows ConPTY CI
  suggesting hosted `windows-latest` runners generally can support ConPTY
  successfully, so this doesn't appear to be a blanket capability gap;

the remaining defect is most plausibly a runner/session-specific quirk in
how this specific GH Actions Windows image handles a *second*, independently
created console session/pseudo-console alongside the runner's own (already
redirected, non-interactive) PowerShell host process -- something that would
require live, interactive debugging directly on the runner image (not
available to this agent), or testing on a genuinely interactive Windows
session (a real desktop or an interactive self-hosted runner), to isolate
further.

**This remains an unresolved harness-level blocker, not a RED or GREEN
verdict on the actual #1976 interactive-passphrase-prompt question.** The
self-test gating the 4 real scenarios has still never passed, so the actual
SSH passphrase prompt boundary has never been exercised by this fixture. No
product code has been touched at any point; PR #2242 remains draft and must
not be merged.

### Recommendation

Given the depth of harness-level investigation now exhausted without
resolving the underlying capture defect, further progress on this specific
evidence question likely requires either (a) an interactive/attended Windows
session (not a GH Actions hosted runner) to drive the real ConPTY scenario
and observe it directly, or (b) live debugging tools on the runner image
itself (e.g. attaching a debugger to the child process) that a fully
automated CI-only agent cannot use. I recommend #1976's remaining Windows
boundary be closed out on the strength of the *other* already-confirmed
evidence path (forced `SSH_ASKPASS` supplying the passphrase, known to work)
rather than continuing to sink effort into automating the genuinely
interactive case in a hosted CI runner.

danielmeppiel:
## Update 4: handle-lifecycle fix falsified; independent minimal differential confirms runner-level ConPTY limitation

Per the coordinator's two follow-up instructions, ran two more single-variable tests.

### Test A: documented handle-close lifecycle fix (falsified)

Verified against Microsoft's own docs for "Creating a Pseudoconsole session," which state verbatim:

> Upon completion of the CreateProcess call ... the handles given during creation should be freed from this process. This will decrease the reference count on the underlying device object and allow I/O operations to properly detect a broken channel when the pseudoconsole session closes its copy of the handles.

`ConPtyNative.cs` was closing `inputReadSide`/`outputWriteSide` immediately after `CreatePseudoConsole` succeeded -- **before** `UpdateProcThreadAttribute`/`CreateProcessW` ran at all, deviating from that documented order. Commit `93726671cb` made the single-variable fix: retain both handles through `CreateProcessW` and close them immediately after it returns successfully (catch-block failure cleanup already handled the case correctly, since those variables are no longer nulled early).

Hosted run [`29474766326`](https://github.com/microsoft/apm/actions/runs/29474766326): **falsified**. Identical symptom -- 16-byte VT100 boot sequence, `ResizePseudoConsole`/`Resize` kicks still only surface host bookkeeping, marker text never observed in the captured transcript.

### Test B: independent minimal standalone differential (decisive)

Per the pre-authorized next step, added `scripts/windows/conpty-differential/` (commit `abe4534892`): a **from-scratch, minimal, standalone ConPTY console app** (`MiniTerm.exe`), compiled via `dotnet build` into a real `.exe` and invoked directly by the workflow -- **not** PowerShell `Add-Type` running in-process inside `powershell.exe`. It deliberately excludes every workaround accumulated in `ConPtyNative.cs` this session (DSR responder, resize kicks, process-tree diagnostics) and reuses only the same minimal Win32 call sequence from Microsoft's own reference sample.

Hosted run [`29475020094`](https://github.com/microsoft/apm/actions/runs/29475020094):

```
[*] Differential: running standalone MiniTerm.exe (outside PowerShell Add-Type) with marker MINITERM_DIFF_52304737
MINITERM_DIFF_52304737                                          <- leaks into the raw, un-redirected job log, same as always
[i] MiniTerm: pseudo console 0x1F19AB6FF60, child pid 1828
[i] MiniTerm transcript (16 bytes): ^[[?9001h^[[?1004h          <- captured transcript: ONLY the boot sequence
[x] MiniTerm FAIL: marker NOT observed in transcript (same symptom as fixture)
```

This is decisive: **a completely independent, minimal ConPTY consumer with zero shared code or workarounds shows the identical symptom on this same runner.** This rules out the fixture's own implementation approach (PowerShell hosting, `Add-Type` compilation, or any of the workarounds layered on `ConPtyNative.cs`) as the cause. The defect is not specific to how this fixture drives ConPTY -- it reproduces with the plainest possible Microsoft-sample-equivalent pattern, strongly indicating a runner/environment-specific ConPTY limitation rather than a code defect in this repository's fixture.

### Corrected narrative

To be unambiguous per the coordinator's request: the DSR(6)/`CSI 6n` handshake theory (comment #1) was **tested and disproven** -- it was never claimed as the root cause, only investigated and ruled out with direct evidence (no `CSI 6n` ever observed). The handle-lifecycle fix (this comment) was also **tested and falsified**. Neither is the root cause. The process-tree diagnostic (comment #1) and both resize experiments (comments #2-#3) remain accurate as originally stated.

### Updated recommendation

With two independent implementations (this fixture's PowerShell/Add-Type-hosted harness, and a from-scratch standalone `.exe`) both showing the exact same symptom -- child process demonstrably running and producing output (which leaks into the raw job log every time), yet that output never lands in either implementation's pseudo-console buffer -- further debugging from inside a CI-only agent is unlikely to be productive. This now looks like a genuine limitation of the specific hosted `windows-latest` runner environment/session context for ConPTY consumers, not a fixable code defect in this repository.

Recommend closing out #1976's remaining Windows boundary on the strength of the already-confirmed forced-`SSH_ASKPASS` evidence path (known-working, not being retested), rather than continuing to sink effort into automating the interactive case inside hosted CI. If the interactive case remains a priority, it would need either a genuinely interactive/attended Windows session (real desktop, or an interactive self-hosted runner) or lower-level tracing tools (ETW, Sysinternals) unavailable to a CI-only agent.

No product code changed. This PR remains **draft** and will **not be merged**.

danielmeppiel:
Closing this draft **without merge** as an evidence-only lane.

Hosted Windows could not yield a trustworthy RED/GREEN product verdict: both the PowerShell/Add-Type fixture and an independent standalone .NET ConPTY host reproduced the same harness-level signature (child marker escaped to the Actions log while the private ConPTY transcript contained only the 16-byte startup sequence). The documented handle-lifetime correction, DSR response, and resize probes did not resolve it. No product code changed, no SSH scenario passed the self-test gate, and no #1976 fix is justified from this lane.

The run history and comments remain preserved here. The remaining prerequisite is an attended Windows desktop or interactive self-hosted runner that can exercise packaged `apm.exe` -> Git for Windows -> Windows OpenSSH with an encrypted key and no askpass helper.

PR #2269, opened 2026-07-16
fix(deps): preserve shared transitive deps still reachable from survivors

danielmeppiel:
## TL;DR

`apm uninstall` used `LockedDependency.resolved_by` — a single-valued, first-wins field set at install time — as an implicit reachability oracle when deciding which transitive dependencies to garbage-collect. In a diamond shape (two direct packages both declaring the same transitive dependency), removing whichever direct package happened to install first deleted the shared transitive dependency from the lockfile and `apm_modules/`, even though the surviving direct package still declares and needs it. This adds a single canonical forward-reachability owner (`src/apm_cli/deps/reachability.py`) that recomputes reachability on demand from the real manifest/lock graph, with an explicit fail-closed completeness contract and a `resolved_by`/`local_path` repair so a rescued dependency doesn't silently become un-collectable forever.

> [!NOTE]
> No matching open issue or PR existed for this defect. Follows up on config-lifecycle session `87c17351-ebf2-4812-b7bf-b93ec1b20401`, commit `13f7e54b2e50501ba1e497f571b7601a79266c55` (empirical RED discovery) — that branch is **not** the basis of this PR and was **not** cherry-picked; the defect is independently reproduced below with a smaller CLI integration contract.

## Problem (WHY)

- [x] `apm uninstall ../root-a` on a diamond (`root-a`, `root-b` both declaring local `../shared`) deletes `_local/shared` from the lockfile and `apm_modules/`, even though `root-b` still declares and reaches it — reproduced on unmodified `main` via `click.testing.CliRunner` against the real `apm install`/`apm uninstall` commands.
- [x] Root cause: `_build_children_index()` in `src/apm_cli/commands/uninstall/engine.py` is a *backward* index keyed on `LockedDependency.resolved_by` (`src/apm_cli/deps/lockfile.py`), which only ever records the first parent that introduced a dependency — never every parent that legitimately depends on it.
- [!] A latent second defect only becomes observable once the first is fixed: rescuing `shared` without repairing its now-stale `resolved_by` (still pointing at the removed parent) means the backward index can never again treat `shared` as a candidate keyed on the real surviving parent, and `path_anchoring.resolve_local_dep_dir` raises `LocalResolutionError` on the dangling chain — degrading the fix from "occasionally under-deletes" into "permanently never garbage-collects once rescued."
[...]

PR #2266, opened 2026-07-16
test(integration): add primitive target lifecycle covering array

danielmeppiel:
# add(integration): cover primitive target lifecycle

## TL;DR

Add a hermetic covering array that drives the built CLI through install, reinstall, target widening and narrowing, prune, and uninstall across deployable primitive categories. The strict post-prune ledger equality assertion remains unchanged. Current main and the accepted #2275 hook-target reconciliation are direct merge dependencies, so once #2275 lands the effective PR diff is three test-only files.

> [!NOTE]
> Human merge only after #2275 lands; its dependency ancestry collapses the implementation and sidecar-cleanup work from this PR.

## Problem (WHY)

- [x] Target lifecycle regressions span project/global scope, harness-specific layouts, and target contraction; a single happy-path test cannot prove those boundaries together.
- [x] A post-narrow prune must keep a strict equality oracle for the remaining deployment-ledger targets rather than a permissive subset check.
- [!] Hook/sidecar target-contraction behavior belongs to #2275, which is merged as a dependency rather than copied into this PR.

Why these matter: ["Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/)

## Approach (WHAT)

| # | Fix |
|---|---|
| 1 | Add one sparse, valid primitive/target covering array using local Git fixtures and the installed CLI. |
| 2 | Preserve exact post-prune ledger-target equality and run the dropped-target hook/sidecar dependency coverage. |
| 3 | Register the module in the finite e2e taxonomy. |

## Implementation (HOW)

- **`tests/integration/test_primitive_target_covering_array.py`** -- adds the hermetic lifecycle matrix. It reads compatibility from `KNOWN_TARGETS`, exercises refused dynamic targets without normalizing their failure, and retains the strict post-narrow prune assertion.
- **`tests/quality/critical_suite.toml`** -- registers the new module as an e2e critical-suite member.
- **`tests/quality/test_test_taxonomy.py`** -- advances the finite manifest count to 20.

## Diagrams

Legend: each row drives the same installed-CLI lifecycle; target narrowing is checked by the strict ledger oracle before cleanup.

```mermaid
flowchart LR
    A["Covering-array row"] --> B["Install with local Git fixture"]
    B --> C["Reinstall: semantic state unchanged"]
    C --> D["Optional target widen"]
    D --> E["Narrow targets"]
    E --> F["Install then prune"]
[...]

Issue #2250, opened 2026-07-16
Hook wipe scans all merge targets but rebuild only re-populates canonical_targets (uninstall + prune)


danielmeppiel:
### Summary

`HookIntegrator.sync_integration(managed_files=set())` wipes `_apm_source`-tagged entries and deletes the `apm-hooks.json` sidecar across **all** `KNOWN_TARGETS` (it defaults `targets` to `list(KNOWN_TARGETS.values())` when no explicit `targets` is passed -- see `src/apm_cli/integration/hook_integrator.py`'s `sync_integration()`). Both existing "clear + rebuild" callers -- `apm uninstall`'s `_sync_integrations_after_uninstall()` (`src/apm_cli/commands/uninstall/engine.py`) and the new `reconcile_after_removal()` added in #2249 -- only *rebuild* hooks for `resolve_targets(explicit_target=apm_package.canonical_targets or None)`, i.e. scoped to the package's currently-declared `targets:` list in `apm.yml`.

If a project's `targets:` list is narrowed (e.g. `targets: [claude, cursor]` -> `targets: [claude]`) while a dependency stays installed and its old harness's hook file (e.g. `.cursor/hooks.json`) still exists on disk, the next `apm uninstall` or `apm prune` run wipes that harness's entire `_apm_source`-tagged entry set and deletes its `apm-hooks.json` sidecar -- **including entries for packages that are still declared and still installed** -- because the rebuild phase never targets that now-undeclared harness.

### Reproduction (confirmed during #2249 review)

1. `apm install` two packages with merged hooks under `targets: [claude, cursor]`.
2. Edit `apm.yml` to `targets: [claude]` (drop `cursor`), leaving both dependencies installed.
3. Remove one (unrelated) dependency and run `apm prune` (or `apm uninstall <other-dep>`).
4. Result: `.cursor/hooks.json` loses **both** packages' `_apm_source` entries (not just the pruned/uninstalled one), and `.cursor/apm-hooks.json` is deleted entirely -- silently, with no error logged, since this is treated as "nothing to rebuild" rather than a loss.

### Scope note

[...]

Issue #2059, opened 2026-07-06
[BUG] Lockfile deployed_files records inactive-target paths from package-declared targets; audit deployed-files-present unfixable on fresh checkouts


Cianai:
**Describe the bug**

When a dependency package declares `targets:` in its own `apm.yml` that include harnesses the consumer has NOT activated, the consumer's `apm.lock.yaml` records `deployed_files` entries for those inactive targets — but `apm install`/`apm update` (correctly) never writes them. `apm audit --ci` then fails `deployed-files-present` on any fresh checkout, permanently.

Example: `microsoft/azure-skills`, `microsoft/skills-for-copilot-studio`, and `microsoft/skills-for-fabric` declare windsurf among their package targets. A consumer with:

```yaml
targets:
- claude
- codex
- copilot
- cursor
- gemini
```

gets 67 `.windsurf/skills/...` entries in the lock's `deployed_files` (24 + 33 + 10) that never exist on disk:

```
│ deployed-files-present │ 67 deployed file(s) missing -- run 'apm install' to restore │
```

Running `apm install` does not restore them (windsurf is not an active target, so nothing writes them) — the message loops.

**To Reproduce**

1. Consumer `apm.yml` with the 5 targets above and `dependencies.apm: [microsoft/azure-skills]`.
2. `apm update --yes microsoft/azure-skills` (or `apm install` on first resolve).
3. Inspect the lock: `.windsurf/...` paths appear under the dep's `deployed_files`; no `.windsurf/` directory exists.
4. `apm audit --ci --no-fail-fast` → `deployed-files-present` fails with those paths.

Also tried: consumer-side per-dep restriction

```yaml
- git: https://github.com/microsoft/azure-skills.git
  targets: [claude, codex, copilot, cursor, gemini]
```

— fresh re-resolution still records the `.windsurf/` entries.

A likely-related symptom: the `drift` check's cache replay produces different bytes than the install write path for ~100 of the same packages' files (fabric/copilot-studio SKILL.mds), so `drift` also fails on a state that install itself just wrote. Both smell like the same root: lock/replay bookkeeping uses package-declared targets where install uses the active intersection.

**Expected behavior**

`deployed_files` in the lock should record the files actually deployed for the **active install-target intersection** (the same set install writes), so `apm audit` is self-consistent with `apm install` output on a fresh clone.

**Environment**

- OS: macOS (darwin-arm64)
[...]

PR #2275, opened 2026-07-17
fix(install): reconcile dropped-target merge-hook state on target contraction

### LINKED ISSUE #2253: No diagnostic when apm uninstall/prune leaves stale hook entries in a dropped-target harness
## Context

Follow-up from the apm-review-panel on #2252 (fix for #2250: hook-wipe target-scope asymmetry).

After #2252, `apm uninstall`/`apm prune` correctly scope the merged-hook JSON wipe to the project's currently-resolved `targets:` set (matching the rebuild scope). This is the correct fix for #2250 — a still-declared sibling package's hooks in a dropped-target harness (e.g. `.cursor/hooks.json` when `cursor` is removed from `targets:`) are no longer destroyed.

However, this also means a **removed package's own** merged-hook entries in a dropped-target harness are left behind as stale/orphaned JSON, with **zero user-visible signal**. Concretely:

- The harness is outside the project's declared `targets:` scope, so neither `apm install` nor `apm uninstall`/`apm prune` will ever touch it again (by design — this is the whole point of #2250's fix).
- The existing error-path guidance in `prune.py` ("Hook reconciliation failed... run 'apm install' to rebuild hook configuration.") is subtly incorrect for this specific scenario: `apm install` itself only ever rebuilds within the currently-resolved target scope, so it would never clean up a dropped-harness's stale entries either.

## Impact

Low severity, not a regression (pre-#2250-fix behavior was strictly worse — ALL hooks were wiped indiscriminately). But a user who drops a target from `targets:` gets no indication that stale hook entries remain in that harness's config file, potentially causing confusing behavior if they later inspect or manually edit that file.

## Suggested follow-up

Add an info-level diagnostic (e.g. during `apm prune`/`apm uninstall` or `apm audit`) when a merged-hook harness outside the resolved `targets:` scope still contains `_apm_source`-tagged entries, pointing the user at how to actually clean it up (e.g. re-add the target temporarily, or a dedicated cleanup command/flag).

## Evidence trail

Raised by the `devx-ux-expert` persona during the manually-executed apm-review-panel for #2252, cross-referenced against `src/apm_cli/commands/prune.py`'s diagnostic wording and `src/apm_cli/integration/hook_integrator.py`'s scoping logic.

danielmeppiel:
# fix(install): reconcile dropped-target merge-hook state on target contraction

## TL;DR

Narrowing `apm.yml`'s `targets:` (e.g. `[claude, codex]` -> `[claude]`) never cleaned up the dropped target's APM-owned merge-hook JSON config + `apm-hooks.json` ownership sidecar, even after `apm install` + `apm prune` -- the residue survived forever. This PR extends the two existing canonical owners (`manifest_reconcile.py` and `HookIntegrator`) with one new method each, rather than inventing a third authority, and fails closed on any malformed/partial state instead of silently swallowing it.

> [!NOTE]
> Closes #2253. Evidence-only relationship to #2266's `fixture-claude-codex-hook-narrow` covering-array row (see Validation) -- that PR is neither a base nor modified here.

## Problem (WHY)

- [x] `apm install --no-policy` with `targets: [claude, codex]` deploys `.claude/*` and `.codex/*` merge-hook JSON, each entry tagged `_apm_source`. Narrowing to `targets: [claude]` and re-running `apm install` + `apm prune` leaves `.codex/hooks.json` and `.codex/apm-hooks.json` fully intact with the marker still present.
- [x] `HookIntegrator.sync_integration`/`reconcile_after_removal` are intentionally scoped (#2250) to the SAME resolved `targets` the rebuild loop uses -- correct for that bug, but it permanently walls prune/uninstall off from a target dropped from `targets:` entirely, since the rebuild loop never touches it again.
- [x] Merge-hook config files are deliberately excluded from `deployed_files` tracking (`hook_integrator.py`, "Don't track the config file in target_paths -- it's a shared file"), so the generic path-list reconciliation in `manifest_reconcile.py` can never see this state either.
- [!] #2253 (open) already documented this exact gap but proposed only a diagnostic, not real cleanup.

Why these matter: this breaks the target-contraction promise every other primitive type already keeps (`update.md`'s existing "Target contraction is reconciled" bullet), and leaves stale, APM-attributed state on disk that a user has no supported way to remove short of manual deletion.

## Approach (WHAT)

| # | Fix (and why, if non-obvious) |
|---|-------------------------------|
| 1 | `manifest_reconcile.py` gains `reconcile_dropped_merge_hook_targets()`, reusing the existing "allowed = active union declared" detection semantics (preserves the `declared_targets=None` legacy no-op, #2059 symmetry). |
| 2 | `hook_integrator.py` gains `HookIntegrator.reconcile_dropped_targets()`, the canonical owner of cleanup for a target dropped entirely from `targets:` -- fails closed on malformed/partial JSON instead of the best-effort posture kept by existing prune/uninstall callers. |
| 3 | Wired at the *top* of `LockfileBuilder.build_and_save()` (fires on every non-`lockfile_only` install, independent of whether any package installed/orphaned this run) and at the end of `manifest_reconcile.reconcile_deployed_state()` (`apm compile`/`apm update`). |
| 4 | New semantic AST lint guard (`check_hook_config_write_owner.py`, wired as AC15) rejects any write-mutating call on a merge-hook path composed outside `hook_integrator.py`, plus a lexical guard keeping `prune`/`uninstall` from calling the new APIs directly (#2250 scope). |

## Implementation (HOW)

- **`src/apm_cli/install/manifest_reconcile.py`** -- new `reconcile_dropped_merge_hook_targets(project_root, active_targets, declared_targets, *, user_scope=False)`; `reconcile_deployed_state`/`reconcile_project_deployed_state` now thread `user_scope` and call it once at the end.
- **`src/apm_cli/install/phases/lockfile.py`** -- new `_reconcile_dropped_merge_hook_targets()` helper called at the top of `build_and_save()`, guarded by `not self.ctx.lockfile_only`; unreachable under `--dry-run` (that path never calls `build_and_save()` at all).
- **`src/apm_cli/integration/hook_integrator.py`** -- new public `reconcile_dropped_targets()`. Its docstring documents the canonical-owner rationale and delegates immediately to a private sibling module; the file's existing `_clean_apm_entries_from_json` and `sync_integration`/`reconcile_after_removal` are otherwise untouched (`reconcile_after_removal`'s own docstring was condensed, prose-only, purely to make room under the repo's 2100-line file guardrail after adding the wrapper -- no logic changed).
- **`src/apm_cli/integration/_hook_dropped_targets.py`** (new) -- the actual implementation of `reconcile_dropped_targets`, including fail-closed sidecar-only-orphan handling. This is **not** a second target-cleanup authority: it is invoked exclusively from `HookIntegrator.reconcile_dropped_targets`, and every JSON mutation still routes through the unchanged `HookIntegrator._clean_apm_entries_from_json` primitive. It exists only because `hook_integrator.py` had ~0 lines of budget left under the CI file-length gate.
- **`scripts/check_hook_config_write_owner.py`** (new) + **`scripts/lint-architecture-boundaries.sh`** (AC15) -- semantic AST guard (bounded one-hop alias resolution) flagging `open`/`.open`/`.write_text`/`.write_bytes`/`.unlink` on a composed merge-hook path anywhere outside `hook_integrator.py`; plus the lexical prune/uninstall exclusion.
- **`tests/unit/install/test_manifest_reconcile_dropped_hook_targets.py`**, **`tests/unit/integration/test_hook_integrator_reconcile_dropped_targets.py`**, **`tests/integration/test_hook_target_contraction_reconciliation.py`** (all new) -- unit, component, and CLI-level integration-contract coverage (see Scenario Evidence).
- **`tests/integration/test_architecture_authorities.py`** -- 3 new tests: two sandbox-copy mutation twins proving the AST/lexical guards actually reject a synthetic bypass, one positive control proving the checker doesn't false-positive on the real post-fix tree.
- **Docs** (`install.md`, `update.md`, `compile.md`, `prune.md`, `packages/apm-guide/.apm/skills/apm-usage/commands.md`) + **`CHANGELOG.md`** -- in-place edits documenting the new reconciliation behavior per command.
- **`docs/src/content/docs/specs/openapm-v0.1.md`** + **`docs/public/specs/manifests/openapm-v0.1.requirements.yml`** + **`CONFORMANCE.{json,md}`** -- new `req-lk-021` (Section 5.2, consumer MUST), extending `req-lk-020`'s preserve/remove decision to merge-based hook configuration; `req-lk-020` itself only reaches `deployed_files`/`local_deployed_files`, which merge-hook JSON is deliberately excluded from. This closed the repo's own Mode B (silent-extension) CI gate, which correctly fired on the first push of this PR: real new normative behavior under a critical path (`install`/`integration`) with no spec citation. See `tests/spec_conformance/test_lockfile_reqs.py::test_dropped_target_merge_hook_state_reconciled_fail_safe`.

## Diagrams

Legend: call flow for `apm install` when `targets:` has been narrowed -- each existing owner gains one new method (dashed border = new), and reconciliation fires before the early-return so it never depends on whether a package happened to install/orphan this run.

```mermaid
sequenceDiagram
    participant LB as LockfileBuilder.build_and_save
    participant MR as manifest_reconcile
    participant HI as HookIntegrator
    participant FS as .codex/*.json

    CLI->>LB: build_and_save()
    rect rgb(255, 247, 200)
        Note over LB,MR: NEW: fires before the early-return, independent of installed_packages
        LB->>MR: reconcile_dropped_merge_hook_targets(active, declared)
        MR->>MR: dropped = KNOWN_TARGETS - (active | declared)
        MR->>HI: reconcile_dropped_targets(project_root, dropped)
        HI->>HI: delegate to _hook_dropped_targets (private, line-budget split)
        HI->>FS: _clean_apm_entries_from_json(json_path, sidecar_path)
        FS-->>HI: entries/sidecar removed, or fails-closed with warning
    end
    HI-->>MR: stats
    MR-->>LB: changed
    LB-->>CLI: lockfile saved
```

Legend: rejected fix sites versus the chosen path -- `apm prune`/`apm uninstall` are kept out on purpose, per #2250's target-scope contract, and enforced by the new AC15 lint guard.

```mermaid
flowchart LR
    subgraph Rejected["Rejected fix sites"]
        R1["commands/prune.py<br/>ad hoc .codex delete"]:::bad
        R2["HookIntegrator.reconcile_after_removal<br/>#2250-scoped, walled off by design"]:::bad
    end
    subgraph Chosen["Chosen: extend existing owners"]
        C1["manifest_reconcile.py<br/>reconcile_dropped_merge_hook_targets"]:::new
        C2["hook_integrator.py<br/>reconcile_dropped_targets"]:::new
    end
    C1 --> C2
    C2 --> C3["_hook_dropped_targets.py<br/>private impl, line-budget split only"]:::new
    classDef bad stroke:#c00,stroke-dasharray: 3 3;
```

## Trade-offs

- **Split `HookIntegrator.reconcile_dropped_targets`'s implementation into a private sibling module.** `hook_integrator.py` was already at 2096/2100 lines against the repo's CI file-length guardrail before this change; adding a full public method with docstring left ~0 lines of room. Rejected trimming ~100 lines of unrelated pre-existing docstrings as riskier than a thin, delegate-only public wrapper backed by a private implementation module that is invoked from exactly one call site.
- **Rejected fixing this in `commands/prune.py`.** Would bypass `HookIntegrator`, risking deletion of user-authored entries and creating a second, ad hoc authority over merge-hook JSON mutation -- violates the one-canonical-owner rule.
- **Rejected fixing this by widening `reconcile_after_removal`'s target scope.** That method's #2250 scoping is load-bearing (its own mutation-break test, `test_scope_symmetry_is_load_bearing_mutation_break`, proves widening it regresses the original collateral-wipe bug).
- **Fail-closed on malformed JSON, not best-effort.** Existing prune/uninstall callers of `_clean_apm_entries_from_json` keep their unchanged best-effort posture; the new dropped-target path is stricter (leaves files byte-identical + logs a warning) because it runs unattended on every install, not on an explicit user-invoked cleanup command.
- **`#2266`'s fixture is evidence-only.** Reproduced independently on current `main` rather than basing this branch on or cherry-picking from #2266's scenario lane; that PR's own committed assertion for `fixture-claude-codex-hook-narrow` is intentionally left untouched here.
- **Spec citation via a new requirement, not a waiver.** The Mode B gate offers an `apm-spec-waiver: <rationale>` bypass, but that's reserved for changes with "no observable behaviour delta" -- this PR has real, user-observable new behavior, so using the waiver would misrepresent the change. Added `req-lk-021` as the honest path instead: an extension of the existing `req-lk-020` family (same Section 5.2, same preserve/remove decision), not a new requirement family.

## Benefits

1. `.codex/hooks.json` + `.codex/apm-hooks.json` (and the equivalent for any of the 6 merge-hook targets) are removed within one `apm install`/`apm compile`/`apm update` after a target is dropped from `targets:` -- no manual cleanup required.
2. Zero regressions: the full pre-existing `test_hook_wipe_target_scope_e2e.py` (8 tests) and `test_prune_hook_reconciliation_e2e.py` (7 tests) suites, plus 1838 `tests/unit/install/` tests, pass unchanged.
3. O(k) cost per run, k = 6 (the fixed `_MERGE_HOOK_TARGETS` catalog) -- independent of dependency graph size.
4. A new semantic AST guard makes any future bypass of `HookIntegrator`'s merge-hook write ownership fail CI, not just this PR's own code.

## Validation

CI-mirror lint chain, run in full:

```
$ uv run --extra dev ruff check src/ tests/
All checks passed!

$ uv run --extra dev ruff format --check src/ tests/
1547 files already formatted

    --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/

$ bash scripts/lint-auth-signals.sh
[+] auth-signal lint clean

$ bash scripts/lint-architecture-boundaries.sh
[*] AC15: hook target-contraction cleanup authority
[+] architecture boundary lint clean

$ find src/ -name '*.py' -print0 | xargs -0 -I{} awk -v max=2100 \
    'END { if (NR > max) printf "%s: %d lines (max %d)\n", FILENAME, NR, max }' {}
(no output -- hook_integrator.py is 2094 lines)

$ uv run python -m tests.spec_conformance.orphan_check
[+] orphan_check OK: 101 requirements aligned across anchors / manifest / Appendix C / pytest markers

$ BASE_REF=origin/main bash tests/spec_conformance/mode_b_detector.sh
[+] mode_b: spec-concurrent edit detected; orphan_check owns this PR

$ uv run --extra dev pytest tests/spec_conformance/ -q
141 passed, 2 skipped
```

<details><summary>Full pytest output (80 tests: new/modified files + full hook e2e regression suites)</summary>

```
tests/unit/install/test_manifest_reconcile_dropped_hook_targets.py .... [  3 tests]
tests/unit/integration/test_hook_integrator_reconcile_dropped_targets.py .......... [ 8 tests]
tests/integration/test_hook_target_contraction_reconciliation.py ...... [ 12 tests]
tests/integration/test_architecture_authorities.py (40 tests, incl. 3 new guard tests)
tests/integration/test_hook_wipe_target_scope_e2e.py (8 tests, unmodified)
tests/integration/test_prune_hook_reconciliation_e2e.py (7 tests, unmodified)

======================== 80 passed in 244.15s (0:04:04) ========================
```

Also re-ran independently: `tests/unit/install/` (1838 passed), and the broader `tests/integration/ tests/unit/integration/ -k "hook or prune or compile or install"` sweep (2430 passed, 89 skipped; the sole failure, `test_tls_frozen_hook.py::test_user_override_still_wins_under_frozen_hook`, is a pre-existing, order-dependent test that passes cleanly in isolation and is unrelated to this change).

</details>

**Mutation-break check** (manual, uncommitted, performed before this commit, both files cleanly restored): stubbing the `LockfileBuilder.build_and_save()` wiring call to a no-op reverted `test_narrow_then_install_removes_dropped_target_hook_state` to RED with the exact expected assertion failure; restoring it returned the suite to GREEN. Repeated identically for the `manifest_reconcile.reconcile_deployed_state()` call site against `test_compile_reconciles_dropped_target_hook_state`.

**Independent reproduction on current `main`** (before any fix code existed): a real local-Git hook package fixture (Claude + Codex targets) was installed, `targets:` narrowed to `[claude]`, then `apm install --no-policy` + `apm prune` run -- `.codex/apm-hooks.json` survived both commands with its `_apm_source` marker intact, confirming the defect. This mirrors (does not modify, is not based on) PR #2266's `fixture-claude-codex-hook-narrow` covering-array row.

### Scenario Evidence

| 1 | Narrowing `targets:` and running `apm install` removes the dropped target's merge-hook JSON entries and ownership sidecar | Portability by manifest, Secure by default | `tests/integration/test_hook_target_contraction_reconciliation.py::test_narrow_then_install_removes_dropped_target_hook_state` (regression-trap for #2253) | integration |
| 2 | The retained target's own hook state is untouched by the same narrow+install | Portability by manifest | `tests/integration/test_hook_target_contraction_reconciliation.py::test_narrow_then_install_preserves_retained_target_state` | integration |
| 3 | A user's own hand-authored `.codex/hooks.json` entries survive narrow+install (APM only removes its own entries) | Secure by default | `tests/integration/test_hook_target_contraction_reconciliation.py::test_narrow_then_install_preserves_user_owned_codex_entries` | integration |
| 4 | `apm prune` alone (no target narrowing) still does not clean a dropped target's hook state -- #2250/#2252's contract is preserved | Governed by policy | `tests/integration/test_hook_target_contraction_reconciliation.py::test_prune_alone_still_does_not_clean_dropped_target_state`<br/>`tests/integration/test_hook_wipe_target_scope_e2e.py::test_scope_symmetry_is_load_bearing_mutation_break` | integration |
| 5 | `apm install --dry-run` never mutates hook state, even after a narrow | Secure by default, DevX | `tests/integration/test_hook_target_contraction_reconciliation.py::test_dry_run_install_does_not_reconcile_dropped_target_state` | integration |
| 6 | `apm compile`/`apm update` reconcile dropped-target hook state identically to `apm install` | Multi-harness support | `tests/integration/test_hook_target_contraction_reconciliation.py::test_compile_reconciles_dropped_target_hook_state` | integration |
| 7 | A one-off `apm install --target claude` (transient override) does not clean codex's hook state for a package still declared for both targets in `apm.yml` | Governed by policy | `tests/integration/test_hook_target_contraction_reconciliation.py::test_explicit_transient_target_does_not_clean_other_declared_target` | integration |
| 8 | Malformed native/sidecar hook JSON is left byte-identical (never destructively rewritten) with a warning, not silently repaired or deleted | Secure by default | `tests/unit/integration/test_hook_integrator_reconcile_dropped_targets.py::TestReconcileDroppedTargetsMalformedNative::test_malformed_native_json_fails_closed_with_warning`<br/>`TestReconcileDroppedTargetsSidecarOnlyOrphan::test_malformed_sidecar_only_orphan_fails_closed` | unit |
| 9 | Any future write to a merge-hook config/sidecar path composed outside `HookIntegrator` fails CI, not just this PR's own code | Secure by default, OSS / community-driven | `tests/integration/test_architecture_authorities.py::test_hook_config_write_guard_rejects_composed_path_outside_hook_integrator`<br/>`test_hook_ownership_guard_rejects_prune_calling_contraction_api` | integration |

### Update: merged `main` + added permanent Cursor coverage (required-gate "node-2" evidence)

Merged current `main` (`fa71f950c9`) additively (no rebase, no force-push),
resolving two conflicts in files touched by both branches: `scripts/
lint-architecture-boundaries.sh` and `tests/integration/
test_architecture_authorities.py`, both against `main`'s new AC15
("post-uninstall reachability owner authority", #2269) -- renumbered
`main`'s incoming rule to AC16 so it coexists with this PR's existing AC15
("hook target-contraction cleanup authority") without either side losing
coverage. Both AC15 and AC16 pass; full related test suite (243 tests) and
the CI-mirror lint chain were re-run clean after the merge.

Also added permanent test coverage proving the fix's `HookIntegrator.
reconcile_dropped_targets` generalizes to Cursor, not just Codex/Claude --
per an explicit evidence request describing a real widen-then-narrow
(Claude -> Claude+Cursor -> Claude) run where `.cursor/hooks.json` and its
`apm-hooks.json` sidecar previously remained byte-for-byte from the widened
install. Direct reproduction confirmed the existing fix already handles
this correctly (Cursor is registered in `_MERGE_HOOK_TARGETS` with the same
`schema_strict=True` default as Codex), so **no production code changed**
for this update -- only new tests, asserting the native merged config and
sidecar directly rather than relying on audit:

| # | Scenario | Test(s) | Type |
|---|----------|---------|------|
| 10 | Widening `targets:` to include `cursor` then narrowing back removes the dropped target's merge-hook JSON entry and ownership sidecar | `tests/integration/test_hook_target_contraction_reconciliation.py::test_widen_then_narrow_removes_dropped_cursor_hook_state` | integration |
| 11 | A hand-authored, unowned `.cursor/hooks.json` entry survives the same widen-then-narrow cleanup | `tests/integration/test_hook_target_contraction_reconciliation.py::test_widen_then_narrow_preserves_user_owned_cursor_entries` | integration |
| 12 | Direct-call proof mirroring the exact reported evidence shape (owned `PreToolUse` entry + user-authored sibling entry + `_apm_source` sidecar, retained Claude state untouched) | `tests/unit/integration/test_hook_integrator_reconcile_dropped_targets.py::TestReconcileDroppedTargetsCursor::test_cursor_native_and_sidecar_cleaned_preserving_user_entry` | unit |

Out of scope for this update (owned separately): the `.agents/skills`
`target=agents` orphan-row defect.

## How to test

- [ ] Build the `fixture-org/hook-narrow-pkg` local-Git package fixture with Claude + Codex hook targets (see `tests/integration/test_hook_target_contraction_reconciliation.py` for the fixture shape), install with `targets: [claude, codex]`, then narrow to `[claude]` and re-run `apm install` -> observe `.codex/hooks.json` and `.codex/apm-hooks.json` no longer exist.
- [ ] Run `uv run python -m pytest tests/integration/test_hook_target_contraction_reconciliation.py -v` -> all 14 pass (12 original + 2 Cursor).
- [ ] Run `bash scripts/lint-architecture-boundaries.sh` -> AC15 passes; inject a raw `.codex/hooks.json` write into `manifest_reconcile.py` and re-run -> AC15 fails with the "must stay owned by HookIntegrator" message.
- [ ] Run `uv run python -m pytest tests/integration/test_hook_wipe_target_scope_e2e.py tests/integration/test_prune_hook_reconciliation_e2e.py -v` -> all 15 pass unchanged (no #2250/#2252 regression).


danielmeppiel:

> Closes a real orphaned-config gap when users switch AI assistants; spec and CHANGELOG text need a tightening pass before release but nothing blocks merge.

All eight active panelists converge on ship. Zero blocking findings, zero severity disagreements, and the test-coverage-expert confirms the critical surfaces are regression-trapped: 12 integration + 13 unit + 3 architecture-guard + 1 spec-conformance tests all pass (`tests/integration/test_hook_target_contraction_reconciliation.py` -- 12 passed in 4.65s). The three-layer canonical-owner design (manifest_reconcile decides WHICH targets dropped, HookIntegrator decides HOW to clean, `_clean_apm_entries_from_json` is the JSON primitive) is validated by python-architect as proportionate and sound. Supply-chain-security confirms all path components are hardcoded constants and the fail-closed posture is genuine. Performance impact is negligible (~0.3ms steady-state from ~5 stat() calls, O(k) with k=6 constant).

The highest-signal follow-ups cluster around the spec text and release communication, not the code. Doc-writer raises two substantive gaps in req-lk-021: the sidecar-only-orphan cleanup case (where the native document is already absent) is not textually covered by the precondition clause, and the fail-closed-on-malformed-JSON guarantee -- which the PR body treats as a headline differentiator with two dedicated test scenarios -- is not codified in the requirement text despite APM's pattern of writing such guarantees directly into spec requirements (see req-tg-006, req-lk-013). These are worth an in-PR tightening pass because req-lk-021 is a new spec artifact landing for the first time; getting the text right now is cheaper than amending it post-release. Separately, both oss-growth-hacker and doc-writer independently flag the CHANGELOG entry as too verbose and internally-voiced (~125 words, 4 PR/issue xrefs, 2 class::method names, a validation-methodology sentence). The CHANGELOG is a public-facing surface; tighten to user-impact-first before the release cut.

The cli-logging-expert's recommended finding about dual-channel `_rich_warning` is valid -- warnings that fire during `apm install` (a forward command) should reach the user's console, not just the structured log. This is consistent with the existing forward-command pattern in the same file and is a small fix. Weighted below the spec-text items because the structured log still captures the information; it's a UX polish gap, not a silent failure.

**Dissent.** No actual dissent between panelists on severity or classification. The CHANGELOG finding appears from two independent panelists (oss-growth-hacker at recommended, doc-writer at nit) -- this is convergence, not disagreement; both say the same thing (too verbose, too internal). Treated as recommended given the CHANGELOG's public-facing role.

**Aligned with:** secure by default (fail-closed on malformed JSON is genuine; new AST lint guard adds defense-in-depth restricting write access on merge-hook config paths to the canonical owner), portable by manifest (req-lk-021 codifies target-contraction cleanup in the OpenAPM spec so implementations converge on the same reconciliation contract across harnesses), governed by policy (the new AST lint guard enforces single-owner write access at CI time -- governance by static analysis, not convention), pragmatic as npm (target-contraction cleanup on install matches the npm mental model of "install fixes everything"; DevX-UX confirms the analogy holds and dry-run inertness is correct).

**Growth signal.** "apm now cleans up after itself when you switch AI assistants" is a strong, user-centric story angle worth amplifying in the release narrative. Pairing it with other hook-lifecycle fixes from this cycle for a before/after in the release announcement is a good idea, provided the CHANGELOG entry is tightened to match this framing first.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 2 | Private module split preserves single-owner API surface; AST guard is sound with one documented narrow blind spot; ship. |
| CLI Logging Expert | 0 | 1 | 2 | Warning messages follow existing reconcile-phase conventions (_log.warning only); recommend adding _rich_warning parallel so malformed-state warnings reach the user's console. |
| DevX UX Expert | 0 | 0 | 3 | Target-contraction cleanup on install/compile/update matches the npm mental model; dry-run inertness is correct; docs clearly document the prune boundary; warning messages name the file and the problem but could add a concrete recovery action. |
| Supply Chain Security Expert | 0 | 0 | 2 | Path composition is inherently safe (all components from hardcoded constants); marker-based ownership is sound; fail-closed behavior is genuine. Two nits on the AST lint guard's coverage gaps. |
| OSS Growth Hacker | 0 | 1 | 1 | CHANGELOG entry is maintainer-voiced (mid-text PR xrefs, class names, test methodology); docs additions are clear and well-placed. Strong story angle for release narrative. |
| Doc Writer | 0 | 4 | 1 | req-lk-021 mirrors req-lk-020's voice/structure well, but its trigger clause misses the sidecar-only-orphan and fail-closed-on-malformed guarantees the code actually ships and tests; three CLI pages now duplicate the same hook-path parenthetical verbatim with no offsetting cut. |
| Test Coverage Expert | 0 | 0 | 1 | All critical surfaces have regression traps; 12 integration + 13 unit + 3 architecture-guard + 1 spec-conformance tests cover the bug fix, negative twins, malformed-state fail-closed, dry-run, and prune/#2250 boundary. Ship. |
| Performance Expert | 0 | 0 | 2 | O(k) reconciliation with k=6 constant; ~0.3ms steady-state overhead from ~5 stat() calls; AST checker is O(N) single-pass at ~3.7s/CI run; no regressions. |

### Top 5 follow-ups

1. **[Doc Writer]** req-lk-021 precondition does not textually cover the sidecar-only-orphan cleanup case (native document already absent, only sidecar remains) -- This is a real, tested, newly-added behavior (with dedicated test scenarios) that the spec requirement meant to cite is silent on. Fixing now is cheaper than amending post-release; the precondition clause needs one sentence covering the "native absent, sidecar orphaned" trigger.
2. **[Doc Writer]** req-lk-021 does not codify the fail-closed-on-malformed-JSON guarantee the PR treats as a headline differentiator -- The PR body names this posture explicitly and ships two dedicated test scenarios for it. APM's own spec pattern (req-tg-006, req-lk-013) writes such guarantees into requirement text. Omitting it from a brand-new requirement landing for the first time is a missed opportunity to lock the contract.
3. **[OSS Growth Hacker]** CHANGELOG entry needs tightening: drop internal class names, PR-review provenance, and validation methodology; lead with user impact -- Both oss-growth-hacker and doc-writer independently flagged this. The CHANGELOG is a public surface; ~125 words with 4 PR xrefs and 2 class::method names reads as internal archaeology, not a release note. Tighten to one user-impact sentence + one closing PR ref.
4. **[CLI Logging Expert]** Add _rich_warning parallel for fail-closed warnings so they reach the user's console during apm install/compile/update -- Warnings currently emit to the structured logging channel only (_log.warning); default console output shows nothing. Since these fire on forward commands (install, not just prune), user-facing channel is warranted per the existing dual-channel pattern in hook_integrator.py.
5. **[Doc Writer]** Deduplicate the merge-hook path/sidecar parenthetical across install.md, update.md, and compile.md; single-source in hooks-and-commands.md -- ~130 net words added to three CLI pages with no compensating cut. hooks-and-commands.md already has a canonical section; shorten each CLI page to a clause + cross-link.

### Architecture

```mermaid
classDiagram
    direction LR
    class BaseIntegrator {
      <<Base>>
      +collision_check()
      +manifest_sync()
    }
    class HookIntegrator {
      <<Delegate>>
      +sync_integration()
      +reconcile_after_removal()
      +reconcile_dropped_targets(root, names)
      -_clean_apm_entries_from_json(path, stats)$
    }
    class _hook_dropped_targets {
      <<PrivateImpl>>
      +reconcile_dropped_targets(root, names)
      -_reconcile_sidecar_only_orphan(path, stats)
    }
    class _MergeHookConfig {
      <<ValueObject>>
      +config_filename str
      +target_key str
      +event_container_key str
      +schema_strict bool
    }
    class manifest_reconcile {
      <<Module>>
      +reconcile_deployed_state()
      +reconcile_dropped_merge_hook_targets(root, active, declared)
      +declared_target_profiles()
    }
    class LockfileBuilder {
      <<Orchestrator>>
      +build_and_save()
      -_reconcile_dropped_merge_hook_targets()
    }
    class Violation {
      <<ValueObject>>
      +path Path
      +line int
      +qualname str
      +render() str
    }
    class check_hook_config_write_owner {
      <<StaticGuard>>
      +find_violations(root) list~Violation~
    }
    BaseIntegrator <|-- HookIntegrator
    HookIntegrator *-- _MergeHookConfig : reads _MERGE_HOOK_TARGETS
    HookIntegrator ..> _hook_dropped_targets : delegates reconcile_dropped_targets
    _hook_dropped_targets ..> HookIntegrator : calls _clean_apm_entries_from_json
    manifest_reconcile ..> HookIntegrator : creates instance, calls reconcile_dropped_targets
    LockfileBuilder ..> manifest_reconcile : calls reconcile_dropped_merge_hook_targets
    check_hook_config_write_owner *-- Violation : produces
    note for manifest_reconcile "Canonical Owner: computes\nWHICH targets dropped\n(allowed = active | declared)"
    note for HookIntegrator "Canonical Owner: HOW to\nclean merge-hook JSON/sidecar\nvia _clean_apm_entries_from_json"
    note for _hook_dropped_targets "Line-budget split only;\ninvoked from exactly one call site"
    class HookIntegrator:::touched
    class _hook_dropped_targets:::touched
    class manifest_reconcile:::touched
    class LockfileBuilder:::touched
    class Violation:::touched
    class check_hook_config_write_owner:::touched
```

```mermaid
flowchart TD
    CLI["apm install / apm compile / apm update"]
    CLI --> BSave["LockfileBuilder.build_and_save()\nsrc/apm_cli/install/phases/lockfile.py"]
    BSave --> LockOnly{"self.ctx.lockfile_only?"}
    LockOnly -->|Yes: apm lock| Skip["Skip reconciliation"]
    LockOnly -->|No| ReconcileHook["_reconcile_dropped_merge_hook_targets()\nsrc/apm_cli/install/phases/lockfile.py"]
    ReconcileHook --> MR["reconcile_dropped_merge_hook_targets()\nsrc/apm_cli/install/manifest_reconcile.py"]
    MR --> DeclNone{"declared_targets is None?"}
    DeclNone -->|Yes: legacy no-op| NoOp["Return files_removed=0"]
    DeclNone -->|No| ComputeDrop["dropped = KNOWN_TARGETS - active | declared"]
    ComputeDrop --> AnyDrop{"Any dropped names?"}
    AnyDrop -->|No| NoOp
    AnyDrop -->|Yes| HI["HookIntegrator.reconcile_dropped_targets()\nsrc/apm_cli/integration/hook_integrator.py"]
    HI --> Impl["_hook_dropped_targets.reconcile_dropped_targets()\nsrc/apm_cli/integration/_hook_dropped_targets.py"]
    Impl --> Loop["For each dropped target name"]
    Loop --> InRegistry{"name in _MERGE_HOOK_TARGETS?"}
    InRegistry -->|No: e.g. copilot| SkipTarget["Skip silently"]
    InRegistry -->|Yes| NativeExists{"[I/O] json_path.exists()?"}
    NativeExists -->|No| SidecarOnly{"[I/O] sidecar_path.exists()?"}
    SidecarOnly -->|Yes| ValidSidecar{"[I/O] json.load sidecar"}
    ValidSidecar -->|Valid| UnlinkSidecar["[FS] sidecar_path.unlink()"]
    ValidSidecar -->|Malformed| FailClosed1["Warn + leave byte-identical"]
    SidecarOnly -->|No| SkipTarget
    NativeExists -->|Yes| CleanJSON["[FS] _clean_apm_entries_from_json()\nstrip _apm_source entries + unlink sidecar"]
    CleanJSON --> ErrorCheck{"errors increased?"}
    ErrorCheck -->|Yes| FailClosed2["Warn: unreadable/malformed,\nleft unmodified"]
    ErrorCheck -->|No| NextTarget["Continue loop"]
    FailClosed1 --> NextTarget
    FailClosed2 --> NextTarget
    UnlinkSidecar --> NextTarget
    NextTarget --> Loop
    Loop -->|Done| ReturnStats["Return stats dict"]
    ReturnStats --> BSave
    BSave --> EarlyReturn{"No installed_packages\nand not lockfile_only?"}
    EarlyReturn -->|Yes| EarlyExit["Return early"]
    EarlyReturn -->|No| NormalBuild["Normal lockfile build"]

    CLI2["apm compile / apm update\nvia reconcile_deployed_state()"] --> MR2["reconcile_deployed_state()\nsrc/apm_cli/install/manifest_reconcile.py"]
    MR2 --> MR
```

### Recommendation

The code is sound, the test coverage is comprehensive (all 29 tests pass), and the architectural approach is validated by every relevant specialist. Merge is safe. The two spec-text gaps in req-lk-021 and the CHANGELOG tightening are the highest-signal follow-ups -- ideally addressed in-PR before the next release cut since they touch public-commitment surfaces landing for the first time. The `_rich_warning` and docs-deduplication items are lower priority and can land as fast follow-up PRs without blocking this one.

---

<details>

#### Python Architect

- **[nit]** AST guard does not catch native hook-config paths composed as a single string literal at `scripts/check_hook_config_write_owner.py:225`
  `_is_hook_config_path` checks root-dir and filename as separate exact-match fragments (e.g. `.codex` + `hooks.json`), but a bypass like `Path('.codex/hooks.json').write_text('{}')` produces a single fragment `.codex/hooks.json` that matches neither the root-dir set nor the filename set. Sidecar paths are NOT affected. Low real-world risk.
- **[nit]** Design patterns assessment: Delegate/Canonical-Owner/Fail-Closed-Guard patterns correctly applied; no over-engineering.
  Three-layer single-owner-per-decision design (manifest_reconcile decides WHICH, HookIntegrator decides HOW, `_clean_apm_entries_from_json` is the JSON primitive) is sound and proportionate to scope.

#### CLI Logging Expert

- **[recommended]** Fail-closed warnings use _log.warning only, not the dual-channel _rich_warning + _log.warning pattern -- invisible in default console output at `src/apm_cli/integration/_hook_dropped_targets.py:83`
  The three `_log.warning` calls in `_hook_dropped_targets.py` fire during user-facing commands (apm install/compile/update) but only emit to the structured logging channel; default console shows nothing. Existing `reconcile_after_removal` uses the same `_log.warning`-only convention (consistent with prune/uninstall), but forward-install paths elsewhere in the same file DO use the dual-channel pattern. Since this fires on `apm install` (forward command), user-facing channel is warranted.
  *Suggested:* Add a `_rich_warning` parallel for each `_log.warning`, naming the path and consequence, matching the pattern at hook_integrator.py lines 253-263/1631-1640.
- **[nit]** No verbose-mode output on successful cleanup at `src/apm_cli/integration/_hook_dropped_targets.py:88`
  When `files_removed > 0`, no `verbose_detail`/`_log.info` is emitted, unlike the parallel `_sync_cache_pin_markers` path.
- **[nit]** reconcile_dropped_merge_hook_targets return value (stats dict with error count) is discarded at both call sites at `src/apm_cli/install/manifest_reconcile.py:447`
  Aggregate error count is lost at the call boundary even though per-file warnings already surface individual details.

#### DevX UX Expert

- **[nit]** prune.md uses internal 'lifecycle owner' framing instead of user-centric guidance at `docs/src/content/docs/reference/cli/prune.md:93`
  *Suggested:* Reword to user-centric language naming `apm install` as the remedy without "lifecycle owner" jargon.
- **[nit]** Warning for malformed JSON says 'manual review' without naming a concrete next action at `src/apm_cli/integration/_hook_dropped_targets.py:72`
  *Suggested:* Change to "left unmodified; delete or fix the file manually, then re-run apm install".
- **[nit]** install.md parenthetical 'not apm lock' adds noise for most readers at `docs/src/content/docs/reference/cli/install.md:207`

- **[nit]** Lint guard OWNER_MODULE exclusion should explicitly list _hook_dropped_targets.py at `scripts/check_hook_config_write_owner.py:90`
  The private sibling evades the guard only because its path-composition uses attribute references rather than string literals -- accidental resilience, not explicit exclusion. Should be an explicit sanctioned-writer list.
  *Suggested:* Change OWNER_MODULE to a frozenset including both hook_integrator.py and _hook_dropped_targets.py.
- **[nit]** Lint guard does not cover os.remove(), os.unlink(), or Path.rename() as delete surfaces at `scripts/check_hook_config_write_owner.py:160`
  Known-limitation nit per the docstring's acknowledged bounded scope; defense-in-depth could be tightened.

#### OSS Growth Hacker

- **[recommended]** CHANGELOG entry reads as internal archaeology, not a user-facing release note at `CHANGELOG.md`
  References 4 distinct PR/issue numbers mid-narrative, 2 internal class::method names, and a validation-methodology sentence that belongs in the PR body not the changelog. ~125 words vs ~2x sibling entries.
  *Suggested:* Tighten to user-impact-first, one closing PR ref, no class names, no validation methodology.
- **[nit]** install.md bullet could name the user scenario ('switching AI assistants') not just the mechanism name at `docs/src/content/docs/reference/cli/install.md`

#### Auth Expert -- inactive

PR touches merge-hook JSON config reconciliation on target contraction; no authentication behavior, token management, credential resolution, host classification, or remote-host fallback semantics are changed.

#### Doc Writer

- **[recommended]** req-lk-021's precondition may not textually cover the sidecar-only-orphan cleanup the code implements and tests at `docs/src/content/docs/specs/openapm-v0.1.md:924`
  req-lk-021 only engages "when the implementation maintains merge-based hook configuration" (native document exists) and permits removing "any ownership record left empty by that removal." But `_reconcile_sidecar_only_orphan` handles the case where the native document is already ABSENT and only the sidecar remains -- no "removal" event exists to leave the record "empty by." A real, tested, newly-added behavior undercounted by the spec text meant to cite exactly this.
  *Suggested:* Add a sentence explicitly covering "the native document is already absent but an ownership record remains" triggering removal after well-formedness verification.
- **[recommended]** req-lk-021 doesn't codify the fail-closed-on-malformed-JSON guarantee the PR treats as a headline differentiator at `docs/src/content/docs/specs/openapm-v0.1.md:924`
  PR body's Trade-offs section calls fail-closed-on-malformed-JSON a deliberate posture distinct from prune/uninstall's best-effort, with two dedicated test scenarios. Spec elsewhere writes such guarantees directly into requirement text (req-tg-006, req-lk-013) when genuinely new. req-lk-021 is silent on malformed-state handling.
  *Suggested:* Add a sentence mirroring req-tg-006's pattern for malformed-state fail-closed + actionable diagnostic.
- **[recommended]** Same merge-hook path/sidecar parenthetical duplicated verbatim across install.md, update.md, and (near-verbatim) compile.md with no compensating cut at `docs/src/content/docs/reference/cli/install.md:207`
  Adds ~130 net words to the doc corpus with nothing removed elsewhere, violating "state once, reference elsewhere." `producer/author-primitives/hooks-and-commands.md` already has a canonical table/prose section that is the natural single source of truth.
  *Suggested:* Shorten each CLI page's bullet to a short clause + cross-link to hooks-and-commands.md; drop repeated file-name enumeration from update.md/compile.md.
- **[recommended]** hooks-and-commands.md's 'cleaned up automatically' claim is now materially incomplete and isn't cross-linked to the new target-contraction contract at `docs/src/content/docs/producer/author-primitives/hooks-and-commands.md:153`
  That page states the sidecar "is created and cleaned up automatically alongside the native config" with no qualification -- exactly the claim #2253 falsified for dropped targets. Not touched by this diff, no cross-reference to the new CLI notes or req-lk-021.
  *Suggested:* Add a cross-reference sentence pointing to the install.md notes / exact preserve-remove contract.
- **[nit]** New CHANGELOG entry names internal Python symbols and PR-review provenance that no sibling entry uses at `CHANGELOG.md:49`
  *Suggested:* Drop internal function-name citations and the evidence-provenance clause; keep user-facing behavior + closes trailer.

#### Test Coverage Expert

- **[nit]** No explicit apm-update parity test paralleling test_compile_reconciles_dropped_target_hook_state at `tests/integration/test_hook_target_contraction_reconciliation.py:427`
  apm compile (tested) and apm update (untested) both call `reconcile_project_deployed_state()` at the same entry point, so the compile test transitively proves the update path. Marginal regression-trap value near-zero given the shared entry point.
  *Suggested:* Optionally parametrize the compile test to also cover the update argv shape. Low priority.
  *Proof (passed):* `tests/integration/test_hook_target_contraction_reconciliation.py::test_compile_reconciles_dropped_target_hook_state`

#### Performance Expert

- **[nit]** HookIntegrator() instantiated on every reconcile call despite self being unused at `src/apm_cli/install/manifest_reconcile.py:315`
  `reconcile_dropped_merge_hook_targets` instantiates `HookIntegrator()` to call `reconcile_dropped_targets`, which never touches self -- immediately delegates to the module-level `_impl` function. Cost ~1-2 microseconds per install, negligible, but semantically misleading.
  *Suggested:* Make `HookIntegrator.reconcile_dropped_targets` a `@staticmethod`, or call the `_impl` module function directly.
- **[nit]** AST checker CI cost (~3.7s) could be halved with a cheap pre-filter at `scripts/check_hook_config_write_owner.py:289`
  `find_violations()` parses all 415 `*.py` files under `src/apm_cli/` into AST even though >95% contain no relevant string literal. A substring pre-filter before `ast.parse` would cut wall-time to ~0.2-0.3s. CI-only cost, bounded to developer feedback latency.
  *Suggested:* Add a fast substring pre-filter in the per-file loop before calling `ast.parse`.

</details>

danielmeppiel:
## APM Spec Guardian: `fold_and_ship`

> Scope: **editorial-patch**; diff = +41/-2 lines across 1 in-scope file. Shocked-meter avg: **7.75/10**.

All four panels recommend shipping with follow-ups and report zero new B findings. The two highest-signal themes are correctness gaps in the new requirement and are mechanically fixable now: req-lk-021 currently makes compile/update depend on req-lk-020's install-only trigger, and Section 5.7 omits the new consumer requirement. Fold F1-F4, rerun conformance checks, then ship.

The deferred items are legitimate v0.1.1 improvements, but they add normative surface or need a cohesive design pass. The highest-value follow-up is splitting req-lk-021 into lettered subclauses so independent conformance tests can cite its distinct obligations.

### Convergence

| Panel | Stance | Shocked | New B | New R | New N |
|---|---|---:|---:|---:|---:|
| Swagger / OpenAPI editor | ship_with_followups | 8/10 | 0 | 1 | 2 |
| OCI distribution editor | ship_with_followups | 8/10 | 0 | 3 | 2 |
| Package-manager contract editor | ship_with_followups | 7/10 | 0 | 4 | 2 |
| Web-platform architect | ship_with_followups | 8/10 | 0 | 3 | 2 |

> B = new blocking findings, R = new recommended, N = new nits. Counts are advisory signal, not an automated decision.

### Convergent themes

- **T1 - Trigger scope mismatch** (oci-rec-r1-3, pkg-rec-r1-2): req-lk-021 names compile/update but subordinates activation to install-only req-lk-020.
- **T2 - Consumer enumeration gap** (sw-nit-r1-1, pkg-rec-r1-1): Section 5.7 omits req-lk-021 and the pre-existing req-lk-019.
- **T3 - Cross-reference completeness** (oci-rec-r1-1, tag-rec-r1-1): threat-table and hooks-section discovery are incomplete.
- **T4 - Requirement density** (sw-rec-r1-1, tag-nit-r1-1): req-lk-021 is difficult to cite as one monolithic block.

### Fold now

1. **[F1 / T2] Section 5.7** - Add req-lk-019 and req-lk-021 to the Consumer MUST list in ascending order.
   *Success criterion:* `Section 5.7 contains req-lk-017, req-lk-019, req-lk-020, and req-lk-021 before the Consumer SHOULD line.`
2. **[F2 / T1] req-lk-021** - Make the trigger self-standing for non-frozen install, compile, or update operations that rewrite deployed state. Keep req-lk-020 as the semantic source for the preserve-or-remove decision, not as the activation precondition.
   *Success criterion:* `The opening clause no longer says "reconciles installed state under req-lk-020"; the decision-semantics reference remains.`
3. **[F3 / T3] Section 10.11** - Add req-lk-021 to row 7, "Unverified content cleanup", alongside req-lk-020.
   *Success criterion:* `Section 10.11 row 7 cites req-lk-021.`
4. **[F4 / standalone] req-lk-021** - Replace the indeterminate-entry double negative with direct preserve language while retaining fail-safe behavior.
   *Success criterion:* `The clause directly says to preserve indeterminate entries and no longer uses "rather than remove it solely because it was not".`

<details>
<summary>Defer to v0.1.1 (7 items)</summary>

- **[F5 / T4] req-lk-021** - Refactor the distinct obligations into lettered subclauses without changing normative content.
- **[F6 / standalone] req-lk-021** - Define output-side well-formedness verification before replacing the original document.
- **[F7 / standalone] req-lk-021** - Define semantic idempotence for already-reconciled state.
- **[F8 / T3] Section 8 / req-lk-021** - Add a non-normative hooks-section cross-reference.
- **[F9 / standalone] req-lk-021** - Replace the self-declared activation threshold with observable consumer behavior.
- **[F10 / standalone] req-lk-021** - Define absent ownership records as indeterminate and preserve their entries.
- **[F11 / standalone] req-lk-021 / Section 11.3.1** - Add explicit sidecar-only conformance evidence.

</details>

### Declined findings

- **oci-nit-r1-1** - The malformed/orphan clauses already state the fail-safe behavior; the proposed disclaimer edit was not specific enough to verify.
- **pkg-nit-r1-1** - A section separator has no normative or conformance effect.
- **tag-nit-r1-2** - The version-history density preserves traceability and matches prior entries.
- **oci-nit-r1-2** - "Literal text" accurately explains the previous requirement boundary.
- **pkg-nit-r1-2** - F1 makes the existing version-history claim accurate without another edit.

### Linter notes

- **[7] Relative-anchor checker** - The checklist's simple heading-slug algorithm reports 51 unresolved heading links on both `main` and this PR head. The sets are identical, so PR #2275 introduces no new link-resolution failure.
- **[10] CHANGELOG spec-path mention** - The user-facing entry describes the target-contraction fix but does not name an `openapm-*` path. This is advisory and may be folded with F1-F4 if the changelog owner wants explicit spec traceability.
- **[11] Non-spec Python files changed** - Expected for this production fix. The general APM review panel already ran and reported no B findings.

Checks 1-6 and 9 pass. Check 8 is not applicable because the artifact has no Mermaid blocks. Conformance evidence: orphan check aligned 101 requirements; spec suite 141 passed, 2 existing skips; fixture cross-citations pass.

> **Linter handoff:** After F1-F4, recheck Section 5.7 ordering, req-lk-021 trigger wording, Section 10.11 row 7, the 101-statement count across Section 1.3 / Appendix C / Appendix D, and Section 11.3.2 consumer enumeration.

---

<details>
<summary>Full per-panel findings</summary>

#### Swagger / OpenAPI editor - shocked 8/10, confidence high

- **[sw-rec-r1-1] req-lk-021** - Eight distinct MUST obligations need lettered subclauses for independent citability.
- **[sw-nit-r1-1] Section 5.7** - Consumer enumeration omits req-lk-021 and req-lk-019.
- **[sw-nit-r1-2] req-lk-021** - Replace the target-attribution double negative with direct language.

#### OCI distribution editor - shocked 8/10, confidence high

- **[oci-rec-r1-1] Section 10.11** - Threat-table row 7 omits req-lk-021.
- **[oci-rec-r1-2] req-lk-021** - Output-side well-formedness is not explicitly verified.
- **[oci-rec-r1-3] req-lk-021** - Compile/update activation is subordinated to install-only req-lk-020.
- **[oci-nit-r1-1] req-lk-021** - The closing disclaimer under-reports the full binding surface.
- **[oci-nit-r1-2] Appendix D** - Consider "normative scope" instead of "literal text".

#### Package-manager contract editor - shocked 7/10, confidence high

- **[pkg-rec-r1-1] Section 5.7** - Consumer enumeration omits req-lk-021 and req-lk-019.
- **[pkg-rec-r1-2] req-lk-021** - Trigger wording makes compile/update and no-target clauses unreachable.
- **[pkg-rec-r1-3] req-lk-021** - Semantic idempotence is unstated.
- **[pkg-rec-r1-4] req-lk-021** - Absent ownership records are not distinguished from malformed records.
- **[pkg-nit-r1-1] req-lk-021** - A section separator could aid navigation.
- **[pkg-nit-r1-2] Appendix D** - The enumeration claim depends on folding F1.

#### Web-platform architect - shocked 8/10, confidence high

- **[tag-rec-r1-1] req-lk-021** - The inline-only merge-hook definition needs better second-implementer discovery.
- **[tag-rec-r1-2] req-lk-021** - Activation should use observable consumer behavior rather than self-description.
- **[tag-rec-r1-3] req-lk-021** - Sidecar-only orphan behavior needs explicit conformance evidence.
- **[tag-nit-r1-1] req-lk-021** - The monolithic requirement is difficult to cite precisely.
- **[tag-nit-r1-2] Appendix D** - The version-history entry is dense.

</details>

<sub>This panel is advisory. Re-apply the `spec-review` label after addressing feedback to re-run.</sub>

PR #2286, opened 2026-07-17
fix(compile): refuse to clobber hand-authored root context files

fmferrari:
## Problem

`apm compile` writes the primary output target (`CLAUDE.md`, `AGENTS.md`,
`GEMINI.md`, and the per-client variants) by overwriting whatever is already
there, with no check for whether the existing file was hand-authored. A repo
that already ships a hand-written `CLAUDE.md` loses it silently on the first
compile — no warning, no backup, recoverable only from VCS.

The existing `--clean` promise (*"hand-authored files are never deleted"*)
covers **orphan cleanup**, not the **primary output target**, which was always
overwritten.

## Root cause

The compile file-writer treated the output path as always-owned-by-APM. The
tool already distinguishes generated vs hand-authored files elsewhere —
generated files carry a `<!-- Generated by APM CLI … -->` marker, and both
`--clean` orphan cleanup and the `--global` path already gate on it — it just
wasn't applied to the file the writer overwrites.

## Fix

Reuse that same marker as a read-side ownership check at the write chokepoint
(`CompiledOutputWriter`). Before writing a target:

- **Missing, or present *with* the marker** → generate/overwrite as today
  (APM-owned; normal recompiles unaffected).
- **Present *without* the marker (hand-authored)** → refuse with an
  actionable, non-zero-exit error; the file is left untouched.
- **`--force`** → overwrite, writing a `<file>.bak` copy of the original first.
- **`--dry-run`** → report the collision instead of silently planning an
  overwrite.

Protection is opt-in per write, so managed-section mode (which edits a
hand-authored file in place by design) and low-level callers keep the raw
overwrite contract. The guard is applied symmetrically to every single-file
target (`CLAUDE.md`, `AGENTS.md` distributed + single-file, `GEMINI.md`, and
per-client variants), and the marker scan mirrors the first-4096-byte
convention already used by `--clean`.

## Acceptance coverage

- Compiling into a repo with a hand-authored (unmarked) `CLAUDE.md`/`AGENTS.md`
  fails with guidance and leaves the file untouched.
- Compiling into a repo whose file carries the APM marker overwrites as before
  (recompile regression).
- `--force` overwrites and leaves a `<file>.bak` with the original content.
- `--dry-run` reports the collision instead of planning an overwrite.

## Tests

- `tests/unit/compilation/test_output_writer.py` — writer-level guard,
  `--force` + `.bak`, opt-in default, all-or-nothing batch.
- `tests/unit/compilation/test_compile_hand_authored_guard.py` — CLI-level
  coverage for `CLAUDE.md` and `AGENTS.md`, `--force`, `--dry-run`, marked-file
  recompile, and `--global` rejection.

Compilation + compile-command unit suites pass locally (1260 tests); ruff lint
and format are clean.

---


fmferrari:
Thanks for the review — both points addressed in the latest push:

1. **`file_is_hand_authored()` byte vs. character scan** (`output_writer.py`) — now opens the file in binary (`"rb"`), reads exactly `_MARKER_SCAN_BYTES` (4096) **bytes**, then `.decode("utf-8", errors="replace")`, matching `distributed_compiler._file_has_apm_marker` exactly. The scan window is now a true byte bound and can no longer diverge on non-ASCII content.

2. **Missing type annotation on `_dry_run_overwrite_notes(paths, ...)`** (`agents_compiler.py`) — annotated as `paths: Iterable[Path]` (added `from collections.abc import Iterable`), consistent with the surrounding helpers.

Also added a docs commit to `docs/src/content/docs/reference/cli/compile.md`: a `--force` row in the Output-control table and a new "Hand-authored files are never clobbered" section generalizing the marker/ownership contract (previously documented only for `.github/copilot-instructions.md`) to all root-context targets.

sergio-sisternes-epam:

> Ship with follow-ups: hand-authored guard is correct, adoption-unblocking, and test/doc gaps are bounded post-merge work.

cc @fmferrari @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

This PR delivers a clear user-facing safety contract: `apm compile` will no longer silently destroy hand-authored CLAUDE.md, AGENTS.md, or GEMINI.md files. The behavioral correctness is not in dispute across any panelist. The python-architect's concern about `protect_hand_authored=False` as default is architecturally sound but non-blocking: the current call sites pass True explicitly, so the guard is active where it matters today. Flipping the default is a pit-of-success improvement best landed as a fast follow-up rather than gating an adoption-unblocking fix.

The supply-chain marker-spoofing finding is real but pre-existing and bounded: it requires a malicious package to control the literal first bytes of a compiled output template, which is already a trust boundary violation detectable at review time. It does not regress the security posture and should be tracked as a hardening issue, not a ship-blocker. The GEMINI.md test gap is the most defensible blocking candidate -- a `secure-by-default` code path with zero exercise -- but the code is structurally identical to the CLAUDE.md path that IS tested, and the risk of silent regression is low. I side with shipping now and requiring the GEMINI.md test in a follow-up PR within the same milestone.

Doc gaps (exit codes, --global+--force) are important for completeness but do not affect runtime safety. They should land before the next minor release note but do not block this patch.

**Dissent.** python-architect rated `protect_hand_authored=False` default as recommended; test-coverage-expert rated GEMINI.md missing test as recommended with `secure-by-default` principle tag. Under the panel's evidence-weighting rules, missing tests on `secure-by-default` surfaces rank near blocking. I side with shipping because the untested path is a mechanical duplicate of the tested CLAUDE.md path (same function, same branch), making silent divergence unlikely -- but the follow-up test is non-negotiable and must land before any release that includes this commit.

**Aligned with:** Secure by default (guard refuses overwrite by default; --force is explicit opt-in; marker-spoofing tracked separately), Pragmatic as npm (mirrors npm's refusal to overwrite modified files during init/publish with clear --force escape hatch and .bak backup), OSS community driven (directly unblocks adoption for teams with existing AI context files).

**Growth signal.** This is a top-of-funnel unblock for the CLAUDE.md power-user segment. Release note angle: 'APM is now safe to try in repos that already have hand-authored AI context files.' The growth-hacker's suggestion of a targeted social beat is well-founded -- this addresses a concrete fear that blocks trial adoption. Frame in CHANGELOG as a feature (safety), not just a bugfix.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 2 | 1 | Sound guard design; protect_hand_authored should default True (opt-out) and the catch in cli.py duplicates the catch in agents_compiler. |
| CLI Logging Expert | 0 | 2 | 2 | Messages are actionable and well-structured; minor symbol consistency and verbose-mode gaps worth addressing. |
| DevX UX Expert | 0 | 0 | 2 | Solid protective UX; --force semantics and error guidance are well-aligned with npm/cargo conventions. |
| Supply Chain Security | 0 | 2 | 2 | Marker-spoofing risk is real but bounded; backup atomicity and TOCTOU are minor concerns worth noting. |
| OSS Growth Hacker | 0 | 2 | 1 | Strong adoption-unblock: protects existing CLAUDE.md users from silent data loss, clears a real barrier to trying APM in repos with prior AI context files. |
| Auth Expert | -- | -- | -- | No auth surface touched; inactive. |
| Doc Writer | 0 | 3 | 2 | New section is accurate and well-placed; three gaps: --global+--force undocumented, exit-codes table missing the hand-authored-file case, migration path buries simplest option. |
| Test Coverage | 0 | 1 | 1 | CLAUDE.md and AGENTS.md guard paths well-tested; GEMINI.md guard and OSError-returns-True branch untested. |
| Performance Expert | 0 | 0 | 3 | Negligible overhead: 3 x 4KB reads on a cold path; no hot-path regression. |

### Top 5 follow-ups

1. **[Test Coverage]** Add GEMINI.md hand-authored guard end-to-end test -- Missing test on a secure-by-default code path; structurally identical to tested CLAUDE.md path but must be exercised to prevent silent regression.
2. **[Python Architect]** Flip protect_hand_authored default to True (opt-out) -- Pit-of-success: new call sites should be safe by default without remembering to pass True.
3. **[Supply Chain Security]** Track marker-spoofing hardening as a follow-up issue -- Malicious package could inject marker prefix to bypass guard; bounded but worth hardening.
4. **[Doc Writer]** Document exit codes and --global+--force rejection in compile reference -- Completeness gap; users hitting the refusal need to find the exit code and workaround in docs.
5. **[Test Coverage]** Add unit test for OSError->True branch in file_is_hand_authored -- Secure-by-default fallback (unreadable file treated as hand-authored) has no direct exercise.

### Architecture

```mermaid
classDiagram
    direction LR
    class CompiledOutputWriter {
        <<Service>>
        +write_many(outputs, force, protect_hand_authored) ScanVerdict
        +prepare(outputs) tuple
    }
    class HandAuthoredOverwriteError {
        <<Exception>>
        +paths list~Path~
    }
    class AgentsCompiler {
        <<Facade>>
        +compile(config) CompilationResult
    }
    class CompilationConfig {
        <<ValueObject>>
        +force bool
        +dry_run bool
    }
    class CompilationResult {
        <<ValueObject>>
        +diagnostics list
    }
    AgentsCompiler *-- CompiledOutputWriter : delegates writes
    AgentsCompiler ..> CompilationConfig : reads
    AgentsCompiler ..> CompilationResult : returns
    CompiledOutputWriter ..> HandAuthoredOverwriteError : raises
    note for CompiledOutputWriter "Guard pattern: scan-before-write"
    class CompiledOutputWriter:::touched
    class HandAuthoredOverwriteError:::touched
    class CompilationConfig:::touched
```

```mermaid
flowchart TD
    A["cli.py: compile command"] --> B["AgentsCompiler.compile(config)"]
    B --> C{"config.dry_run?"}
    C -->|Yes| D["_dry_run_overwrite_notes()\n[FS] read heads"]
    C -->|No| E["CompiledOutputWriter.write_many()"]
    E --> F{"protect_hand_authored?"}
    F -->|No| G["[FS] atomic_write_text all targets"]
    F -->|Yes| H["[FS] file_is_hand_authored() scan heads"]
    H --> I{"hand_authored found?"}
    I -->|None| G
    I -->|"Yes, force=False"| J["raise HandAuthoredOverwriteError"]
    I -->|"Yes, force=True"| K["[FS] _backup_hand_authored .bak"]
    K --> G
    J --> L["AgentsCompiler catches -> CompilationResult.diagnostics"]
    D --> M["Report collisions in dry-run output"]
```

### Recommendation

Ship now. The guard is behaviorally correct, adoption-unblocking, and well-aligned with APM's secure-by-default principle. All recommended follow-ups (GEMINI.md test, default flip, docs, marker hardening) are bounded and should land within the same milestone but do not gate this merge.

---

<details>

#### Python Architect

- **[recommended]** protect_hand_authored=False default inverts the safety contract at `src/apm_cli/compilation/output_writer.py`
  write_many defaults protect_hand_authored to False, making the guard opt-in. Every new call site must remember to pass True. The safer contract is opt-out: default True, let callers pass False when they knowingly want raw overwrite. This follows the pit-of-success principle.
  *Suggested:* Change signature to protect_hand_authored=True and update only the internal/test call sites that need raw writes to pass False explicitly.
- **[recommended]** HandAuthoredOverwriteError caught in both cli.py and agents_compiler -- split authority on error handling at `src/apm_cli/commands/compile/cli.py`
  agents_compiler.compile() catches HandAuthoredOverwriteError to wrap it into a CompilationResult, and cli.py also catches it around the AGENTS.md write. Two catch sites means divergent UX. The canonical owner of compile-error rendering should be one layer only.
- **[nit]** APM_GENERATED_MARKER_PREFIX duplicates knowledge of marker format at `src/apm_cli/compilation/constants.py`
  The marker string is already produced by a template or constant elsewhere. If the write-side template changes, the read-side prefix silently stops matching.

#### CLI Logging Expert

- **[recommended]** logger.error() in cli.py should use symbol='error' to guarantee [x] prefix at `src/apm_cli/commands/compile/cli.py`
  Verify the error path renders as '[x] CLAUDE.md exists and was not generated by APM -- refusing to overwrite.' consistently.
- **[recommended]** dry-run progress uses symbol='info' but should use symbol='warning' for refusal variant at `src/apm_cli/compilation/agents_compiler.py`
  [!] for 'would refuse to overwrite'; [i] for 'would overwrite (backup)'. Communicates severity correctly at a glance.
- **[nit]** hand_authored_overwrite_message is a wall of text -- break into lead outcome + fix lines
- **[nit]** No --verbose detail about WHY the file is considered hand-authored

#### DevX UX Expert

- **[nit]** Error message is long; split into problem line + guidance line
  npm and cargo separate the 'what went wrong' line from 'hint:' lines.
- **[nit]** Dry-run + force combination message could name the backup path explicitly

- **[recommended]** Malicious package can inject marker prefix to bypass hand-authored protection
  A dependency's compiled output could prepend the APM marker prefix to a root context file template, causing file_is_hand_authored() to return False and allowing apm compile to silently overwrite.
  *Suggested:* Consider checking that the marker appears on the very first line rather than anywhere in 4096 bytes, or include a hash/version in the marker only the CLI emits.
- **[recommended]** Backup file written non-atomically with path.write_bytes()
  _backup_hand_authored uses path.write_bytes() which is not atomic -- a crash mid-write leaves a corrupt .bak. User's only safety net could be lost.
  *Suggested:* Use the same atomic_write pattern (write to temp, fsync, rename). Preserve source file permissions.
- **[nit]** TOCTOU window between hand-authored check and write (local-only, low exploitability)
- **[nit]** path.with_name() for backup does not flow through path_security guards

#### OSS Growth Hacker

- **[recommended]** Frame as adoption-unblock in release notes, not just a bug fix
  Story-shaped change: 'APM now respects your existing AI context files.' One-line repostable claim: 'apm compile no longer silently overwrites your hand-crafted context files.'
- **[recommended]** Migration path should be discoverable from the error message itself
  Verify error includes: (1) why it refused, (2) --force with .bak mention, (3) short pointer to migration guidance.
- **[nit]** Quickstart or first-package guide could mention existing-repo safety

#### Auth Expert -- inactive

No auth fast-path files touched; PR changes compile-time file ownership logic with no credential resolution, token management, or host classification surface.

#### Doc Writer

- **[recommended]** --global + --force interaction not documented at `docs/src/content/docs/reference/cli/compile.md`
  The --global section does not mention whether --force is accepted or rejected. Users will hit this wall; the --global flag description should explicitly address --force.
- **[recommended]** Exit-codes table does not cover the hand-authored-file refusal case at `docs/src/content/docs/reference/cli/compile.md`
  Extend exit code 1 row with 'hand-authored root context file detected without --force'.
- **[recommended]** Migration path buries the simplest recoverable action at `docs/src/content/docs/reference/cli/compile.md`
  Lead with marker-prepend (fastest), then primitive migration, then --force.
- **[nit]** Section placement H3 under H2 'Output layout per target' slightly off for discoverability
- **[nit]** Dry-run collision message in prose uses ellipsis, inconsistent with --clean examples style

#### Test Coverage Expert

- **[recommended]** GEMINI.md hand-authored guard path has no test at `tests/unit/compilation/test_compile_hand_authored_guard.py`
  No test exercises target=gemini with a pre-existing hand-authored GEMINI.md. The GEMINI write site in agents_compiler.py is covered by the new guard logic but never exercised end-to-end.
  *Proof (missing):* `tests/unit/compilation/test_compile_hand_authored_guard.py::test_refuses_to_overwrite_hand_authored_gemini_md` -- proves: GEMINI.md hand-authored file is protected by compile guard [secure-by-default,devx]
- **[nit]** OSError -> True branch in file_is_hand_authored has no direct unit test at `tests/unit/compilation/test_output_writer.py`
  *Proof (missing):* `tests/unit/compilation/test_output_writer.py::test_file_is_hand_authored_returns_true_on_oserror` -- proves: Unreadable files are treated as hand-authored (safe default) [secure-by-default]

#### Performance Expert

- **[nit]** Double read in --force path: file_is_hand_authored() reads 4KB head, then _backup_hand_authored() re-reads entire file
- **[nit]** _dry_run_overwrite_notes() re-scans files already checked by write_many() in non-dry-run flow
- **[nit]** 4096-byte head read is fine but 512 bytes would be equally correct (marker appears in first ~80 bytes)

</details>

tillig:
#2779 (merged) may overlap/supersede this PR.

danielmeppiel:
Thanks @fmferrari for tackling the risk of losing hand-authored context files, and @tillig for pointing out the overlap.

I'm closing this as superseded by #2779 and #2790. Current main protects hand-authored root AGENTS.md, CLAUDE.md, and GEMINI.md through the shared, catalog-derived protection owner. The safety goal supported by the earlier reviews has now landed through those successors; there is no need to rebase another implementation of it.

We are explicitly not adopting this PR's compile-level --force, automatic .bak backup, or whole-batch-abort behavior. The chosen contract preserves the user's file, warns, and continues generating unrelated safe outputs. An ownership-transfer operation would be a separate product decision, not an outstanding requirement of this fix.

This closure reflects the implementation on main, rather than a claim that it is already in a published release. Thank you for the contribution and the review iterations.

PR #2301, opened 2026-07-17
fix(install): carry forward mcp_target_servers to stop generated_at churn

### LINKED ISSUE #2297: [BUG] Non-idempotent apm install: generated_at churns on every run (regression of #1532) on 0.25.0
**Describe the bug**

On APM `0.25.0`, `apm install` rewrites apm.lock.yaml on **every** run even when nothing has changed — neither apm.yml, nor any resolved dependency, nor any MCP server. The only diff is the `generated_at` timestamp. This is a regression of the idempotency gate introduced in #450 (fixed v0.8.6) and extended to MCP deps in #1532 (fixed v0.16.1).

The manifest declares local + virtual (subpath) APM dependencies and one self-defined HTTP MCP server. During the failing installs the CLI logs confirm nothing was re-resolved: all deps are `(cached)` and the MCP server reports `already configured` / `All servers up to date`, yet `generated_at` is still rewritten.

**To Reproduce**

1. In a project whose apm.yml declares at least one MCP server and local/virtual APM deps, run `apm install` and commit apm.lock.yaml.
2. Snapshot the file: `cp apm.lock.yaml /tmp/snap.yaml`
3. Run `apm install` again (change nothing).
4. `diff /tmp/snap.yaml apm.lock.yaml`

Observed diff (reproduced twice back-to-back on the same machine):
```diff
2c2
< generated_at: '2026-07-17T21:20:46.288063+00:00'
---
> generated_at: '2026-07-17T21:21:02.654594+00:00'
```
Relevant install output during the same runs:
```
+- MCP Servers (1)
|  [+] mcp-server (already configured)
[*] All servers up to date
[*] Installed 13 APM dependencies in 1.9s.
```

**Expected behavior**

`apm install` should be idempotent: when neither apm.yml nor any resolved dependency/MCP state has changed, apm.lock.yaml must not be modified (no `generated_at` rewrite) — matching the behavior established in #450 and #1532. Committed lockfiles should stay diff-free across repeated installs so teams don't get spurious `git status` noise and merge conflicts.

- OS: macOS
- Python Version: 3.9.21 (system); APM installed as the standalone CLI binary
- APM Version: 0.25.0 (d73e6ac)
- VSCode Version (if relevant): n/a (reproduced from the CLI)

**Logs**

`apm install` exits 0 both times. No error is emitted; the only side effect is the `generated_at` rewrite shown above. (A benign, unrelated warning is also present: `Could not fetch org policy ... HTTP 410 ... proceeding without policy enforcement`.)

**Additional context**

The apm.yml uses `targets: [claude, copilot]`, `includes: auto`, local-path devDependencies, virtual (subpath) deps resolved from a self-hosted GitLab, and one self-defined HTTP MCP server (`registry: false`, `transport: http`).

Possibly related / adjacent:
- #1532 — the MCP determinism gate this appears to regress.
- #2155 / #2158 (v0.25.0) — large refactor of the deployment ledger and `mcp_target_servers` ownership, a plausible regression point.

Separate but observed in the same repo (may warrant its own issue): the MCP deployment block in the committed lockfile (`deployments[].runtime`, `mcp_target_servers`) is **machine-dependent** — it records the MCP runtime detected locally (e.g. `copilot`/`vscode` vs `claude`). Developers with different installed harnesses produce alternating diffs to the committed lockfile, causing recurring merge conflicts, in the spirit of the "keep regenerated lockfiles machine-independent" fix in #2155.

rrazvd:
## Description

`apm install` rewrote `apm.lock.yaml` (diff only in `generated_at`) on every run, even when nothing changed, whenever the project had an MCP server deployed to a target.

Root cause: `LockfileBuilder._preserve_existing_mcp_state` (`src/apm_cli/install/phases/lockfile.py`) carried `mcp_servers`, `mcp_configs`, and `mcp_config_provenance` forward from the on-disk lockfile, but not `mcp_target_servers`. The freshly-built lockfile therefore always started from `mcp_target_servers={}`, which looked like a real change against the on-disk file and forced a write in `_write_if_changed`. `MCPIntegrator.update_lockfile` then ran right after, re-detected the correct `mcp_target_servers`, and had to correct it with a *second* write. Both writes stamp a fresh `generated_at`, so every install produced a diff-only-in-timestamp change even though the final content was correct.

Fix: carry `mcp_target_servers` forward too, through the same `DeploymentLedgerCodec.replace_mcp_target_servers` path `MCPIntegrator` already uses, so the first write starts from consistent state and both writes become true no-ops when nothing changed.

Confirmed via a hermetic end-to-end repro (local-path deps + several GitLab-shorthand virtual/subpath deps + one self-defined `registry: false` HTTP MCP server, matching the shape reported in #2297): before the fix, a second back-to-back `apm install` rewrote the lockfile twice and changed `generated_at`; after the fix, zero writes occur on the no-op reinstall.

Fixes #2297

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

Details:
- Added `test_unchanged_mcp_target_servers_do_not_rewrite_lockfile` to `tests/unit/install/test_mcp_lockfile_determinism.py`, which fails on `main` and passes with this fix.
- `tests/unit/install/`, `tests/unit/deps/`, `tests/test_lockfile.py`, `tests/unit/core/` — 4404 passed, 4 pre-existing failures unrelated to this change (reproduced identically on `main` without this patch; multi-harness-detection/registry-import tests, not lockfile-related).
- `tests/integration/test_global_mcp_lockfile_e2e.py`, `tests/unit/integration/test_mcp_config_view.py`, `tests/unit/deps/test_lockfile_deployment_paths.py`, `tests/unit/deps/test_lockfile_consumer_contract.py`, `tests/unit/deps/test_lockfile_git_semver.py` — 40 passed.
- `ruff check` on changed files — clean.

This PR does not add or change a normative requirement -- it fixes a **Mode A silent regression** against an existing one. [req-lk-005](https://github.com/microsoft/apm/blob/main/docs/src/content/docs/specs/openapm-v0.1.md#L1042) already states: *"A no-op install operation MUST NOT rewrite a lockfile whose only changed fields would be [`generated_at`, `apm_version`]."* Per CONTRIBUTING.md's Mode A guidance ("a code change in `src/apm_cli/**` breaks an assertion bound to a `req-XXX`... fix the code, do not touch the spec"), no spec/manifest edit is needed here.

Note: the existing `@pytest.mark.req("req-lk-005")` test in `tests/spec_conformance/test_lockfile_reqs.py` only checks schema shape and the canonical-ordering clause of req-lk-005 -- it does not exercise the no-op-rewrite behavioral clause end-to-end, which is how this regression shipped in 0.25.0 without being caught. Happy to follow up with a behavioral `req-lk-005` spec-conformance test (mirroring the new unit test) in a separate PR if maintainers want that gap closed -- flagging it here rather than bundling it into this bug-fix PR.

- [ ] Spec edit: `docs/src/content/docs/specs/openapm-v0.1.md` updated (new/changed `<a id="req-XXX"></a>` anchor + prose + Appendix C row).
- [ ] Manifest edit: `docs/src/content/docs/specs/manifests/openapm-v0.1.requirements.yml` updated.
- [ ] Test edit: a `@pytest.mark.req("req-XXX")` test under `tests/spec_conformance/` added or extended.
- [ ] `CONFORMANCE.{md,json}` regenerated via `uv run --extra dev python -m tests.spec_conformance.gen_statement` and committed.
- [x] N/A -- this PR does not change OpenAPM-observable behaviour (it restores conformance with existing req-lk-005; no new requirement, no spec text change).

### DIFF
--- a/src/apm_cli/install/phases/lockfile.py
+++ b/src/apm_cli/install/phases/lockfile.py
@@ -427,6 +427,20 @@ class LockfileBuilder:
             lockfile.mcp_config_provenance = copy.deepcopy(
                 self.ctx.existing_lockfile.mcp_config_provenance
             )
+            # Also carry forward mcp_target_servers (and its ledger rows) via
+            # the same codec path MCPIntegrator itself uses. Without this, the
+            # freshly-built lockfile always starts with mcp_target_servers={},
+            # which reads as a real change against the on-disk file and forces
+            # an extra write here -- MCPIntegrator then has to correct it
+            # with a second write of its own, churning generated_at twice per
+            # install even when nothing changed (issue #2297).
+            if self.ctx.existing_lockfile.mcp_target_servers:
+                from apm_cli.core.deployment_ledger import DeploymentLedgerCodec
+
+                DeploymentLedgerCodec.replace_mcp_target_servers(
+                    lockfile,
+                    copy.deepcopy(self.ctx.existing_lockfile.mcp_target_servers),
+                )
             if self.ctx.logger:
                 self.ctx.logger.verbose_detail(
                     "MCP state unchanged -- carrying forward "
--- a/tests/unit/install/test_mcp_lockfile_determinism.py
+++ b/tests/unit/install/test_mcp_lockfile_determinism.py
@@ -109,6 +109,8 @@ def _run_lockfile_phase_and_mcp_persist(
     project_root: Path,
     package: APMPackage,
     instant: datetime,
+    *,
+    mcp_target_servers: dict[str, list[str]] | None = None,
 ) -> None:
     lock_path = get_lockfile_path(project_root)
     dep_ref = package.get_apm_dependencies()[0]
@@ -138,6 +140,7 @@ def _run_lockfile_phase_and_mcp_persist(
             MCPIntegrator.get_server_names(mcp_deps),
             lock_path,
             mcp_configs=MCPIntegrator.get_server_configs(mcp_deps),
+            mcp_target_servers=mcp_target_servers,
         )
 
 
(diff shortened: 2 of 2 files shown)


danielmeppiel:
Thank you for the original work on this fix. To land it promptly we have opened a superseding PR (#2306) under microsoft/apm that preserves your authorship via commit authorship/trailers and resolves the follow-ups surfaced by the shepherd-driver pass.

Closing this PR in favor of #2306. Your contribution is credited on every cherry-picked commit; the superseding PR body links back here. Please raise concerns on the superseding PR if the changes diverge from your intent -- we want your sign-off too.

danielmeppiel:
Superseded by #2306.

PR #2329, opened 2026-07-21
[BUG] fix(audit): self-hydrate cold-cache ci audit from lock pins

### LINKED ISSUE #2328: [BUG] config-consistency fails for virtual claude_skill deps in setup-only CI (no apm install)
**Describe the bug**

`apm audit --ci` fails the `config-consistency` check for virtual `claude_skill` dependencies (`is_virtual: true`, `package_type: claude_skill`, no `apm.yml`) when it runs **without a prior `apm install`** — i.e. `apm_modules/` is not materialized on disk.

This is a regression in the topology that #2214 was meant to fix. #2214 correctly stopped flagging valid manifestless virtual Claude skills, but the merged v0.26.0 implementation of `_allows_missing_manifest` tightened the exemption to **probe the installed on-disk shape** (`detect_package_type(package_dir)` requiring a root `SKILL.md`). In a setup-only CI job (install the CLI, then `apm audit --ci`, no `apm install`) `apm_modules/` is never populated, so `detect_package_type` sees an absent directory, returns `INVALID`, the exemption is denied, and `config-consistency` hard-fails.

Note the asymmetry with the `drift` check: `drift` replays into a throwaway scratch dir and **tolerates a cold cache** (it skips gracefully with `drift skipped: install cache not populated`). `config-consistency` instead reads the live `apm_modules/` and hard-fails when it is absent — so the same cold-cache CI run gets `drift` skipped but `config-consistency` failed.

Root cause in `src/apm_cli/integration/mcp_config_view.py`:

```python
def _allows_missing_manifest(dependency, package_dir) -> bool:
    if dependency.package_type == PackageType.SKILL_BUNDLE.value:
        return True
    dependency_ref = dependency.to_dependency_ref()
    if not dependency_ref.is_virtual_subdirectory():
        return False
    package_type, _ = detect_package_type(package_dir)   # <- probes disk; empty in setup-only CI
    return package_type is PackageType.CLAUDE_SKILL
```

The E2E tests added in #2214 (`tests/integration/test_virtual_skill_audit_e2e.py`) always run a full `apm install` (then `apm install --frozen`, which re-materializes `apm_modules/` with `SKILL.md`) *before* auditing, so they never exercise the pure setup-only + audit topology.

**To Reproduce**
1. A project whose `apm.yml`/`apm.lock.yaml` has one or more virtual Claude-skill deps (`git: <owner>/<repo>`, `path: <subdir>`), locked as `is_virtual: true`, `package_type: claude_skill`.
2. On a fresh checkout (no `apm_modules/` on disk; `apm_modules/` is typically gitignored), install only the CLI and run the audit — no `apm install`:
   ```yaml
   - uses: microsoft/apm-action@v1
     with:
       setup-only: true
       apm-version: 0.26.0
   - run: apm audit --ci
   ```
3. See `config-consistency` fail:
   ```
   config-consistency  5 MCP config inconsistenc(ies) -- run 'apm install' to reconcile
     - <owner>/<repo>/<path>: package manifest not found at
       apm_modules/<owner>/<repo>/<path>/apm.yml; re-run 'apm install' to restore it
   [x] 1 of 8 check(s) failed
   ```
   (`drift` is reported as skipped in the same run: `drift skipped: install cache not populated`.)

**Expected behavior**
`config-consistency` should treat a virtual `claude_skill` dep whose package directory is not materialized on disk the same way `drift` treats a cold cache: trust the frozen lockfile classification rather than hard-fail. Auditing a clean, valid lockfile in a setup-only CI must pass. When `apm_modules/` *is* materialized, the strict on-disk shape probe must still run (so a mislabeled or malformed installed package is still rejected).

 - OS: Linux (GitHub Actions `ubuntu-latest`); also reproduced on Windows
 - Python Version: 3.14 (also observed with the released binary)
 - APM Version: 0.26.0 (commit 64a7fb5)
 - VSCode Version (if relevant): n/a

**Logs**
```
                           [>] APM Policy Compliance
 config-consistency  5 MCP config inconsistenc(ies) -- run 'apm install' to reconcile
   config-consistency details:
     - angular/skills/angular-developer: package manifest not found at
       apm_modules/angular/skills/angular-developer/apm.yml; re-run 'apm install' to restore it
     - luccasa/ai-tools/globals/skills/front/ngrx-signals: package manifest not found ...
     - luccasa/ai-tools/globals/skills/front/ui/accessibility: package manifest not found ...
     - luccasa/lucca-front/.github/skills/lucca-front/lucca-front-21-2-5: package manifest not found ...
     - microsoft/playwright-cli/skills/playwright-cli: package manifest not found ...
 drift  drift skipped: install cache not populated (run 'apm install' first or pass --no-drift)
[x] 1 of 8 check(s) failed
```

**Additional context**
Proposed fix: in `_allows_missing_manifest`, when the package directory does not exist (cold cache / setup-only audit), fall back to the frozen lockfile classification:

```python
    if not package_dir.exists():
        return dependency.package_type == PackageType.CLAUDE_SKILL.value
    package_type, _ = detect_package_type(package_dir)
    return package_type is PackageType.CLAUDE_SKILL
```

This mirrors the `drift` check's cold-cache tolerance, preserves the strict on-disk probe when modules are materialized, and keeps the four E2E scenarios from #2214 green. PR to follow.

Related: #2207, #2213 (both closed by #2214).

### LINKED ISSUE #2392: [BUG] On repos that commit deployed files, `apm audit --ci` cannot fail in CI: the required `apm install` re-syncs the drift before the check runs
**Describe the bug**

Applies to repos that commit their deployed files rather than gitignoring them.

- Drift detection (#1071) needs a populated `apm_modules/`, which is gitignored. In CI that means `apm audit` has to be preceded by `apm install` — without it, drift is skipped with `drift skipped: install cache not populated` and, on apm >= 0.26.0, `config-consistency` fails instead (#2328).
- But `apm install` also re-deploys into the working tree, overwriting the stale deployed files with fresh ones.
- `apm audit` then compares the working tree against what APM produces, which right after an install always matches.

So on these repos the audit cannot fail in CI for the case it exists to catch: committed deployed files lagging their `.apm/` source. The job is green, `git status` is dirty, and nothing reports it.

Audit answers "does the working tree match what APM produces". What a CI job on a commit-the-deployment repo needs answered is "does the *commit* match".

**To Reproduce**

On a repo that commits its deployed files, with `targets: [claude, copilot]`:

1. `apm install` — clean tree, `apm_modules/` populated.
2. Edit a file under `.apm/instructions/`; do not re-install.
3. `apm audit --ci` → exit 1, drift reported. Correct.
4. `apm install`
5. `apm audit --ci` → exit 0.
6. `git status --porcelain` → the deployed files are modified. Push this and the committed deployment is stale, with no check saying so.

Step 1 matters: from a genuinely cold cache, step 3 exits 1 for a different reason — drift is skipped and `config-consistency` fails.

**Expected behavior**

Some way for `apm audit --ci` to catch this in CI. Two possible shapes:

- Let the drift replay populate its own cache without deploying into the working tree. `apm install --root` already redirects writes under a given directory, so the machinery may be close.
- Or an explicit compare-against-the-commit mode, which could reuse the VCS adapter interface proposed in #1342 (`is_tracked`, `is_ignored`).

**Environment**

- OS: macOS 26.6 (also observed on ubuntu-latest in CI)
- Python Version: 3.14.6
- APM Version: 0.26.0 (64a7fb5)

**Logs**

File names genericised; output otherwise as emitted.

Step 3, with the cache populated:

```
[!] Drift detected: 2 file(s)
  modified (2):
    - .claude/rules/foo.md
    - .github/instructions/foo.instructions.md
  [i] Run 'apm install' to re-sync deployed files with the lockfile.
[x] 2 of 9 check(s) failed
```

The two failing checks are `content-integrity` (`2 file(s) with hash drift`) and `drift`.

Step 5, same repo, the same edit still in place, nothing committed:

```
[+]      drift    no drift detected against lockfile
[*] All 10 check(s) passed
```

Step 6:

```
 M .apm/instructions/foo.instructions.md
 M .claude/rules/foo.md
 M .github/instructions/foo.instructions.md
```

For contrast, the same edit from a cold cache — exit 1, but not for the drift:

```
[+]      drift               drift skipped: install cache not populated
         config-consistency  package manifest not found at ...
[x] 1 of 8 check(s) failed
```

**Additional context**

We run apm across ~40 repos, a mix of committed and gitignored deployments. Our workaround: after `apm install`, fail the job when `git status --porcelain -- .` is non-empty, scoped per install context so a multi-context repo reports each one separately, and a no-op on gitignored deployments since ignored files never appear.

It works, but every organisation with committed deployments has to independently discover the gap and maintain the same shell guard, which is why we think the comparison belongs in `apm audit`.

hugoguitton-lucca:
## Summary

Fixes #2328.
Fixes #2392.

`apm audit --ci` previously had two cold-cache blind spots:
- `config-consistency` failed on valid virtual `claude_skill` deps when `apm_modules/` was absent.
- `drift` skipped on cold caches, so repos that commit deployed files could not fail CI for stale committed outputs.

## What changed

- add one lock-pinned scratch replay owner for `apm audit --ci`
  - runs a frozen manifest/lock preflight
  - materializes missing packages into a scratch-private `apm_modules/`
  - never writes the checkout, `apm.lock.yaml`, or live `apm_modules/`
- route both `config-consistency` and `drift` through that shared replay when `apm_modules/` is absent
- fail closed on replay errors instead of reporting a green drift skip in CI
- only treat missing replay outputs as drift when the path is git-tracked, so committed deployment repos fail correctly while clean gitignored deployments stay green
- tighten the manifestless virtual-skill waiver so the materialized shape and lock metadata must both identify `claude_skill`
- add the shared-owner architecture row, boundary lint, and regression coverage
- add real-CLI lifecycle proofs for tracked committed deployments, gitignored controls, and cold-cache virtual-skill allow and deny rows

### Lifecycle suites
- `APM_E2E_TESTS=1 uv run --extra dev pytest -q tests/integration/test_required_lifecycle_state_machine.py tests/integration/test_audit_cold_cache_ci_e2e.py tests/integration/test_virtual_skill_audit_e2e.py`
- `APM_E2E_TESTS=1 uv run --extra dev pytest -q tests/integration/test_audit_cold_cache_ci_e2e.py::test_cold_cache_audit_fails_for_tracked_stale_deployments_and_reconciles_idempotently tests/integration/test_audit_cold_cache_ci_e2e.py::test_cold_cache_audit_stays_green_for_gitignored_clean_deployments_when_comparison_runs tests/integration/test_virtual_skill_audit_e2e.py::test_valid_manifestless_virtual_skill_audits_clean_from_cold_cache tests/integration/test_virtual_skill_audit_e2e.py::test_cold_cache_malformed_virtual_skill_bytes_fail_closed`

### Targeted regression suite
- `uv run --extra dev pytest -q tests/unit/install/test_architecture_invariants.py tests/unit/install/test_drift_detection.py tests/unit/install/test_drift_phase3.py tests/unit/integration/test_mcp_config_view.py tests/unit/policy/test_ci_checks.py tests/unit/test_audit_ci_command.py tests/integration/test_virtual_skill_audit_e2e.py tests/integration/test_audit_cold_cache_ci_e2e.py tests/integration/test_architecture_authorities.py -k 'not throttle'`

### Quality and taxonomy
- `uv run --extra dev pytest -p no:cacheprovider -q tests/quality`
- `uv run --frozen python scripts/check_test_assertions.py`
- `uv run --frozen python scripts/check_exact_test_duplicates.py`

### Lint mirror
- `uv run --extra dev ruff check src/ tests/`
- `uv run --extra dev ruff format --check src/ tests/`
- `bash scripts/lint-architecture-boundaries.sh`
- `uv run --extra dev python -m pylint --disable=all --enable=R0801 --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/`
- `bash scripts/lint-auth-signals.sh`

## Notes

README.md was not changed.

apm-spec-waiver: CI audit cold-cache replay changes CLI behavior only; no OpenAPM spec artifact or wire-format contract changes.


danielmeppiel:

> Cold-cache CI audit self-hydrates from lock pins without mutating the checkout, unlocking zero-install CI gates for committed AI tool configs.

cc @hugoguitton-lucca @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

All nine review lenses converged on the same core signal: the single-owner preflight + lock-pinned scratch materialization design is sound, the no-write invariant holds, and the npm-ci-like contract is correctly mapped to `apm audit --ci`. The key shepherd reservations are satisfied: one shared replay owner now feeds both `config-consistency` and `drift`, there is no `apm ci` command, lock pins are used directly, cold-cache replay fails closed, and the regression/architecture guardrail set exists.

The remaining signal is follow-up, not stop-ship: docs now need one more same-cycle cleanup pass to remove stale setup-only guidance and old check counts, and the replay path could use a little more verbose progress and defense-in-depth around scratch-bound validation. None of those findings undercut the correctness of the shipped behavior.

**Dissent.** CLI logging and DevX UX both want a little more visible progress during self-hydration; performance wants more concurrency. I would treat those as polish follow-ups, not a reason to hold the bug fix.

**Aligned with:** secure_by_default: Lock-pinned replay trusts resolved_commit/resolved_url/resolved_hash without re-resolving refs; fail-closed on mismatch; governed_by_policy: CI audit no longer needs a warm cache to enforce governance; portable_by_manifest: replay materializes from manifest + lockfile alone; pragmatic_as_npm: `apm audit --ci` now maps to the deterministic npm-ci mental model.

**Growth signal.** This removes the last major friction point for CI-only adopters: teams no longer need a pre-warmed cache or `--no-drift` workaround to gate committed AI tool configs.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 1 | Clean single-owner design; only minor typing cleanup suggested. |
| CLI Logging Expert | 0 | 2 | 2 | Output is truthful, but self-hydration could expose better bounded progress and slightly clearer wording. |
| DevX UX Expert | 0 | 1 | 2 | Behavior now matches CI expectations; only progress and phrasing polish remain. |
| Supply Chain Security Expert | 0 | 1 | 1 | Lock-pinned replay is correct; one defense-in-depth scratch-bound check would tighten it further. |
| OSS Growth Hacker | 0 | 2 | 2 | Strong CI adoption unlock; docs should lead with the simpler user value. |
| Auth Expert | 0 | 0 | 1 | Auth delegation looks correct; just verify the wrapped replay error keeps remediation text intact. |
| Doc Writer | 0 | 4 | 0 | Several touched or adjacent docs still describe the old setup-only semantics. |
| Test Coverage Expert | 0 | 0 | 2 | The matrix is broad enough; CI should certify the integration tier on the new cold-cache e2e tests. |
| Performance Expert | 0 | 2 | 2 | Correctness wins, but cold-cache replay is currently conservative on concurrency and copy cost. |

### Top 5 follow-ups

1. **[Doc Writer]** Fix the stale docs claims in `enterprise/drift-detection.md`, `integrations/ci-cd.md`, `concepts/lifecycle.md`, and `consumer/drift-and-secure-by-default.md` -- the code changed, and these pages are the first places users land when wiring CI.
2. **[Supply Chain Security Expert]** Add `_assert_scratch_bound(project_root, scratch_root)` inside `run_replay()` even when `scratch_root` is supplied by a caller -- this closes a future-caller defense-in-depth gap.
3. **[CLI Logging Expert]** Emit one bounded stderr progress signal during self-hydration in verbose mode -- CI log readers currently see the heaviest phase happen silently.
4. **[Python Architect]** Add `PreparedCiAuditReplay | None` annotations to the new `prepared_replay` parameters in `ci_checks.py` -- cheap clarity win for static tooling.
5. **[OSS Growth Hacker]** Rewrite the `enforce-in-ci.md` comparison table to lead with the zero-config benefit, not the implementation detail.

### Recommendation

Merge as-is. The design is correct, the regression matrix holds, and the fail-closed behavior satisfies the approved scope. I would treat the docs cleanup as the highest-priority same-cycle follow-up, then take the scratch-bound hardening and verbose-progress polish in the next patch cycle.

---

<details>

#### Python Architect
- [recommended] Add explicit `PreparedCiAuditReplay | None` annotations to the new `prepared_replay` parameters in `ci_checks.py`.
- [nit] `atexit` cleanup is fine for CLI use, but would be worth revisiting if this path is ever reused in a long-lived daemon.

#### CLI Logging Expert
- [recommended] Make `replay_start` context-aware so users can tell cache-only from self-hydrating replay.
- [recommended] Let verbose users see a bounded progress signal during self-hydration instead of swallowing all stderr.
- [nit] Reword the `--frozen:` prefix in replay-failed messages; the user never passed that flag.
- [nit] Guarantee a non-empty wrapped replay error if `str(exc)` is empty.

#### DevX UX Expert
- [recommended] Surface one bounded progress line for slow self-hydration work in CI.
- [nit] Reword the replay mismatch message in user terms instead of internal flag terms.
- [nit] Keep the docs count synchronized everywhere the baseline set is described.

- [recommended] Re-run `_assert_scratch_bound()` in `run_replay()` even when the caller supplied the scratch path.
- [nit] Consider a helper-level containment assertion before `_clear_path()` / `_copy_install_tree()` writes into the scratch modules tree.

#### OSS Growth Hacker
- [recommended] Rewrite the audit-only comparison table in `enforce-in-ci.md` to lead with the user benefit.
- [recommended] Shorten the CI/CD dogfood note so the proof point is instantly scannable.
- [nit] Call out the removal of `--no-drift` from setup-only snippets in the release notes.
- [nit] Consider a friendlier user-facing synonym than `self-hydrates` in public docs.

#### Auth Expert
- [nit] Verify that wrapped replay download failures still surface the existing auth remediation context clearly in CI output.

#### Doc Writer
- [recommended] `enterprise/drift-detection.md` still says CI has eight baseline checks.
- [recommended] `integrations/ci-cd.md` still overstates how the repo's own dogfood workflow is wired.
- [recommended] `concepts/lifecycle.md` overstates drift scope by naming `apm_modules/` rather than deployed/generated outputs.
- [recommended] `consumer/drift-and-secure-by-default.md` still describes the old cold-cache audit behavior.

#### Test Coverage Expert
- [nit] CI should be the final integration-tier certifier for the new cold-cache e2e tests.
- [nit] The mismatch no-write unit test already covers the user-facing error path adequately.

#### Performance Expert
- [recommended] Cold-cache replay currently serializes downloads; this is acceptable now but will be the next wall-time pressure point.
- [recommended] Live-cache reuse copies whole trees instead of using a cheaper hardlink or clonefile path.
- [nit] `_git_tracked_files()` cost is fine because the result is cached in the prepared replay object.
- [nit] `AuthResolver()` could be lazily created on first cache miss in the future.

</details>

PR #2411, opened 2026-07-31
refactor(auth): centralize indexed Git config cleanup

### LINKED ISSUE #2398: Extract shared retain/reindex predicate for git auth-config entries (follow-up to #2368)
## Description

PR #2382 (fixing #2368) added `set_authorization_header_git_env()` / `set_ado_bearer_git_env()` in `src/apm_cli/utils/github_host.py`. These rewrite the indexed `GIT_CONFIG_{COUNT,KEY_N,VALUE_N}` env-var set in place: existing auth-channel entries are stripped, everything else is preserved and re-indexed, and the new `Authorization` header is appended last.

That retain/reindex predicate intentionally mirrors the pre-existing `AuthResolver._clear_git_auth_env` in `src/apm_cli/core/auth.py` (same policy: an entry whose key contains `extraheader`, or whose value is an `Authorization:` header, is dropped; everything else is retained and re-indexed).

This is a genuine split authority per `.apm/instructions/architecture.instructions.md`: the SAME decision ("which `GIT_CONFIG_*` entries are auth-channel entries to strip") is now implemented in two places. If the predicate ever needs to change (e.g. a new auth-channel key pattern), only one site is likely to get patched, and the two will silently drift.

## Why this was not done in #2382

`_clear_git_auth_env` is a security-critical function that PR #2382 does not otherwise touch. The PR (and the apm-review-panel review, python-architect + auth-expert personas both flagged the same duplication independently) intentionally deferred this extraction to keep the original bug fix small, reviewable, and low-risk, rather than bundling a refactor of `_clear_git_auth_env` into a fix for a different, narrower defect.

## Proposed follow-up

Extract a single shared helper (suggested home: `src/apm_cli/utils/git_env.py`, a new module) that both `AuthResolver._clear_git_auth_env` (`src/apm_cli/core/auth.py`) and `set_authorization_header_git_env` (`src/apm_cli/utils/github_host.py`) delegate to for:
- parsing `GIT_CONFIG_COUNT` (tolerating blank/invalid values),
- the auth-channel-entry predicate (`extraheader` in key, or value is an `Authorization:` header),
- retaining + re-indexing the non-auth-channel entries.

This PR should also update `.apm/instructions/architecture.instructions.md`'s canonical owner table to add a row for this newly-centralized decision (git auth-config retain/reindex), and add the dual guardrail required for a split-authority repair (behavioral regression test + a static boundary guard in `scripts/lint-architecture-boundaries.sh` + a matching `tests/integration/test_architecture_*.py` assertion + mutation-break evidence), since collapsing the two existing implementations into one is exactly the "split authority repair" case the architecture doc describes.

## Origin

Surfaced by the apm-review-panel (python-architect + auth-expert personas, independently converging) during shepherd-driver review of #2382. Cross-referenced via a `TODO` comment in `src/apm_cli/utils/github_host.py` next to `set_authorization_header_git_env`.

Pybsama:
refactor(auth): centralize indexed Git config cleanup

## TL;DR

Centralize ownership of indexed `GIT_CONFIG_*` authentication-header cleanup in
`apm_cli.utils.git_env`.

This removes two subtly different cleanup implementations from the authentication
resolver and the GitHub-host helper.

The shared owner now:

- identifies only `http.*.extraheader` entries whose value contains an actual
  `Authorization:` header field;
- preserves proxy, TLS, CA, and other unrelated Git configuration;
- handles LF, CRLF, and CR-separated header values;
- compacts retained entries into a stable contiguous sequence;
- removes orphaned indexed keys left by stale or malformed counts.

Fixes #2398.

## Problem

APM has two consumers that need to remove stale Git authentication headers before
constructing a child-process environment:

- `AuthResolver._clear_git_auth_env`;
- `set_authorization_header_git_env`.

Each consumer previously implemented its own scan, filtering, and reindexing.

The two implementations did not agree on what counted as an authentication entry.
One checked whether the entire value started with `authorization:` while the other
removed any value containing the substring `authorization`.

That split ownership created two concrete risks:

1. a multiline `extraheader` value could retain a stale credential when its
   Authorization field was not first;
2. an unrelated value containing text such as `x-authorization-mode` could be
   discarded even though it was not an Authorization header.

Because the environment is inherited by Git subprocesses, both false negatives and
false positives matter: stale credentials are a security boundary, while removed
proxy or TLS configuration can make a valid clone fail.

## Approach

Introduce one canonical helper that owns the indexed representation and make both
call sites delegate to it.

| Concern | Decision |
| --- | --- |
| Header recognition | Parse line boundaries and match the exact `Authorization` field name |
| Line endings | Accept LF, CRLF, and CR |
| Whitespace | Allow optional horizontal whitespace around the field name and colon |
| Key scope | Restrict removal to `http.*.extraheader` entries |
| Retention | Preserve every non-authentication entry without modifying its value |
| Reindexing | Rewrite retained pairs from index zero in their original order |
| Stale metadata | Remove indexed keys that are not represented by a trustworthy count |
| Consumer behavior | Keep direct credential channels local to the resolver |

The helper stages the new indexed state before mutating the environment. This
avoids leaving a partially rewritten sequence if the input contains malformed
count metadata.

### Canonical owner

[`src/apm_cli/utils/git_env.py`](https://github.com/microsoft/apm/blob/3e97d647fda69aa97cfc1e22aea456b55c2a88ba/src/apm_cli/utils/git_env.py#L53-L129)
now provides:

- `_is_git_auth_config_entry(key, value)` for exact field recognition;
- `strip_git_auth_config_entries(env)` for filtering, cleanup, and stable
  compaction.

The function returns the next free index so a caller that needs to append a new
header can do so without recomputing the sequence.

It treats a missing, negative, non-integer, or otherwise stale count defensively
and still inspects actual indexed key slots present in the environment.

### Consumers

[`AuthResolver._clear_git_auth_env`](https://github.com/microsoft/apm/blob/3e97d647fda69aa97cfc1e22aea456b55c2a88ba/src/apm_cli/core/auth.py#L963-L969)
delegates indexed cleanup to the shared helper,
then continues to clear the resolver-owned direct credential channels.

[`set_authorization_header_git_env`](https://github.com/microsoft/apm/blob/3e97d647fda69aa97cfc1e22aea456b55c2a88ba/src/apm_cli/utils/github_host.py#L501-L537)
rejects CR/LF in the new token, delegates stale
indexed cleanup, and appends exactly one fresh Authorization header at the returned
index.

Its contract now explicitly describes that sole indexed Git authentication header.

### Ownership guard

[`scripts/check_git_auth_config_owner.py`](https://github.com/microsoft/apm/blob/3e97d647fda69aa97cfc1e22aea456b55c2a88ba/scripts/check_git_auth_config_owner.py#L1-L164)
establishes the helper as the only
production owner of indexed Git authentication cleanup.

The architecture lint invokes this checker as AC20.

The checker understands relevant syntax rather than relying on a broad text search,
so comments, fixtures, and approved helper implementation details do not create
false positives.

The owner declaration is mirrored in `.apm`, `.github`, and the lock file so the
published and repository-local architecture guidance remain aligned.

### Tests

The change adds focused tests for:

- exact header-field matching;
- LF, CRLF, and CR multiline values;
- optional horizontal whitespace;
- safe values that merely contain authorization-related text;
- malformed and stale count metadata;
- orphan-key cleanup;
- stable reindexing and return values;
- resolver delegation and direct-channel cleanup;
- setter replacement and injection rejection;
- the architecture owner boundary.

## Diagrams

```mermaid
flowchart LR
    subgraph Consumers["Consumers"]
        C1["AuthResolver._clear_git_auth_env"]
        C2["set_authorization_header_git_env"]
    end

    subgraph Owner["Canonical owner"]
        O1["strip_git_auth_config_entries"]:::new
        O2["_is_git_auth_config_entry"]:::new
    end

    subgraph Result["Indexed environment"]
        R1["Stable retained entries"]
        R2["Fresh header appended by setter"]
    end

    C1 --> O1
    C2 --> O1
    O1 --> O2
    O1 --> R1
    O1 --> C2
    C2 --> R2

    classDef new stroke-dasharray: 5 5
```

The dashed nodes are introduced by this change. Existing consumers now share the
same filtering and compaction path before continuing with their own responsibilities.

## Trade-offs

- Exact header-field matching is deliberately narrower than a substring search.
  This preserves vendor-neutral configuration while still rejecting a real
  Authorization field anywhere in a multiline value.
- The helper scans the indexed slots actually present in the environment in
  addition to a valid declared range. This makes stale metadata recoverable without
  trusting an arbitrarily large count.
- Direct credential variables remain resolver-owned. Moving those channels into
  the indexed-config utility would broaden its contract beyond issue #2398.
- The architecture checker adds a small maintenance cost, but turns an otherwise
  informal ownership rule into an executable boundary.

## Benefits

- One implementation defines the security-sensitive cleanup behavior.
- Both credential paths preserve unrelated proxy, TLS, CA, and hardening settings.
- Multiline stale Authorization fields cannot bypass cleanup.
- Safe values containing authorization-related words are no longer over-filtered.
- Future duplicate cleanup loops fail the architecture check before review.
- Focused regression tests describe the retained and removed cases independently.

### Automated checks

- Focused Git-auth test selection:
  `119 passed in 0.40s`
- Full unit suite:
  `19098 passed, 2 skipped, 21 xfailed, 19 warnings, 87 subtests passed in 162.63s`
- Ruff check:
  `All checks passed!`
- Ruff format:
  `1550 files already formatted`
- Pylint:
  `Your code has been rated at 10.00/10`
- Authentication-signal lint:
  `[+] auth-signal lint clean`
- Architecture lint AC1-AC20:
  `[+] architecture boundary lint clean`
- Package build:
  source distribution and wheel both built successfully with `uv build`
- Patch hygiene:
  `git diff --check` completed with no output

### Mutation evidence

| Deliberate regression | Detection |
| --- | --- |
| Replace exact field matching with broad substring matching | 5 focused tests fail |
| Change extraheader key scope from OR to AND | 6 focused tests fail |
| Reintroduce a consumer-owned cleanup loop | AC20 fails |

All mutations were temporary and the correct implementation was restored before
the full validation run.

### Scenario Evidence

| Scenario | Principle demonstrated | Test level | Regression trap |
| --- | --- | --- | --- |
| A bearer retry inherits corporate proxy and CA entries while stale auth is removed | Secure by default; vendor-neutral behavior | Unit | Resolver preservation test for #2398 |
| The setter replaces an old indexed header and retains hardening entries | Secure by default; predictable developer experience | Unit | Setter replacement and ordering tests |
| An Authorization field follows another field in one multiline value | Secure by default | Unit | LF, CRLF, and CR multiline tests |
| A second production consumer attempts to own indexed cleanup | Community-maintainable architecture | Integration lint | AC20 mutation test |

## How to test

From the repository root:

1. Run the focused regression set:
   `uv run pytest tests/unit/test_git_env.py tests/unit/test_auth.py tests/unit/test_github_host.py -q`.
2. Run the full suite:
   `uv run pytest tests/unit -q`.
3. Run formatting and lint checks:
   `uv run ruff check .` and `uv run ruff format --check .`.
4. Run the repository authentication and architecture lint scripts.
5. Build both package artifacts with `uv build`.

## Repository checklist

- [x] Maintenance/refactor change
- [x] Tested locally
- [x] Existing tests pass
- [x] New regression tests added
- [x] OpenAPM specification change: N/A
- [x] Architecture ownership metadata synchronized
- [x] Changelog entry references #2411
- [x] No new runtime dependency

### DIFF
--- a/.apm/instructions/architecture.instructions.md
+++ b/.apm/instructions/architecture.instructions.md
@@ -38,6 +38,7 @@ semicolon-delimited, and specific to the file(s) that own the fact.
 |---|---|---|
 | Accepted target vocabulary | core/target_catalog.py | `src/apm_cli/core/target_catalog.py` |
 | Host + credential resolution | core/auth.py (AuthResolver), core/host_providers.py | `src/apm_cli/core/auth.py`; `src/apm_cli/core/host_providers.py` |
+| Indexed Git auth-config retain/reindex policy | utils/git_env.py (strip_git_auth_config_entries) | `src/apm_cli/utils/git_env.py` |
 | Runtime descriptors | runtime/registry.py | `src/apm_cli/runtime/registry.py` |
 | User-facing output / diagnostics | CommandLogger / console owner | `src/apm_cli/core/command_logger.py`; `src/apm_cli/utils/console.py` |
 | Compiled-output writes (atomic) | CompiledOutputWriter | `src/apm_cli/compilation/output_writer.py` |
--- a/.github/instructions/architecture.instructions.md
+++ b/.github/instructions/architecture.instructions.md
@@ -38,6 +38,7 @@ semicolon-delimited, and specific to the file(s) that own the fact.
 |---|---|---|
 | Accepted target vocabulary | core/target_catalog.py | `src/apm_cli/core/target_catalog.py` |
 | Host + credential resolution | core/auth.py (AuthResolver), core/host_providers.py | `src/apm_cli/core/auth.py`; `src/apm_cli/core/host_providers.py` |
+| Indexed Git auth-config retain/reindex policy | utils/git_env.py (strip_git_auth_config_entries) | `src/apm_cli/utils/git_env.py` |
 | Runtime descriptors | runtime/registry.py | `src/apm_cli/runtime/registry.py` |
 | User-facing output / diagnostics | CommandLogger / console owner | `src/apm_cli/core/command_logger.py`; `src/apm_cli/utils/console.py` |
 | Compiled-output writes (atomic) | CompiledOutputWriter | `src/apm_cli/compilation/output_writer.py` |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
+- Indexed Git authentication cleanup now strips actual `Authorization` headers, including multiline values, while preserving unrelated proxy, TLS, CA, and hardening entries — by @Pybsama (#2411)
 - Corporate git security settings -- SSL CA pins (`http.sslCAInfo`), bare-repo
   protection (`safe.bareRepository=explicit`), and other inherited `GIT_CONFIG_*`
   hardening -- are no longer silently dropped when apm injects an
--- a/apm.lock.yaml
+++ b/apm.lock.yaml
@@ -2783,7 +2783,7 @@ deployments:
   owners:
   - .
   active_owner: .
-  content_hash: sha256:c8cd4b941415ffae010102fa6936c0654e3fe55a80b114af2ba7076e2c5e572e
+  content_hash: sha256:e859b2db5e892a9e97be74366238a68d9f0be0437e0ce9f820d49cb5542214e7
 - kind: project-relative
   target: copilot
   value: .github/instructions/changelog.instructions.md
@@ -3290,7 +3290,7 @@ local_deployed_file_hashes:
   .github/agents/spec-tag-architect.agent.md: sha256:82907265c5e7cf1ac61ad96866fa7c5683b69c8f09b7a4c5f3cc241acc9568ca
   .github/agents/supply-chain-security-expert.agent.md: sha256:8fb8cc426d6af17ba084a28b3f026c2b475b62e3ca63ed2f88b83bd823f877af
   .github/agents/test-coverage-expert.agent.md: sha256:48c2172d1f18a394fa83ef9dc2be0b9b921a4e51e976498165250fed66369711
-  .github/instructions/architecture.instructions.md: sha256:c8cd4b941415ffae010102fa6936c0654e3fe55a80b114af2ba7076e2c5e572e
+  .github/instructions/architecture.instructions.md: sha256:e859b2db5e892a9e97be74366238a68d9f0be0437e0ce9f820d49cb5542214e7
   .github/instructions/changelog.instructions.md: sha256:1e51ec4c74e847967962bd279dc4c6e582c5d3578490b3c28d5f3acd3e05f73e
   .github/instructions/cicd.instructions.md: sha256:08d87b7d635761cb41deb8fc71d5d83f54678de463db484afb16d2d4f8713ecb
   .github/instructions/cli.instructions.md: sha256:8e39e8d5047ce88575cb02f87c2bcede584dfef258bd86f7466c7badf136541a
--- a/scripts/check_git_auth_config_owner.py
+++ b/scripts/check_git_auth_config_owner.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python3
+"""Enforce one owner for indexed Git auth-config retain/reindex policy."""
+
+from __future__ import annotations
+
+import argparse
+import ast
+from pathlib import Path
+
+OWNER_MODULE = "apm_cli.utils.git_env"
+OWNER_SYMBOL = "strip_git_auth_config_entries"
+OWNER_PATH = Path("src/apm_cli/utils/git_env.py")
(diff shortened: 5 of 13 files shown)


Pybsama:
Closing this PR to avoid duplicate review work.

[#2544](https://github.com/microsoft/apm/pull/2544) now implements the same #2398 canonical-owner extraction, has an assignee, and has received a collaborator-hosted advisory review. This PR predates #2544 but has not been merged, so keeping both open would leave maintainers with two conflicting implementations of the same scoped refactor.

The public branch and commits remain available. They include additional edge-case coverage for exact `Authorization` field matching across CR/LF-separated values, preservation of ordinary authorization-related text, bounded handling of malformed indexed counts, orphan cleanup, and a consumer-loop mutation guard. Those pieces can still be reused if they are useful to the implementation that proceeds.

PR #2461, opened 2026-08-03
fix(marketplace): effective output path (supersedes #2448, closes #2427)

danielmeppiel:
# fix(marketplace): use effective output path for clean checks

## TL;DR

`apm pack --check-clean` now resolves marketplace artifacts through the same
owner as normal pack output, including `--marketplace-path` overrides. This
supersedes #2448 and closes #2427 after rebasing on current `main`, folding the
panel's command-guide consolidation, and preserving the original author's work.

> [!NOTE]
> Closes #2427. Supersedes #2448.

## Problem (WHY)

- [x] Pack wrote an overridden Claude artifact to `.github/plugin/marketplace.json`
  while clean checking selected the default Claude path.
- [x] The reported contract is explicit: ["The clean-check should use the same
  effective output path as normal pack output generation."](https://github.com/microsoft/apm/issues/2427)
- [!] Separate writer and drift selectors allowed precedence changes to reach only
  one flow.

## Approach (WHAT)

| # | Fix |
|---|-----|
| 1 | Centralize override, configuration, and default precedence in `resolve_effective_output_path()`. |
| 2 | Route producer, builder compatibility paths, and drift checking through that owner. |
| 3 | Pass CLI overrides to the clean gate and reject future parallel selectors with AC31. |
| 4 | Cover the bug with integration, installed-binary lifecycle, and architecture tests. |

## Implementation (HOW)

- **`src/apm_cli/marketplace/output_profiles.py`** owns effective output-path
  resolution and project-root containment.
- **`src/apm_cli/{core/build_orchestrator.py,marketplace/builder.py,marketplace/drift_check.py}`**
  consume that resolver for writes and clean comparisons.
- **`src/apm_cli/commands/pack.py`** forwards marketplace overrides into the
  clean gate.
- **`scripts/check_marketplace_output_path_authority.sh`** and its integration
  tests prevent a second selector; docs and the usage guide describe the
  observable contract once.

## Diagrams

Legend: the dashed resolver is the new common decision point for both artifact
writing and clean verification.

```mermaid
flowchart LR
    A["apm pack --marketplace-path"] --> B["pack_cmd"]
    B --> C["MarketplaceProducer"]
    C --> D["resolve_effective_output_path"]
    D --> E["write effective artifact"]
    B --> F["check_marketplace_drift"]
    F --> D
    D --> G["compare effective artifact"]
    classDef new stroke-dasharray: 5 5;
    class D new;
```

## Trade-offs

[...]

PR #2481, opened 2026-08-05
fix(install): re-download aliased deps after a manifest ref bump

danielmeppiel:
## TL;DR

Fixes the three release-blocking integration failures that surfaced on the
`v0.28.0` tag run. Two are product bugs (aliased-dependency updates were
silently ignored; `--target all` crashed on `grok-build`); one is stale test
fixtures.

## Problem (WHY)

The `v0.28.0` release run reached the full `tests/integration` stage for the
first time this cycle and failed four tests across the macOS and Linux jobs:

| Test | Symptom |
| --- | --- |
| `test_kiro_agent_lifecycle.py::test_kiro_agent_update_replacement` | Deployed agent stays at v1 after the manifest `ref:` is bumped to v2 |
| `test_intellij_target.py::...[all,intellij-True]` | `Unsupported client type: grok-build` |
| `test_marketplace_builder_hermetic.py::TestOutputPath::test_output_override_used_when_no_marketplace_output` | `MarketplaceYmlError: 'name' is required` |
| `test_marketplace_builder_phase3w4.py::TestOutputPath::...` | same |

### 1. Aliased dependency updates were silently ignored

Two competing install-path derivations exist. The BFS resolver, lockfile
`get_installed_paths()`, drift detection, and cache pinning all use the
alias-blind canonical path `apm_modules/<org>/<repo>`. The download and
integration phases instead use `apm_modules/<alias>` when an alias is
declared.

When the user bumps `ref:` in `apm.yml`, `detect_ref_change` correctly
returns `True` and the BFS callback refreshes the *canonical* directory. But
the integration loop reads from the *alias* directory, and
`_compute_skip_download` returned `True` there via either the
`is_cacheable and not update_refs` clause (a commit pin is always cacheable)
or the `already_resolved and not update_refs` clause (the callback did fetch
-- just to a different path). The stale alias bytes were then deployed.

### 2. `--target all` crashed on `grok-build`

`grok-build` was added to `TARGET_CAPABILITIES` with `in_all=True` in #2420
but has no entry in `_MCP_CLIENT_REGISTRY`. `EffectiveTargetDecision.runtime_targets`
returns every target's runtime name with no MCP-capability filter, so MCP
install hard-failed on `--target all`.

### 3. Stale marketplace fixtures

#2461 changed `MarketplaceBuilder._output_path()` to call `_load_yml()`,
which runs full schema validation. The two builder test fixtures wrote a
minimal yml with none of the required keys and used the snake_case
`tag_pattern` instead of `tagPattern`.

## Approach (WHAT)

Surgical, one fix per failure. Notably this does **not** unify the two
install-path derivations -- that was tried and regressed
`test_hook_js_sidecar_lifecycle_contract.py` (uninstall + reinstall dropped
back to the v1 hook script). Vetoing the cache-reuse clauses on ref drift is
the minimal correct change: it makes the alias/canonical split irrelevant to
correctness without touching path derivation.

## Implementation (HOW)

**`_skip_logic.py` / `integrate.py`** -- `_compute_skip_download` takes a new
`ref_changed: bool = False` parameter that vetoes the two cache-reuse
clauses:

```python
return install_path_exists and (
    (is_cacheable and not update_refs and not ref_changed)
    or (already_resolved and not update_refs and not ref_changed)
    or lockfile_match
)
```

`lockfile_match` is deliberately left ungated: in normal mode it is only
computed under `elif not ref_changed`, and in update mode only after the
resolved remote SHA was confirmed equal to the lockfile SHA. The parameter
defaults to `False` so no existing call site changes behaviour.

**`mcp_integrator_install.py`** -- `_resolve_target_runtimes` filters
resolved targets against `ClientFactory.supported_clients()` (the documented
single source of truth for MCP capability) unless the selection came from an
explicit runtime choice, in which case the user asked for that target by name
and deserves the hard error. Skipped targets are logged; if nothing
MCP-capable remains, it warns and returns `None`. This mirrors the existing
user-scope filter pattern in the same module.

**Test fixtures** -- added `name`, `description`, `version`, and
`owner: {name: ...}` to both `_make_marketplace_yml_file` defaults, and
renamed `tag_pattern` to `tagPattern`.

## Trade-offs

- **Chose:** veto cache reuse on ref drift. **Over:** unifying the two
  install-path derivations. The unification is the deeper fix but regressed
  the hook sidecar lifecycle contract, and a release-blocking hotfix is the
  wrong place to relitigate path derivation. The dual-directory split remains
  and should be tracked separately.
- **Chose:** filter non-MCP targets with a note. **Over:** adding `grok-build`
  to `_MCP_CLIENT_REGISTRY`. `grok-build` genuinely has no MCP client; a stub
  entry would be a lie that fails later and worse.
- **Cost:** a ref bump now always re-downloads even when the new ref happens
  to resolve to identical bytes. Correctness over one avoidable fetch.

## Validation evidence

| Check | Result |
| --- | --- |
| `tests/integration/test_kiro_agent_lifecycle.py` + `test_hook_js_sidecar_lifecycle_contract.py` | 12 passed |
| Marketplace builder suites | 132 passed |
| IntelliJ target suite | passed |
| Full `tests/unit` | 19790 passed, 2 skipped, 21 xfailed |
| Full `tests/integration` | 8417 passed; residual failures are local-harness artifacts (venv console script is a Python shim, so PyInstaller-prefix assertions fail) and xdist collisions on repo-relative `apm_modules/` -- all pass in isolation |
| `ruff check` / `ruff format --check` | clean |
| `pylint R0801` | 10.00/10 |
| `scripts/lint-auth-signals.sh` | clean |

## How to test

```bash
APM_E2E_TESTS=1 APM_BINARY_PATH="$PWD/.venv/bin/apm" \
  uv run --extra dev python -m pytest \
  tests/integration/test_kiro_agent_lifecycle.py \
  tests/integration/test_hook_js_sidecar_lifecycle_contract.py \
  tests/integration/test_marketplace_builder_hermetic.py \
  tests/integration/test_marketplace_builder_phase3w4.py \
  tests/integration/test_intellij_target.py -q
```

Unit regression traps for the new veto live in
`tests/unit/test_install_update_refs.py`.

apm-spec-waiver: bug fixes restoring already-specified install and MCP-target behaviour; no new normative surface

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,10 +34,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `apm install` now re-downloads an aliased dependency after its `ref:` changes
   in `apm.yml`. Previously the pinned-ref and already-resolved cache-reuse
   shortcuts skipped the fetch, so `apm_modules/<alias>/` and every primitive
-  deployed from it stayed on the old revision. (closes #2481)
+  deployed from it stayed on the old revision. (#2481)
 - `apm install --target all` no longer aborts on targets that have no MCP
   client, such as `grok-build`. Non-MCP targets are skipped with a note instead
-  of raising `Unsupported client type`. (closes #2481)
+  of raising `Unsupported client type`. (#2481)
 - `apm init` and install-time auto-bootstrap now reject empty or
   whitespace-only project names, falling back to `my-project` at a filesystem
   root. APM no longer writes an `apm.yml` that every later install, lock, or
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `apm install` now re-downloads an aliased dependency after its `ref:` changes
+  in `apm.yml`. Previously the pinned-ref and already-resolved cache-reuse
+  shortcuts skipped the fetch, so `apm_modules/<alias>/` and every primitive
+  deployed from it stayed on the old revision. (closes #2481)
+- `apm install --target all` no longer aborts on targets that have no MCP
+  client, such as `grok-build`. Non-MCP targets are skipped with a note instead
+  of raising `Unsupported client type`. (closes #2481)
 - `apm init` and install-time auto-bootstrap now reject empty or
   whitespace-only project names, falling back to `my-project` at a filesystem
   root. APM no longer writes an `apm.yml` that every later install, lock, or
--- a/src/apm_cli/install/phases/_skip_logic.py
+++ b/src/apm_cli/install/phases/_skip_logic.py
@@ -15,6 +15,7 @@ def _compute_skip_download(
     update_refs: bool,
     already_resolved: bool,
     lockfile_match: bool,
+    ref_changed: bool = False,
 ) -> bool:
     """Return True when the sequential loop should skip downloading a package.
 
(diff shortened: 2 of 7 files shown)


danielmeppiel:
Superseded by a fresh PR from the same branch: the branch-protection rule requires the last push to be reviewed, and Copilot only reviews on PR open.

PR #2515, opened 2026-08-06
fix(integration): deploy namespaced prompts instead of silently dropping them

kamilpyc-Pi:
## Description

A prompt authored at `.apm/prompts/<ns>/<name>.prompt.md` is never deployed and never
reported. `PromptIntegrator.find_prompt_files` and `CommandIntegrator.find_prompt_files`
both glob `.apm/prompts` exactly one level deep:

```python
# prompt_integrator.py / command_integrator.py -- before
find_files_by_glob(package_path, "*.prompt.md", subdirs=[".apm/prompts"])
```

Anything in a subdirectory is invisible: no warning, no diagnostic, absent from the
lockfile. The install summary just reports a lower count than the package contains.

Agents never had this problem — `AgentIntegrator.find_agent_files` has always searched
`.apm/agents/` with `**/*.agent.md`. Prompts were the odd type out, which is what
suggests an oversight rather than a design stance.

**It also loses content on a supported input path.** `plugin_parser` maps a Claude Code
plugin's `commands/` directory into `.apm/prompts/`, walking it with `rglob("*")` and
preserving subdirectories via `rel_to=source`. Claude plugins routinely namespace
commands that way, so APM faithfully copies `commands/<ns>/x.md` to
`.apm/prompts/<ns>/x.prompt.md` — and then discards it at integration time. Installing
such a plugin silently yields fewer commands than the plugin ships.

Beyond fixing the loss, preserving the subdirectory is what makes namespaced commands
expressible at all: Claude Code reads `commands/opsx/propose.md` as `/opsx:propose`.

### Reproduction

A package with three prompts:

```
.apm/prompts/flat.prompt.md
.apm/prompts/opsx-propose.prompt.md
.apm/prompts/opsx/propose.prompt.md
```

`apm install` against the `claude` target, before this change:

```
2 commands integrated -> .claude/commands/
.claude/commands/flat.md
.claude/commands/opsx-propose.md
```

The third prompt is gone, with nothing said about it. After:

```
3 commands integrated -> .claude/commands/
.claude/commands/flat.md
.claude/commands/opsx-propose.md
.claude/commands/opsx/propose.md
```

The new file is recorded in the lockfile's `deployed_files` with its hash, so uninstall
and `apm audit` track it like any other deployed file.

## Changes

- Search `.apm/prompts/` recursively in both integrators, mirroring `AgentIntegrator`.
  The package root stays a **flat** search, so an unrelated subtree (`docs/`,
  `examples/`) is not swept in — there is a regression test for this.
- Add `BaseIntegrator.primitive_namespace()`, returning the subdirectory components of a
  source file relative to its primitive directory. Every component goes through
  `validate_path_segments`; a file that is flat, or outside the primitive directory,
  yields `()` so callers can join unconditionally. `ensure_path_within` at each call site
  remains the second line of defence.
- Join that namespace onto the deploy directory in both integrators.
- Create the target's parent directory in `PromptIntegrator` before writing —
  `write_text_lf` documents that the caller must ensure the parent exists. The command
  writers (`_write_claude_command`, `_write_gemini_command`) already did this.

### Compatibility

Existing packages are unaffected. A flat prompt yields an empty namespace and lands at
exactly the path it always has. The only files whose behaviour changes are the ones that
were previously dropped, which go from invisible to deployed.

### Deliberately not changed

The copilot-app path deploys workflow rows into SQLite keyed by name rather than files,
where a directory namespace has no meaning. `CopilotAppWorkflowIntegrator` keeps its flat
discovery.

## Open question for maintainers

I preserve the subdirectory for every **file-based** target. That is unambiguously right
for Claude Code (`commands/opsx/propose.md` → `/opsx:propose`). Targets whose prompt
directory has no namespace concept — Copilot's `.github/prompts/`, for instance — will
receive the file nested, where it may not be picked up.

That is still strictly better than today, where the file is not written at all, but a
per-target policy (preserve / flatten-with-separator / reject) is a design call that
belongs to you rather than to this PR. Happy to add a `namespaced: bool` field to the
target registry and gate on it if you prefer that shape.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

Full unit suite as CI runs it (`uv run pytest tests/unit tests/test_console.py`):
**19805 passed, 2 skipped, 21 xfailed**. `ruff check` and `ruff format --check` clean.

Six tests added:

| Test | Covers |
| --- | --- |
| `test_find_prompt_files_recurses_into_apm_prompts_subdirs` | nested discovery |
| `test_find_prompt_files_does_not_recurse_package_root` | guard against over-recursion |
| `test_primitive_namespace_components` | flat, nested, deeper, outside-the-dir |
| `test_nested_prompt_keeps_its_namespace` | end-to-end deploy path |
| `test_nested_prompt_content_is_transformed_normally` | namespacing changes path only |

I reverted `src/` and re-ran them to confirm they are load-bearing: four fail without the
source change. The remaining two are regression guards that pass either way, by design.

Reasoning, so you can disagree knowingly rather than assume the box was ticked
carelessly: `req-tg-002` requires deploying only **under** the registered deploy root,
and `.claude/commands/opsx/propose.md` is under `.claude/commands/`. No requirement
mandates a flat layout within a root, and §8.1 says only that primitives are "hoisted
individually into deploy directories" without constraining the resulting path shape.
`req-pr-001` covers source attribution rather than discovery recursion. If you read
§8.5 as fixing the intra-root layout, say so and I will run the three-step ritual.

### DIFF
--- a/src/apm_cli/integration/base_integrator.py
+++ b/src/apm_cli/integration/base_integrator.py
@@ -12,7 +12,11 @@ from apm_cli.core.deployment_state import MaterializationResult
 from apm_cli.primitives.discovery import discover_primitives
 from apm_cli.utils.atomic_io import normalize_crlf_to_lf
 from apm_cli.utils.console import _rich_warning
-from apm_cli.utils.path_security import PathTraversalError, ensure_path_within
+from apm_cli.utils.path_security import (
+    PathTraversalError,
+    ensure_path_within,
+    validate_path_segments,
+)
 
 
 def _managed_absolute_target_root(candidate: Path, targets: Any) -> Path | None:
@@ -962,3 +966,42 @@ class BaseIntegrator:
                     results.append(f)
 
         return results
+
+    @staticmethod
+    def primitive_namespace(source_file: Path, primitive_dir: Path) -> tuple[str, ...]:
+        """Return the sub-directory components of *source_file* under *primitive_dir*.
+
+        A primitive authored at ``.apm/prompts/opsx/propose.prompt.md`` carries the
+        namespace ``("opsx",)``; one authored flat at ``.apm/prompts/propose.prompt.md``
+        (or in the package root) carries ``()``.  Callers join the namespace onto the
+        deploy directory so the layout an author wrote is the layout the harness sees --
+        Claude Code, for one, reads ``commands/opsx/propose.md`` as ``/opsx:propose``.
+
+        Returns an empty tuple when *source_file* is not under *primitive_dir*, so a
+        caller can always join unconditionally.  Each component is validated: a
+        traversal segment yields ``()`` rather than an escaping path, and
+        ``ensure_path_within`` at the call site remains the second line of defence.
+
+        Args:
+            source_file: Absolute path to the discovered primitive file.
+            primitive_dir: Absolute path to the directory the namespace is relative to.
+
+        Returns:
+            Tuple of directory components, outermost first.
+        """
+        try:
+            relative = source_file.relative_to(primitive_dir)
+        except ValueError:
+            return ()
+
+        parts = relative.parts[:-1]
+        if not parts:
+            return ()
+
+        try:
+            for part in parts:
+                validate_path_segments(part, context="primitive namespace")
+        except PathTraversalError:
+            return ()
+
+        return parts
--- a/src/apm_cli/integration/command_integrator.py
+++ b/src/apm_cli/integration/command_integrator.py
@@ -187,8 +187,17 @@ class CommandIntegrator(BaseIntegrator):
         return True
 
     def find_prompt_files(self, package_path: Path) -> list[Path]:
-        """Find all .prompt.md files in a package."""
-        return self.find_files_by_glob(package_path, "*.prompt.md", subdirs=[".apm/prompts"])
+        """Find all .prompt.md files in a package.
+
+        The package root is searched flat; ``.apm/prompts/`` is searched recursively so
+        a namespaced command survives to the target.  See PromptIntegrator for why the
+        two halves differ.
+        """
+        files = self.find_files_by_glob(package_path, "*.prompt.md")
+        apm_prompts = package_path / ".apm" / "prompts"
+        if apm_prompts.exists():
+            files += self.find_files_by_glob(apm_prompts, "**/*.prompt.md")
+        return files
 
     def _transform_prompt_to_command(
         self,
@@ -510,6 +519,7 @@ class CommandIntegrator(BaseIntegrator):
         self.init_link_resolver(package_info, project_root)
 
         commands_dir = target_root / mapping.subdir
+        pkg_prompts_dir = package_info.install_path / ".apm" / "prompts"
         files_integrated = 0
         files_skipped = 0
         files_adopted = 0
@@ -556,7 +566,12 @@ class CommandIntegrator(BaseIntegrator):
                 files_skipped += 1
                 continue
 
-            target_path = commands_dir / f"{base_name}{mapping.extension}"
+            # Preserve the author's sub-directory under .apm/prompts/.  Harnesses that
+            # namespace commands by directory read it back: Claude Code surfaces
+            # commands/opsx/propose.md as /opsx:propose.  A flat prompt yields () and
+            # lands exactly where it always has.
+            namespace = self.primitive_namespace(prompt_file, pkg_prompts_dir)
+            target_path = commands_dir.joinpath(*namespace, f"{base_name}{mapping.extension}")
 
             # Defense-in-depth: assert the resolved target stays inside
             # the commands directory even if validate_path_segments was
--- a/src/apm_cli/integration/prompt_integrator.py
+++ b/src/apm_cli/integration/prompt_integrator.py
@@ -22,8 +22,13 @@ class PromptIntegrator(BaseIntegrator):
         """Find all .prompt.md files in a package.
 
         Searches in:
-        - Package root directory
-        - .apm/prompts/ subdirectory
+        - Package root directory (flat)
+        - .apm/prompts/ subdirectory, recursively
+
+        The recursive half mirrors AgentIntegrator.find_agent_files, which has always
+        used ``**/*.agent.md`` under ``.apm/agents/``.  Prompts were the odd type out:
+        a package authoring ``.apm/prompts/opsx/propose.prompt.md`` had that file
+        silently dropped, never deployed and never reported.
 
         Args:
             package_path: Path to the package directory
@@ -31,7 +36,11 @@ class PromptIntegrator(BaseIntegrator):
         Returns:
             List[Path]: List of absolute paths to .prompt.md files
         """
-        return self.find_files_by_glob(package_path, "*.prompt.md", subdirs=[".apm/prompts"])
+        files = self.find_files_by_glob(package_path, "*.prompt.md")
+        apm_prompts = package_path / ".apm" / "prompts"
+        if apm_prompts.exists():
+            files += self.find_files_by_glob(apm_prompts, "**/*.prompt.md")
+        return files
 
     def copy_prompt(self, source: Path, target: Path) -> int:
         """Copy prompt file verbatim with link resolution.
@@ -284,7 +293,13 @@ class PromptIntegrator(BaseIntegrator):
                 continue
 
             target_filename = self.get_target_filename(source_file, package_info.package.name)
-            target_path = prompts_dir / target_filename
+            # Preserve the author's sub-directory under .apm/prompts/ so a namespaced
+            # prompt keeps its namespace at the target.  Flat prompts get () here and
+            # land exactly where they always have.
+            namespace = self.primitive_namespace(
+                source_file, package_info.install_path / ".apm" / "prompts"
+            )
+            target_path = prompts_dir.joinpath(*namespace, target_filename)
             # Defense-in-depth: target_filename is derived from source
             # file name; assert containment under prompts_dir to mirror
             # the guard already present in command/instruction
@@ -313,6 +328,11 @@ class PromptIntegrator(BaseIntegrator):
                 files_skipped += 1
                 continue
 
+            # Namespaced prompts deploy into a sub-directory that may not exist yet;
+            # write_text_lf requires the parent.  Created only once the file has
+            # cleared adoption and collision checks, so a skipped prompt leaves no
+            # empty directory behind.
+            target_path.parent.mkdir(parents=True, exist_ok=True)
             links_resolved = self.copy_prompt(source_file, target_path)
             total_links_resolved += links_resolved
             files_integrated += 1
--- a/tests/unit/integration/test_command_integrator.py
+++ b/tests/unit/integration/test_command_integrator.py
@@ -1717,3 +1717,72 @@ class TestCursorCommandPanelFindings:
         assert any(i.severity == "critical" for i in sec_items), (
             f"expected critical security diagnostic, got: {[(i.severity, i.message) for i in sec_items]}"
         )
+
+
+class TestNamespacedCommands:
+    """A prompt under .apm/prompts/<ns>/ keeps <ns> at the target.
+
+    Claude Code reads ``commands/opsx/propose.md`` as ``/opsx:propose``.  Before
+    recursive discovery landed, the nested source was never found at all: it was
+    silently dropped, so the namespaced command simply did not exist.
+    """
+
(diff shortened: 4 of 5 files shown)


kamilpyc-Pi:

sergio-sisternes-epam:

> Namespaced prompt discovery fix is sound and unblocks Claude Code plugin authors; ship with test parity and CHANGELOG followups.

cc @kamilpyc-Pi @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

The fix is architecturally correct: `primitive_namespace` on `BaseIntegrator` is well-placed, path-traversal validation is two-layer and correctly ordered, and the recursive discovery change is minimal and precise. No panelist disputes the direction or the implementation quality. The only strategic tension is between shipping now (unblocking a high-value adopter segment) and closing the test-parity gap that leaves `PromptIntegrator`'s namespace path unexercised at integration level while `CommandIntegrator`'s analogous path IS tested.

The test-coverage-expert's finding carries elevated weight here: the evidence block declares `outcome:missing` on a devx-principle surface, and the analogous `CommandIntegrator` path proves the test pattern already exists and is trivial to replicate. This is not speculative coverage debt -- it is a concrete regression-trap gap on a freshly changed code path. The CHANGELOG and docs gap (converged on independently by doc-writer and oss-growth-hacker) is the second priority: without it, the fix is invisible to the exact adopter segment it unblocks.

The DRY extraction (python-architect) is valid but lower urgency -- the duplication is two-function, same-file-family, and a future PR can consolidate without risk.

**Aligned with:** Portability by manifest -- namespace subdirectories under `.apm/prompts/` are a manifest-driven convention that preserves author intent across host integrations; pragmatic as npm -- silent drops violate the install-does-what-you-declared contract, and the fix restores that invariant.

**Growth signal.** Claude Code plugin authors are a high-leverage early-adopter segment. Silent-drop bugs are silent churn -- authors blame their own config, not APM. This fix plus a CHANGELOG entry and docs mention converts an invisible capability into a discoverable one. Track whether this correlates with new Claude-plugin packages post-release.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 2 | Sound fix; `primitive_namespace` is well-placed on BaseIntegrator. One recommended DRY extraction for the duplicated `find_prompt_files` body. |
| CLI Logging Expert | 0 | 1 | 1 | No new CLI output regressions; two recommended improvements for namespace visibility in verbose/diagnostic paths. |
| DevX UX Expert | 0 | 1 | 2 | Namespaced prompts now deploy correctly; install output shows only a count, limiting discoverability for namespace-aware authors. |
| Supply Chain Security | 0 | 0 | 1 | Two-layer path defense is correctly ordered and catches traversal; minor symlink gap on source side is non-exploitable for path escape. |
| OSS Growth Hacker | 0 | 2 | 1 | Silent-drop fix unblocks Claude Code plugin authors -- a high-leverage adopter segment -- but the win is invisible without a CHANGELOG entry and a docs mention. |
| Doc Writer | 0 | 2 | 1 | Missing CHANGELOG entry for user-facing bug fix; prompts authoring doc omits namespaced subdirectory convention entirely. |
| Test Coverage Expert | 0 | 1 | 0 | CommandIntegrator namespace path is well tested; PromptIntegrator's namespace + mkdir path has no integration test through `integrate_package_prompts`. |
| Performance Expert | 0 | 0 | 2 | Negligible perf impact -- mkdir and validate_path_segments are O(n) in prompt count with tiny constants; no algorithmic regression. |

### Top 5 follow-ups

1. **[Test Coverage Expert]** Add integration test for PromptIntegrator with namespaced prompt file -- CommandIntegrator has TestNamespacedCommands; PromptIntegrator has no equivalent. Missing regression trap on a freshly changed devx-promise surface.
2. **[Doc Writer]** Add CHANGELOG [Unreleased] entry and document namespaced prompt layout in authoring docs -- User-facing behavior fix is invisible without CHANGELOG; authors won't try the pattern without docs showing the convention.
3. **[Python Architect]** Extract duplicated `find_prompt_files` body to BaseIntegrator -- Byte-for-byte identical logic in two integrators; future fix must be applied twice.
4. **[CLI Logging Expert]** Surface namespace context in skip/collision warnings and emit verbose mkdir line -- Namespace authors debugging deploy issues currently see flat filenames only.
5. **[Supply Chain Security Expert]** Consider source-side symlink rejection on rglob-discovered files -- No write-escape but a malicious package could read outside its tree via symlink in `.apm/prompts/`.

### Architecture

```mermaid
classDiagram
    direction LR
    class BaseIntegrator {
      <<Abstract>>
      +find_files_by_glob(path, pattern) list~Path~
      +primitive_namespace(source, dir)$ tuple~str~
    }
    class PromptIntegrator {
      +find_prompt_files(package_path) list~Path~
      +integrate_prompts(package_info)
    }
    class CommandIntegrator {
      +find_prompt_files(package_path) list~Path~
      +integrate_commands(package_info)
    }
    class AgentIntegrator {
      +find_agent_files(package_path) list~Path~
    }
    class PathSecurity {
      <<Module>>
      +validate_path_segments(seg, context)
      +ensure_path_within(path, root)
    }
    BaseIntegrator <|-- PromptIntegrator
    BaseIntegrator <|-- CommandIntegrator
    BaseIntegrator <|-- AgentIntegrator
    BaseIntegrator ..> PathSecurity : delegates validation
    class CommandIntegrator:::touched
    class BaseIntegrator:::touched
    note for BaseIntegrator "primitive_namespace: static helper\nvalidates segments via PathSecurity"
```

```mermaid
flowchart TD
    A["apm install / integrate"] --> B{"Integrator type?"}
    B -->|prompts| C["PromptIntegrator.find_prompt_files"]
    B -->|commands| D["CommandIntegrator.find_prompt_files"]
    C --> E["find_files_by_glob(root, *.prompt.md)"]
    C --> F["find_files_by_glob(.apm/prompts, **/*.prompt.md)"]
    D --> E2["find_files_by_glob(root, *.prompt.md)"]
    D --> F2["find_files_by_glob(.apm/prompts, **/*.prompt.md)"]
    F --> G["BaseIntegrator.primitive_namespace(file, .apm/prompts)"]
    F2 --> G
    G --> H{"relative_to succeeds?"}
    H -->|no| I["return ()"]
    H -->|yes| J{"validate_path_segments per part"}
    J -->|PathTraversalError| I
    J -->|valid| K["return namespace tuple"]
    K --> L["target_path = deploy_dir / *namespace / filename"]
    L --> M["[FS] target_path.parent.mkdir(parents=True)"]
    M --> N["[FS] write transformed file"]
```

### Recommendation

The fix is correct, security is non-blocking, and it unblocks a high-value adopter segment. Ship now; open a follow-up issue for the PromptIntegrator integration test and CHANGELOG/docs entry before the next release cut.

---

<details>

#### Python Architect

- **[recommended]** Identical `find_prompt_files` body duplicated across PromptIntegrator and CommandIntegrator at `src/apm_cli/integration/command_integrator.py:189`
  Both integrators now have byte-for-byte identical logic (flat root + recursive .apm/prompts). Per the 'Extract when shared' refactoring principle, this should live once -- either pulled up to BaseIntegrator or as a shared helper.
  *Suggested:* Move `find_prompt_files` to BaseIntegrator as a concrete method.
- **[nit]** `primitive_namespace` silently returns `()` on PathTraversalError -- consider a diagnostic at `src/apm_cli/integration/base_integrator.py:1005`
  A traversal segment is likely an authoring mistake. Returning `()` silently drops the file with no trace in the diagnostic collector.
  *Suggested:* Add `_rich_warning` before returning `()`.
- **[nit]** `@staticmethod` is appropriate but consider a module-level free function in `utils/`
  Pure functions with no instance coupling are easier to test in isolation. Staying on BaseIntegrator is also defensible.

#### CLI Logging Expert

- **[recommended]** Skip/collision warnings lose namespace context at `src/apm_cli/integration/prompt_integrator.py`
  When a namespaced prompt is skipped, the diagnostic message shows flat filename or full path. Surfacing the namespace tuple would help authors debug.
  *Suggested:* Prefix skip messages with namespace path when non-empty.
- **[nit]** No verbose-mode signal when a namespaced sub-directory is first created at the target
  `mkdir(parents=True)` silently creates new directories. Verbose mode users would benefit from seeing `[i] Created commands/opsx/`.
  *Suggested:* Gate behind verbose; emit info line via CommandLogger.

#### DevX UX Expert

- **[recommended]** No per-file output tells authors which namespace a prompt landed in
  `apm install` reports only a total count. A namespace user can't confirm their layout landed correctly without inspecting the target dir manually.
  *Suggested:* Add a `--verbose` line per deployed prompt showing the relative target path.
- **[nit]** Rename scenario leaves orphan namespace file until next full install
  Consistent with other primitives but worth documenting for namespace-aware authors.
- **[nit]** No test asserts that skipped namespaced prompt leaves no empty dir
  `mkdir` only called after skip checks, but no test pins this invariant.

- **[nit]** Source-side symlink traversal in rglob is not guarded at `src/apm_cli/integration/prompt_integrator.py`
  A malicious package could symlink `.apm/prompts/evil -> /etc/secrets` causing rglob to read files outside the package. Target is still contained by `ensure_path_within` so no write-escape occurs.
  *Suggested:* Consider `source.is_symlink()` rejection on discovered files.

#### OSS Growth Hacker

- **[recommended]** Add a CHANGELOG entry framing namespaced prompts as now supported
  Claude Code plugin authors are the early-adopter segment APM needs. Without a CHANGELOG entry, this capability stays invisible.
- **[recommended]** Mention namespaced prompt support in the package authoring docs
  Authors won't try the pattern if docs don't show it.
- **[nit]** Use a real-world namespace pattern in test fixture (e.g. `commands/ops/deploy.prompt.md`) to signal intent to contributors

#### Auth Expert -- inactive

PR changes local filesystem prompt discovery and deployment only; no auth, token, or host-classification surface is touched.

#### Doc Writer

- **[recommended]** No CHANGELOG entry in [Unreleased] for namespaced prompt discovery fix
  User-facing behavioral fix belongs in CHANGELOG. [Unreleased] block has no entry.
- **[recommended]** `docs/producer/author-primitives/prompts.md` does not document namespaced subdirectory layout
  Prompts authoring page shows only flat layout. No mention that subdirectories under `.apm/prompts/` are valid or how the namespace is preserved.
- **[nit]** Pitfalls section has a stale collision warning that doesn't account for namespaced paths
  Two prompts can share a base filename if in different subdirectories. Current warning is incomplete.

#### Test Coverage Expert

- **[recommended]** PromptIntegrator.integrate_package_prompts never tested with a namespaced (nested) prompt file
  The PR adds namespace preservation in `prompt_integrator.py`. The analogous path in CommandIntegrator IS tested (TestNamespacedCommands), but PromptIntegrator has no test exercising `integrate_package_prompts` with a nested prompt.
  *Suggested:* Add `test_integrate_nested_prompt_preserves_namespace`.
  *Proof (missing):* `tests/unit/integration/test_prompt_integrator.py::test_integrate_nested_prompt_preserves_namespace` -- proves: Namespaced prompts are deployed to the correct subdirectory by PromptIntegrator [devx]
  `assert (github_prompts / 'opsx' / 'deploy.prompt.md').exists()`

#### Performance Expert

- **[nit]** `mkdir(parents=True)` called per-file instead of per-unique-namespace
  n syscalls instead of k (where k = unique namespaces). Sub-millisecond even for 100 prompts.
  *Suggested:* Optionally memoize created dirs with a `set()`.
- **[nit]** `validate_path_segments` called per component per file -- O(n*d) but d<=3 in practice
  No action needed.

</details>

danielmeppiel:
Thank you for the work you have put into APM. We are closing this PR as part of a maintainer-led issue-first intake reset because we could not identify an APM issue tracking this change.

Our small maintainer team cannot sustainably review implementations before agreeing which problems and scope we can support. We recognize this process was not clear when some contributions began, and we are sorry for that. This is a review-capacity decision, not a judgment of your work's quality.

Please [open an issue](https://github.com/microsoft/apm/issues/new/choose) describing the problem and proposed scope, or point us to the existing matching issue, and link it here. For substantive changes, a responsible human maintainer must agree the scope, acceptance criteria and review contact under the [contribution rules](https://github.com/microsoft/apm/blob/main/CONTRIBUTING.md#before-you-start-implementation).

Once the tracking and applicable scope decision are in place, please ask here to reopen this PR so the existing work can be reused. There is no need to discard your contribution or start over. The same issue-first expectations apply to maintainer-authored and automated PRs.

PR #2531, opened 2026-08-07
fix(cache): keep git stderr in bare clone error messages

### LINKED ISSUE #2529: [BUG] Clone failures on github.com are reported as "network error, not an auth failure" — git stderr is dropped when wrapping CalledProcessError
**Describe the bug**

`GitCache._create_bare` wraps a failed `git clone` in a `RuntimeError` using f-string interpolation of the exception:

```python
# src/apm_cli/cache/git_cache.py:436
raise RuntimeError(f"Failed to clone {_sanitize_url(url)}: {exc}") from exc
```

`CalledProcessError.__str__` renders only `Command '[...]' returned non-zero exit status 128` — it does not include `stderr`. The underlying git diagnostic is therefore lost at this boundary.

That matters because `AuthResolver.is_public_github_auth_failure()` classifies failures by **string-matching the exception text** for markers like `authentication failed` and `could not read username`. Once `stderr` is gone, the match fails, `public_github_non_auth_failure` is set, and `build_clone_failure_message` emits:

```
Could not connect to github.com (network error, not an auth failure).
Check your internet connection and proxy settings.
```

So a plain authentication failure is reported to the user as a network/proxy problem. The real cause is only visible with `APM_LOG_LEVEL=DEBUG`.

The same pattern is at line 426 (`partial fallback also failed`), so both wrapping sites are affected.

**To Reproduce**

The classifier inversion can be shown directly, no private repo or credentials needed:

```python
import subprocess
from apm_cli.core.auth import AuthResolver

try:
    subprocess.run(
        ['git', 'clone', '--bare',
         'https://github.com/octocat/private-placeholder-xyz.git', '/tmp/repro_bare'],
        check=True, capture_output=True, text=True,
        env={'PATH': '/usr/bin:/bin', 'GIT_TERMINAL_PROMPT': '0',
             'GIT_CONFIG_NOSYSTEM': '1', 'GIT_CONFIG_GLOBAL': '/dev/null'})
except subprocess.CalledProcessError as exc:
    print(AuthResolver.is_public_github_auth_failure(exc))                        # True
    wrapped = RuntimeError(f'Failed to clone URL: {exc}')  # git_cache.py:436
    print(AuthResolver.is_public_github_auth_failure(wrapped))                    # False
```

Output:

```
git stderr: fatal: could not read Username for 'https://github.com': terminal prompts disabled
classifier on original exception : True
classifier on wrapped RuntimeError: False
```

**Expected behavior**

`stderr` should survive the wrap, so `is_public_github_auth_failure()` still classifies correctly and the user is told authentication failed rather than being sent to debug their network and proxy settings.

 - OS: macOS 26.5.2
 - Python Version: 3.12.7
 - APM Version: 0.28.0 (also present on `main` @ `3aa0365`)
 - git: 2.49.0

**Logs**

With `APM_LOG_LEVEL=DEBUG`, the real diagnostic is present:

```
remote: Invalid username or token. [REDACTED] is not supported for Git operations.
fatal: Authentication failed for 'https://github.com/<org>/<repo>.git/'
```

Without it, the user only sees the "network error, not an auth failure" message quoted above.

**Additional context**

Found while installing a subdirectory package from a **private** repo over HTTPS. The message sent me looking at network and proxy configuration for a while; `git clone` by hand appeared to succeed, which reinforced the wrong hypothesis — my manual clone was being served by the `osxkeychain` credential helper, which APM deliberately disables (`credential.helper=""`, `GIT_CONFIG_NOSYSTEM=1`). Reproducing APM's isolated env made the real auth failure visible.

I'm happy to open a PR: pass `exc.stderr` into the wrapped message at both sites, plus a unit test asserting `is_public_github_auth_failure()` stays `True` across the wrap. `cache/` is outside `critical_paths.txt`, so no spec ritual should be needed.

That investigation also surfaced what looks like a **separate** bug: for a private repo, the root package installs fine but a subdirectory package fails, because the shared bare-cache clone runs without credentials. I'll file that separately once I've located the exact line — mentioning it here only so the two aren't conflated.

Naofel-eal:
## Description

Fixes #2529.

`GitCache._ensure_bare_repo` wraps a failed `git clone --bare` in a `RuntimeError` built by interpolating the caught exception:

```python
raise RuntimeError(f"Failed to clone {_sanitize_url(url)}: {exc}") from exc
```

`CalledProcessError.__str__` renders only `Command '[...]' returned non-zero exit status 128` — git's stderr lives on `exc.stderr` and is dropped.

That loss is not cosmetic, because `AuthResolver.is_public_github_auth_failure` classifies a failure by string-matching the message text. With stderr gone, the match fails, the failure is filed as `public_github_non_auth_failure`, and the user is told:

```
Could not connect to github.com (network error, not an auth failure).
Check your internet connection and proxy settings.
```

while git actually said `fatal: could not read Username for 'https://github.com': terminal prompts disabled`. The real cause never reaches the terminal, not even under `--verbose`, so the user debugs connectivity and proxy settings for what is a credential problem.

The fix folds git's stderr into the message at both raise sites, through a `_clone_failure_detail` helper. Two details worth calling out:

- **Redaction.** git can echo a credential back in stderr (`fatal: Authentication failed for 'https://<token>@github.com/...'`). The existing `_sanitize_url` only redacts a URL's `password` component, so the token-as-username form would survive it. `_sanitize_git_stderr` redacts that form and bare platform tokens (`ghp_`/`gho_`/`ghu_`/`ghs_`/`ghr_`/`glpat`) before anything reaches a message. Two tests assert the token never survives.
- **Bounding.** The appended text is capped at 500 characters so a verbose remote cannot flood the terminal.

Behaviour when there is no stderr to add (`TimeoutExpired`, `OSError`, empty stderr) is unchanged — the message falls back to `str(exc)` exactly as today.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Testing

New file `tests/unit/cache/test_git_cache_clone_error_detail.py`, 16 tests:

- **`TestCloneFailureDetail`** — stderr is included; `bytes` stderr is decoded; falls back to `str(exc)` when stderr is absent, empty, or the exception has no `stderr` attribute; a flooding remote is truncated; a token echoed in stderr is redacted.
- **`TestBareCloneFailurePropagatesDetail`** — drives `_ensure_bare_repo` with a patched `subprocess.run`, for both the full-clone and the partial-fallback raise site, and asserts the resulting `RuntimeError` is still classified by `is_public_github_auth_failure`. These two are the regression tests: reverting either raise site to plain `{exc}` makes them fail.
- **`TestAuthFailureStaysClassifiable`** — pins both directions of the contract, including that plain interpolation loses the classification (the reason the helper exists) and that a genuine network failure is *not* reclassified as auth.
- **`TestSanitizeGitStderr`** — token-as-username, basic-auth pair, bare platform token, and a credential-free message left untouched.

Verified locally on Python 3.12, mirroring CI:

```
uv run pytest tests/unit tests/test_console.py -x   -> 19819 passed, 2 skipped, 21 xfailed
uv run --extra dev ruff check src/ tests/           -> All checks passed!
uv run --extra dev ruff format --check src/ tests/  -> 1611 files already formatted
```

## Spec conformance

- [x] N/A — this PR does not change OpenAPM-observable behaviour

Only the text of an error message changes; no CLI surface, manifest, lockfile or resolution behaviour is affected. `src/apm_cli/cache/` is not listed in `tests/spec_conformance/critical_paths.txt`, so the Mode B detector does not apply.

## Note

Issue #2529 mentions a second, separate defect found while investigating this one (a private-repo subdirectory package failing because the shared bare-cache clone runs without credentials). It is **not** addressed here — this PR is scoped to the message-fidelity bug alone. The second one still needs its root cause pinned to a line, and I'd rather send it separately.

### DIFF
--- a/src/apm_cli/cache/git_cache.py
+++ b/src/apm_cli/cache/git_cache.py
@@ -45,6 +45,10 @@ _log = logging.getLogger(__name__)
 # Full SHA pattern: 40 hex characters
 _SHA_RE = re.compile(r"^[0-9a-f]{40}$", re.IGNORECASE)
 
+# Upper bound on git stderr folded into a clone error message, so a
+# verbose remote cannot flood the terminal.
+_GIT_STDERR_MAX_CHARS = 500
+
 
 def _safe_git_args() -> list[str]:
     """Return hardening ``-c`` args prepended to every git subprocess.
@@ -426,14 +430,16 @@ class GitCache:
                         robust_rmtree(staged, ignore_errors=True)
                         raise RuntimeError(
                             f"Failed to clone {_sanitize_url(url)} "
-                            f"(partial fallback also failed): {exc2}"
+                            f"(partial fallback also failed): {_clone_failure_detail(exc2)}"
                         ) from exc2
                 if not fallback_done:
                     # Clean up staged on failure
                     from ..utils.file_ops import robust_rmtree
 
                     robust_rmtree(staged, ignore_errors=True)
-                    raise RuntimeError(f"Failed to clone {_sanitize_url(url)}: {exc}") from exc
+                    raise RuntimeError(
+                        f"Failed to clone {_sanitize_url(url)}: {_clone_failure_detail(exc)}"
+                    ) from exc
 
             # Atomic land (lock is already held; pass it through so the
             # rename completes under the same critical section).
@@ -815,6 +821,43 @@ def _dir_size(path: Path) -> int:
     except OSError:
         pass
     return total
+
+
+def _sanitize_git_stderr(stderr: str) -> str:
+    """Redact credentials from git stderr before it reaches an error message.
+
+    ``_sanitize_url`` only redacts a URL's password component, so the
+    token-as-username form (``https://<token>@host``) would survive it.
+    Both that form and bare platform tokens are redacted here.
+    """
+    redacted = re.sub(r"(https?://)[^@/\s]+@", r"\1***@", stderr)
+    redacted = re.sub(
+        r"(ghp_|gho_|ghu_|ghs_|ghr_|glpat[_-])[A-Za-z0-9_\-]+",
+        "***",
+        redacted,
+    )
+    return redacted.strip()
+
+
+def _clone_failure_detail(exc: BaseException) -> str:
+    """Render a clone failure with git's own diagnostic included.
+
+    ``CalledProcessError.__str__`` reports only the exit status, so
+    interpolating the exception alone drops git's stderr. Callers that
+    classify failures by message text (``AuthResolver.
+    is_public_github_auth_failure``) then read an auth failure as a
+    network failure, sending users to debug connectivity for what is a
+    credential problem.
+    """
+    stderr = getattr(exc, "stderr", None)
+    if isinstance(stderr, bytes):
+        stderr = stderr.decode("utf-8", errors="replace")
+    if not stderr:
+        return str(exc)
+    detail = _sanitize_git_stderr(stderr)
+    if len(detail) > _GIT_STDERR_MAX_CHARS:
+        detail = f"{detail[:_GIT_STDERR_MAX_CHARS]}..."
+    return f"{exc}: {detail}" if detail else str(exc)
 
 
 def _sanitize_url(url: str) -> str:
--- a/tests/unit/cache/test_git_cache_clone_error_detail.py
+++ b/tests/unit/cache/test_git_cache_clone_error_detail.py
@@ -1,0 +1,168 @@
+"""Clone failures must carry git's own diagnostic, redacted.
+
+``CalledProcessError.__str__`` renders only the exit status, so wrapping
+it with plain interpolation drops git's stderr. That loss is not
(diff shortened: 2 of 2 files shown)


Naofel-eal:

Naofel-eal:
Withdrawing this PR. On closer inspection the justification I gave overstated the impact: on the `apm install` path both callers of `GitCache.get_checkout` swallow the exception (`github_downloader.py` lines 1346 and 1852, both falling back to a network clone), so the `RuntimeError` this patch improves never reaches `is_public_github_auth_failure` there. The misleading "network error, not an auth failure" message I quoted originates elsewhere, in `clone_engine.py` -> `build_clone_failure_message`.

Rather than leave a PR up whose stated rationale does not hold, I would rather close it and come back once I have pinned the actual root cause. Apologies for the noise.

PR #2534, opened 2026-08-08
fix(auth): normalize git message locale so auth failures stay classifiable

### LINKED ISSUE #2533: [BUG] Localised git stderr breaks auth-failure classification, suppressing the token retry
**Describe the bug**

`AuthResolver.is_public_github_auth_failure` classifies a failed clone by matching English signal strings against git's stderr (`"authentication failed"`, `"could not read username"`, `"repository not found"`, ...). Git localises those diagnostics through gettext, and APM does not pin a locale for the git subprocesses it spawns itself.

On a machine whose locale is not English, git writes its authentication failure in the user's language. The classifier does not recognise it, and the consequence is not cosmetic: the verdict gates the **token retry** in `try_with_fallback` (`src/apm_cli/core/auth.py:797-802`).

```python
if (
    lazy_public_github
    and path is not None
    and not self.is_public_github_auth_failure(exc)
):
    raise
ctx = _auth_context()
if ctx.token:
    _log(f"Unauthenticated failed, retrying with token (source: {ctx.source})")
```

Since `uses_public_github_anonymous_first` makes every github.com access start anonymously, a misclassified failure means the token is **never tried**. Installing a private-repo dependency fails, and the user is pointed at the wrong cause:

```
Could not connect to github.com (network error, not an auth failure).
Check your internet connection and proxy settings.
```

while git actually said "authentication failed".

Two things make this hard to spot:

- Clones issued through GitPython are unaffected, because GitPython forces `LANGUAGE=C`/`LC_ALL=C` on every command it runs (`git/cmd.py`). Only APM's own `subprocess` calls inherit the ambient locale — and that is the path the shared bare cache uses.
- Git ships message catalogues for 19 languages today. A user whose language has no catalogue (e.g. `ja`) sees English and is unaffected — until that translation is added.

**To Reproduce**

1. Set a locale for which git ships a catalogue: `export LANG=fr_FR.UTF-8` (`de_DE.UTF-8` behaves identically).
2. Have a token available through `gh auth login`, but not exported in the environment, so it has to come from the retry path.
3. Use a cold cache: `export APM_CACHE_DIR=$(mktemp -d)`.
4. `apm install <org>/<private-repo>/<subdir>` — a subdirectory package, which is the case that routes through the shared bare-cache clone.

Same command, same token, same repository; only the locale changes:

| `LC_ALL` | Result |
|---|---|
| `fr_FR.UTF-8` | fails — "network error, not an auth failure" |
| `de_DE.UTF-8` | fails — same |
| `C` | installs successfully |

`LC_ALL=C apm install ...` alone makes the install succeed with no code change, which isolates the locale as the cause.

The exception captured on the failing path (values redacted):

```
CalledProcessError: Command '[... 'clone', '--bare', 'https://github.com/<org>/<repo>.git', ...]'
  returned non-zero exit status 128.
  stderr: "remote: Invalid username or token. Password authentication is not supported
           for Git operations.
           fatal : Échec d'authentification pour 'https://github.com/<org>/<repo>.git/'"
```

`is_public_github_auth_failure` returns `False` for it, so `try_with_fallback` re-raises without trying the token. Instrumenting the same run under `LC_ALL=C` yields the untranslated `fatal: Authentication failed`, a `True` verdict, and a successful token retry.

**Expected behavior**

Classification of an authentication failure should not depend on the user's locale. A private-repo dependency should install regardless of the language configured on the machine, and a genuine auth failure should be reported as such instead of as a network/proxy problem.

 - OS: macOS (`darwin`); reproducible on any non-English locale
 - Python Version: 3.12
 - APM Version: 0.28.0 (also on `main` @ `3aa0365`)
 - git: 2.51.0, shipping catalogues for 19 languages under `share/locale/`

**Additional context**

How this surfaced: a shared agent rule hosted in a subdirectory of a private repository. The repository root installed fine, the subdirectory package did not — the subdirectory is what routes through the shared bare cache, whose `subprocess` calls carry the ambient locale. The subdirectory is the trigger, not the cause. The error message sent me looking at proxy and connectivity settings for what was a credential-resolution problem.

I have a fix and will open a PR referencing this issue: normalise the message locale in `AuthResolver._build_git_env`, the single builder every git env in the auth path routes through. Widening the classifier's vocabulary with translated strings looks like the wrong direction — the catalogues change between git versions and cover a moving set of languages.

Naofel-eal:
## Description

Fixes #2533.

`AuthResolver.is_public_github_auth_failure` classifies a failed clone by matching **English** signal strings against git's stderr (`"authentication failed"`, `"could not read username"`, ...). Git localises those diagnostics through gettext, and APM does not pin a locale for the git subprocesses it spawns itself. On a machine whose locale is not English, an authentication failure becomes unrecognisable.

That is not a message-fidelity problem. The verdict gates the **token retry** in `try_with_fallback` (`core/auth.py:797-802`):

```python
if (
    lazy_public_github
    and path is not None
    and not self.is_public_github_auth_failure(exc)
):
    raise
```

Since `uses_public_github_anonymous_first` makes every github.com access start anonymously, a misclassified failure re-raises before the token is ever tried. A private-repo dependency fails to install, and the user is sent to the wrong place:

```
Could not connect to github.com (network error, not an auth failure).
Check your internet connection and proxy settings.
```

while git actually said "authentication failed".

## Approach

Set `LC_ALL=C` and `LANGUAGE=C` in `AuthResolver._build_git_env` — the single builder that `build_public_github_anonymous_git_env`, `build_noninteractive_git_env` and `git_env_for_context` all route through, so it covers both the anonymous attempt and the authenticated retry. The `C` locale has no message catalogue, so git emits its untranslated source strings.

Both variables are required: `LANGUAGE` takes precedence over `LC_ALL` for translations, so pinning `LC_ALL` alone would still leave the bug for anyone who exports `LANGUAGE`. Verified directly — `LC_ALL=fr_FR.UTF-8 LANGUAGE=C git ...` prints English, while `LC_ALL=C LANGUAGE=fr git ...` also prints English.

GitPython already forces both on every command it runs (`git/cmd.py`), which is exactly why clones issued through it were unaffected while APM's own `subprocess` calls were not. This change makes the two paths consistent rather than introducing a new convention.

Two alternatives I discarded:

- **Widening the classifier's vocabulary with translated strings.** Git ships catalogues for 19 languages today and the strings move between versions. It also fails in a way that is easy to miss: a user whose language has no catalogue (e.g. `ja`) is not affected today, but becomes affected the day that translation lands. Normalising at the source is the stable fix.
- **Patching `git_subprocess_env()`** in `utils/git_env.py`. I tried this first; it does **not** fix the failing path — the shared bare-cache clone does not route through that function. Verified experimentally before settling on `_build_git_env`.

## Scope

The remaining git subprocesses that still inherit the ambient locale are the purely local ones in `deps/bare_cache.py` (`clone --local` from the bare, `config`, `checkout`, sparse-cone). They perform no authentication and their stderr never reaches the classifier, so they are intentionally left alone to keep the diff minimal.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Maintenance / refactor

## Testing

- [x] Tested locally
- [x] All existing tests pass

Added to `tests/unit/core/test_public_github_anonymous_first.py`, the existing home of this contract:

- **`test_git_message_locale_is_normalized_over_inherited_locale`** — parametrised over `LC_ALL`/`LANGUAGE`, `LANG`, and `LC_MESSAGES`, so an inherited locale never reaches git whichever variable carries it.
- **`test_translated_clone_failure_still_retries_with_a_token`** — the regression test. It drives the real transport plan with a `clone_action` whose stderr is rendered through `_localized_git_stderr`, a helper that reproduces gettext's own precedence (`LANGUAGE` first, then `LC_ALL`; `C` yields the untranslated string). It therefore returns the translated form *only if* the locale was left inherited, and asserts the anonymous attempt is followed by an authenticated one. The translated fixture is git's own `id` catalogue entry (`fatal: Autentikasi gagal untuk '<url>'`) — a real gettext output that matches none of the classifier's English signals while staying within printable ASCII, as `.apm/instructions/encoding.instructions.md` requires.
- **`_assert_anonymous_attempt_env`** — the shared assertion for the anonymous call shape now pins the locale alongside `GIT_TERMINAL_PROMPT` and `GIT_ASKPASS`.

Reverting the one-line call site in `_build_git_env` makes 10 tests in that file fail, including both new ones — the tests fail for the intended reason rather than passing vacuously.

End-to-end validation on the real case (private-repo subdirectory package, cold cache, HTTPS forced): fails under `LANG=fr_FR.UTF-8` on `main`, installs successfully with this change. That reproduction used a real French locale on the machine; only the in-source test fixture is ASCII. The A/B across locales is in the issue.

Full CI mirror on Python 3.12, all seven lint gates from `.apm/instructions/linting.instructions.md`:

```
uv run pytest tests/unit tests/test_console.py -x     -> 19807 passed, 2 skipped, 21 xfailed
uv run --extra dev ruff check src/ tests/             -> All checks passed!
uv run --extra dev ruff format --check src/ tests/    -> 1610 files already formatted
pylint --enable=R0801 --min-similarity-lines=10       -> rated at 10.00/10
bash scripts/lint-auth-signals.sh                     -> auth-signal lint clean
YAML I/O guard / file length (<=2450) / relative_to   -> clean
```

### Scenario Evidence

|---|-------------------------|--------------|--------------------|------|
| 1 | `apm install` of a private github.com dependency succeeds on a non-English machine by retrying once with the resolved credential | Secure by default, DevX | `tests/unit/core/test_public_github_anonymous_first.py::test_translated_clone_failure_still_retries_with_a_token`<br/>`tests/integration/test_public_github_anonymous_lifecycle_e2e.py::test_private_github_fallback_normalizes_locale_and_completes_lifecycle` (regression-trap for #2533) | e2e |

No CLI surface, manifest, lockfile or resolution semantics change. The change is confined to the environment handed to git subprocesses; its observable effect is that an authentication failure is now classified as one regardless of the machine's locale.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - YAML expansion guard no longer rejects large anchor-free lockfiles (150K+
   entries) with a false-positive "billion-laughs" error. APM-generated
   lockfiles with no anchors or aliases now load without error. (#2389)
+- `apm install` no longer skips the credential retry on non-English machines.
+  Git localises its diagnostics through gettext, so a translated stderr made an
+  authentication failure unrecognisable and private-repo installs failed with
+  misleading network guidance. Every git subprocess APM spawns now runs with
+  `LC_ALL=C` and `LANGUAGE=C`. (by @Naofel-eal, closes #2533)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `apm install` no longer skips the credential retry on non-English machines.
   Git localises its diagnostics through gettext, so a translated stderr made an
   authentication failure unrecognisable and private-repo installs failed with
-  misleading network guidance. Every git subprocess APM spawns now runs with
-  `LC_ALL=C` and `LANGUAGE=C`. (by @Naofel-eal, closes #2533)
+  misleading network guidance. Git subprocesses in the authentication retry
+  path now run with `LC_ALL=C` and `LANGUAGE=C`. (by @Naofel-eal, closes #2533)
 
 ### Changed
 
--- a/src/apm_cli/core/auth.py
+++ b/src/apm_cli/core/auth.py
@@ -139,6 +139,16 @@ _GIT_CHILD_TOKEN_ENV_NAMES = frozenset(
 )
 _GIT_CHILD_TOKEN_ENV_PREFIXES = ("GITHUB_APM_PAT_",)
 
+# Git localises its diagnostics through gettext, but APM classifies clone
+# failures by matching English signal strings (see
+# ``AuthResolver.is_public_github_auth_failure``). A translated stderr makes an
+# authentication failure unrecognisable, which suppresses the token retry. The
+# ``C`` locale has no message catalogue, so git emits its untranslated source
+# strings. ``LANGUAGE`` takes precedence over ``LC_ALL`` for translations and
+# must be neutralised too. GitPython already does this for every command it
+# runs; this keeps APM's own subprocess calls consistent with it.
+_GIT_MESSAGE_LOCALE_ENV: dict[str, str] = {"LC_ALL": "C", "LANGUAGE": "C"}
+
 
 # ---------------------------------------------------------------------------
 # Data classes
@@ -1208,6 +1218,7 @@ class AuthResolver:
         AuthResolver._clear_git_auth_env(env)
         env["GIT_TERMINAL_PROMPT"] = "0"
         env["GIT_ASKPASS"] = "echo"
+        env.update(_GIT_MESSAGE_LOCALE_ENV)
         if token and host_kind == "ado" and scheme in {"basic", "bearer"}:
             # ADO credentials use an Authorization header, never argv or
             # GIT_TOKEN. This keeps PATs and bearer JWTs out of process lists.
--- a/src/apm_cli/core/auth.py
+++ b/src/apm_cli/core/auth.py
@@ -35,8 +35,9 @@ import re
 import sys
 import threading
 import traceback
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from dataclasses import dataclass, field
+from types import MappingProxyType
 from typing import TYPE_CHECKING, NamedTuple, TypeVar
 
 from apm_cli.core.host_providers import (
@@ -147,7 +148,7 @@ _GIT_CHILD_TOKEN_ENV_PREFIXES = ("GITHUB_APM_PAT_",)
 # strings. ``LANGUAGE`` takes precedence over ``LC_ALL`` for translations and
 # must be neutralised too. GitPython already does this for every command it
 # runs; this keeps APM's own subprocess calls consistent with it.
-_GIT_MESSAGE_LOCALE_ENV: dict[str, str] = {"LC_ALL": "C", "LANGUAGE": "C"}
+_GIT_MESSAGE_LOCALE_ENV: Mapping[str, str] = MappingProxyType({"LC_ALL": "C", "LANGUAGE": "C"})
 
 
 # ---------------------------------------------------------------------------
--- a/tests/integration/test_public_github_anonymous_lifecycle_e2e.py
+++ b/tests/integration/test_public_github_anonymous_lifecycle_e2e.py
@@ -246,14 +246,18 @@ def test_all_public_graph_lifecycle_never_resolves_or_leaks_credentials(
         assert all(not observation.authorization_present for observation in repository_requests)
 
 
-def test_private_github_fallback_resolves_once_and_completes_lifecycle(
+def test_private_github_fallback_normalizes_locale_and_completes_lifecycle(
     tmp_path: Path,
     apm_binary_path: Path,
 ) -> None:
-    """One private dependency uses one path-scoped credential fallback."""
+    """A localized private clone retries once with a path-scoped credential."""
     isolated = IsolatedApmEnvironment.create(
         tmp_path / "private-scenario",
-        base_env=dict(os.environ),
+        base_env={
+            **os.environ,
+            "LC_ALL": "id_ID.UTF-8",
+            "LANGUAGE": "id_ID:id",
+        },
     )
     base_environment = isolated.subprocess_env()
     packages = LocalPackageFactory(isolated.package_root)
@@ -323,6 +327,8 @@ def test_private_github_fallback_resolves_once_and_completes_lifecycle(
 
         remote_events = _remote_events(shim)
         assert remote_events
+        assert all(event["lc_all"] == "C" for event in remote_events)
+        assert all(event["language"] == "C" for event in remote_events)
         remote_attempts = [remote for event in remote_events for remote in event["remotes"]]
         assert remote_attempts[0]["authenticated_url"] is False
         assert any(remote["authenticated_url"] is True for remote in remote_attempts)
--- a/tests/unit/core/test_public_github_anonymous_first.py
+++ b/tests/unit/core/test_public_github_anonymous_first.py
@@ -37,6 +37,28 @@ class _HttpStatusError(RuntimeError):
         super().__init__(f"HTTP {status_code}")
 
 
+_UNTRANSLATED_AUTH_STDERR = (
+    "fatal: Authentication failed for 'https://github.com/acme/private.git/'"
+)
+_TRANSLATED_AUTH_STDERR = (
+    "fatal\u00a0: Échec d'authentification pour 'https://github.com/acme/private.git/'"
+)
+
+
+def _localized_git_stderr(env: dict[str, str]) -> str:
+    """Return git's auth failure the way gettext would render it under *env*.
+
+    Git resolves its message catalogue from ``LANGUAGE`` first, then
+    ``LC_ALL``; the ``C`` locale has no catalogue, so the untranslated
+    source string surfaces. Reproducing that precedence here is what makes
+    the retry contract fail when the locale is left inherited.
+    """
+    catalogue = env.get("LANGUAGE") or env.get("LC_ALL") or ""
+    if catalogue.startswith("C"):
+        return _UNTRANSLATED_AUTH_STDERR
+    return _TRANSLATED_AUTH_STDERR
+
+
 def _indexed_git_config(env: dict[str, str]) -> list[tuple[str, str]]:
     """Return process-scoped Git config entries in declared order."""
     count = int(env.get("GIT_CONFIG_COUNT", "0"))
@@ -85,6 +107,8 @@ def _assert_anonymous_attempt_env(env: dict[str, str]) -> None:
     assert env["GIT_TERMINAL_PROMPT"] == "0"
     assert env["GIT_ASKPASS"] == "echo"
     assert env["GIT_CONFIG_NOSYSTEM"] == "1"
+    assert env["LC_ALL"] == "C"
+    assert env["LANGUAGE"] == "C"
     if os.name == "nt":
         assert Path(env["GIT_CONFIG_GLOBAL"]).is_file()
     else:
@@ -455,6 +479,60 @@ def test_public_github_auth_failure_classifier_signal_vocabulary(
 ) -> None:
     """Every documented auth and non-auth signal stays in its intended bucket."""
     assert AuthResolver.is_public_github_auth_failure(failure) is expected
+
+
+@pytest.mark.parametrize(
+    "locale_environment",
+    (
+        {"LC_ALL": "fr_FR.UTF-8", "LANGUAGE": "fr_FR:fr"},
+        {"LANG": "de_DE.UTF-8"},
+        {"LC_MESSAGES": "ja_JP.UTF-8"},
+    ),
+)
+def test_git_message_locale_is_normalized_over_inherited_locale(
+    locale_environment: dict[str, str],
+) -> None:
+    """An inherited locale never reaches git, whichever variable carries it."""
+    with patch.dict(os.environ, locale_environment, clear=True):
+        env = AuthResolver._build_git_env()
+
+    assert env["LC_ALL"] == "C"
+    assert env["LANGUAGE"] == "C"
+
+
+def test_translated_clone_failure_still_retries_with_a_token() -> None:
+    """A localised git stderr stays classifiable, so the token retry happens."""
+    manager = MagicMock(spec=GitHubTokenManager)
+    manager.setup_environment.side_effect = lambda: dict(os.environ)
+    manager.get_token_for_purpose.return_value = None
+    manager.resolve_credential_from_gh_cli.return_value = None
+    manager.resolve_credential_from_git.return_value = "private-token"
+    resolver = AuthResolver(token_manager=manager)
+    dep_ref = DependencyReference.parse("https://github.com/acme/private.git#main")
+    calls: list[tuple[str, dict[str, str]]] = []
+
+    def clone_action(url: str, env: dict[str, str], _target: Path) -> None:
+        calls.append((url, env))
+        if urlparse(url).username is not None:
+            return
+        raise subprocess.CalledProcessError(
+            128,
+            ("git", "clone"),
+            stderr=_localized_git_stderr(env),
+        )
+
+    with patch.dict(os.environ, {"LC_ALL": "fr_FR.UTF-8", "LANGUAGE": "fr_FR:fr"}, clear=True):
+        downloader = _public_downloader(resolver)
+        downloader._execute_transport_plan(
+            dep_ref.repo_url,
+            Path("unused-target"),
+            dep_ref=dep_ref,
+            clone_action=clone_action,
+        )
+
+    assert len(calls) == 2
+    assert urlparse(calls[0][0]).username is None
+    assert urlparse(calls[1][0]).username is not None
 
 
 def test_clear_git_auth_env_only_removes_real_auth_channels() -> None:
--- a/tests/unit/core/test_public_github_anonymous_first.py
+++ b/tests/unit/core/test_public_github_anonymous_first.py
@@ -40,9 +40,7 @@ class _HttpStatusError(RuntimeError):
 _UNTRANSLATED_AUTH_STDERR = (
     "fatal: Authentication failed for 'https://github.com/acme/private.git/'"
 )
-_TRANSLATED_AUTH_STDERR = (
-    "fatal\u00a0: Échec d'authentification pour 'https://github.com/acme/private.git/'"
-)
+_TRANSLATED_AUTH_STDERR = "fatal: Autentikasi gagal untuk 'https://github.com/acme/private.git/'"
 
 
 def _localized_git_stderr(env: dict[str, str]) -> str:
@@ -51,7 +49,9 @@ def _localized_git_stderr(env: dict[str, str]) -> str:
     Git resolves its message catalogue from ``LANGUAGE`` first, then
     ``LC_ALL``; the ``C`` locale has no catalogue, so the untranslated
     source string surfaces. Reproducing that precedence here is what makes
-    the retry contract fail when the locale is left inherited.
+    the retry contract fail when the locale is left inherited. The
+    translated form is git's own ``id`` catalogue entry, chosen because it
+    is a real translation that stays within printable ASCII.
     """
     catalogue = env.get("LANGUAGE") or env.get("LC_ALL") or ""
     if catalogue.startswith("C"):
@@ -484,7 +484,7 @@ def test_public_github_auth_failure_classifier_signal_vocabulary(
 @pytest.mark.parametrize(
     "locale_environment",
     (
-        {"LC_ALL": "fr_FR.UTF-8", "LANGUAGE": "fr_FR:fr"},
+        {"LC_ALL": "id_ID.UTF-8", "LANGUAGE": "id_ID:id"},
         {"LANG": "de_DE.UTF-8"},
         {"LC_MESSAGES": "ja_JP.UTF-8"},
     ),
@@ -521,7 +521,7 @@ def test_translated_clone_failure_still_retries_with_a_token() -> None:
             stderr=_localized_git_stderr(env),
         )
 
-    with patch.dict(os.environ, {"LC_ALL": "fr_FR.UTF-8", "LANGUAGE": "fr_FR:fr"}, clear=True):
+    with patch.dict(os.environ, {"LC_ALL": "id_ID.UTF-8", "LANGUAGE": "id_ID:id"}, clear=True):
         downloader = _public_downloader(resolver)
         downloader._execute_transport_plan(
             dep_ref.repo_url,
--- a/tests/unit/core/test_public_github_anonymous_first.py
+++ b/tests/unit/core/test_public_github_anonymous_first.py
@@ -5,13 +5,14 @@ from __future__ import annotations
 import os
 import subprocess
 from pathlib import Path
+from types import MappingProxyType
 from unittest.mock import MagicMock, call, patch
 from urllib.parse import urlparse
 
 import pytest
 import requests
 
-from apm_cli.core.auth import AuthResolver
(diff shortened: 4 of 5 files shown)


Naofel-eal:
Good catch on the encoding rule — fixed in bfc144b8. The rest of the comment is not accurate, and the distinction matters for the fixture I chose.

**Fixed.** `.apm/instructions/encoding.instructions.md` does require printable ASCII (U+0020-U+007E) in source, and the French string violated it. The file is now 0 non-ASCII bytes.

**Not the case: no guard fails, and cp1252 was never at risk.** There is no encoding lint gate over `tests/`; `scripts/check_diagnostic_ascii_owner.py` is a narrow architecture-boundary check on `printable_ascii_text` delegation, unrelated to literals. For scale: 174 test files and 53 `src/` files currently hold non-ASCII bytes. And a `str` literal in a UTF-8 source file cannot raise `UnicodeEncodeError` on a cp1252 terminal — only *writing it to a stream* can, which this test never does (it feeds `CalledProcessError.stderr` and asserts on call counts). So the rule is worth following, but not for the reason given.

**Why not an invented ASCII string.** The suggestion — "an ASCII-only string that won't match the classifier's English signals" — would make the test tautological: I would be asserting the classifier fails on a string I made up to make it fail, which proves nothing about git. The fixture's value is that it is a real gettext output.

Git's `id` catalogue happens to satisfy both constraints:

```
$ python3 -c "import gettext; print(gettext.GNUTranslations(open('.../id/LC_MESSAGES/git.mo','rb')).gettext(\"Authentication failed for '%s'\"))"
Autentikasi gagal untuk '%s'
```

`fatal: Autentikasi gagal untuk '<url>'` is a genuine translation git emits under `LANGUAGE=id`, matches none of the eight English signals, and is entirely printable ASCII. The parametrised locales moved to `id_ID` for consistency, and the helper docstring records why that catalogue was picked so nobody "simplifies" it back to a synthetic string later.

Verified after the change: 53 passed in that file, 19807 passed in the full unit suite, ruff check/format clean, R0801 10.00/10, auth-signal lint clean. Reverting the one-line call site in `_build_git_env` still fails 10 tests including both new ones, so the regression test continues to fail for the intended reason.

sergio-sisternes-epam:

> Fixes silent auth-failure misclassification on non-English locales by pinning git subprocess locale to C, restoring credential retry for international users.

cc @Naofel-eal @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

All eight panelists converge: this is a clean, minimal, correctly-scoped fix to a silent reliability regression affecting every non-English-locale user. The auth-expert confirms belt-and-suspenders correctness of setting both LC_ALL and LANGUAGE, backed by a passing unit test that asserts inherited locale vars are always overridden. Supply-chain-security finds no credential leak or scope change. Performance is unmeasurable noise. No panelist raised a blocking finding.

The only substantive gap is a missing CHANGELOG entry, flagged independently by both doc-writer and oss-growth-hacker at recommended severity. This is a legitimate ask -- silent auth misclassification on non-English machines is exactly the class of bug that deserves a visible Fixed line -- but it does not block merge. The contributor can add it in a follow-up commit or the maintainer can fold it at merge time.

This is a high-signal external contribution: first-time contributor, real-world bug affecting an entire locale class, clean TDD approach, CI green at 19807 tests. Ship it.

**Aligned with:** Secure by default: auth-failure classification must never silently degrade; locale pinning ensures the credential-retry path fires reliably regardless of user environment. Pragmatic as npm: private-repo installs now Just Work on non-English machines without requiring users to manually set LC_ALL. OSS community driven: external contributor identified a real-world international reliability gap, delivered a tested fix, and validated the open-source funnel.

**Growth signal.** Story angle: 'APM works correctly worldwide.' Silent auth misclassification on non-English locales quietly kills international adoption. Naming this fix in CHANGELOG and crediting Naofel-eal in release notes signals APM cares about global developers and rewards community contributions.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 1 | Clean, minimal fix. Single-authority env builder correctly extended; constant placement and naming are idiomatic. |
| CLI Logging Expert | 0 | 0 | 0 | Clean infrastructure fix. No new user-facing output paths introduced; locale env is internal plumbing only. |
| DevX UX Expert | 0 | 0 | 0 | Fix correctly restores the 5-minute install promise for non-English locale users with no UX regression. |
| Supply Chain Security Expert | 0 | 0 | 0 | Locale pinning in git subprocess env is safe; no credential leak, no token scope change, no path traversal vector. |
| OSS Growth Hacker | 0 | 1 | 1 | High-value fix for global users; silent auth misclassification on non-English locales quietly kills international adoption. |
| Auth Expert | 0 | 0 | 2 | Locale pinning in _build_git_env is correct and complete for auth-failure classification; no blocking issues found. |
| Doc Writer | 0 | 1 | 1 | CHANGELOG missing a [Unreleased] Fixed entry; no docs pages are stale. |
| Test Coverage Expert | 0 | 0 | 1 | Locale normalization fix is well-defended by two unit tests covering both env-override and end-to-end retry path; no blocking gaps. |

### Top 3 follow-ups

1. **[Doc Writer]** Add CHANGELOG [Unreleased] Fixed entry for locale-based auth misclassification -- Users on non-English locales who hit this bug deserve a visible record; also reinforces international-first positioning.
2. **[OSS Growth Hacker]** Credit Naofel-eal visibly in release notes -- First-time community fix improving reliability for an entire locale class is strong contributor-funnel signal.
3. **[Python Architect]** Consider freezing _GIT_MESSAGE_LOCALE_ENV with MappingProxyType -- Makes immutability intent explicit; low risk but good hygiene for a module-level dict on a security-sensitive path.

### Architecture

```mermaid
classDiagram
    direction LR
    class AuthResolver {
      <<Strategy>>
      +build_public_github_anonymous_git_env() dict
      +build_noninteractive_git_env() dict
      +git_env_for_context() dict
      -_build_git_env(base_env, token, host_kind, scheme) dict
      -_clear_platform_token_env(env)
      -_clear_git_auth_env(env)
      +is_public_github_auth_failure(stderr) bool
    }
    class _GIT_MESSAGE_LOCALE_ENV {
      <<ModuleConstant>>
      +LC_ALL: str
      +LANGUAGE: str
    }
    AuthResolver ..> _GIT_MESSAGE_LOCALE_ENV : applies in _build_git_env
    note for AuthResolver "Single authority for git subprocess env.\nAll public env builders delegate to _build_git_env."
    class _GIT_MESSAGE_LOCALE_ENV:::touched
```

```mermaid
flowchart TD
    A["build_public_github_anonymous_git_env()"] --> D
    B["build_noninteractive_git_env()"] --> D
    C["git_env_for_context()"] --> D
    D["_build_git_env()\nsrc/apm_cli/core/auth.py"]
    D --> E["_clear_platform_token_env(env)"]
    E --> F["_clear_git_auth_env(env)"]
    F --> G["set GIT_TERMINAL_PROMPT=0"]
    G --> H["set GIT_ASKPASS=echo"]
    H --> I["env.update(_GIT_MESSAGE_LOCALE_ENV)\n[NEW] forces C locale"]
    I --> J{"token + ADO?"}
    J -->|Yes| K["inject auth header"]
    J -->|No| L["return env"]
    K --> L
```

### Recommendation

Zero blocking findings across all panelists. Fix is minimal, correctly scoped, well-tested (CI green, 19807 passed), and authored by an external contributor. The missing CHANGELOG entry is the only recommended follow-up worth folding before or at merge, but it does not gate shipment. Merge now; fold CHANGELOG in the merge commit or ask the contributor to amend.

---

<details>

#### Python Architect

- **[nit]** Consider freezing the locale dict at `src/apm_cli/core/auth.py:146`
  Module-level mutable dicts can be accidentally mutated. A MappingProxyType or inline freeze would make the intent explicit, though the risk is low given the underscore prefix.
  *Suggested:* `from types import MappingProxyType` / `_GIT_MESSAGE_LOCALE_ENV = MappingProxyType({"LC_ALL": "C", "LANGUAGE": "C"})`

#### CLI Logging Expert

No findings.

#### DevX UX Expert

No findings.

No findings.

#### OSS Growth Hacker

- **[recommended]** Add CHANGELOG [Unreleased] Fixed entry for international users
  Silent auth misclassification on non-English locales is exactly the kind of bug that quietly kills international adoption. Naming it in CHANGELOG signals APM cares about global devs.
- **[nit]** Credit the external contributor visibly in release notes
  First-time community bug fixes that improve reliability for a whole locale class are a strong signal for the contributor funnel.

#### Auth Expert

- **[nit]** token_manager.py subprocess calls don't set locale env, but this is harmless at `src/apm_cli/core/token_manager.py:230`
  GitHubTokenManager.resolve_credential_from_git and resolve_credential_from_gh_cli call subprocess.run without LC_ALL/LANGUAGE. However these calls only parse stdout for credential values -- they never pattern-match stderr for auth failure classification. Fix is correctly scoped. No action needed.
  *Proof (passed at static):* token_manager subprocess calls don't classify errors by message text
- **[nit]** Comment on gettext precedence is accurate but could be tightened at `src/apm_cli/core/auth.py:148`
  GNU gettext resolves: LANGUAGE > LC_ALL > LC_MESSAGES > LANG. Setting both LC_ALL and LANGUAGE to C is belt-and-suspenders correct. LANG and LC_MESSAGES cannot override both.
  *Proof (passed at unit):* `tests/unit/core/test_public_github_anonymous_first.py::test_git_message_locale_is_normalized_over_inherited_locale` -- proves: LC_ALL and LANGUAGE in env override any inherited LANG/LC_MESSAGES

#### Doc Writer

- **[recommended]** Add CHANGELOG entry for locale-based auth misclassification fix
  The bug silently suppressed the auth-credential retry on any non-English machine, causing private-repo installs to fail with misleading guidance. Users on non-English locales who hit this bug deserve a visible record. Suggested: `apm install no longer silently skips the credential retry on non-English machines. LC_ALL=C and LANGUAGE=C are now set for every git subprocess APM spawns. (closes #2533)`
- **[nit]** Auth docs do not mention locale pinning -- confirm intentional omission
  Implementation detail; no user-facing change needed. Flagged only for deliberate omission confirmation.

#### Test Coverage Expert

- **[nit]** No integration-with-fixtures test for locale normalization against a real git binary
  Unit tests mock the clone action. The auth critical surface floor is integration-with-fixtures per the tier matrix. However, the fix is a single env-dict update whose correctness is fully captured by asserting dict contents. Promoting to integration would require live git + real private repo, impractical for CI without credentials.
  *Proof (passed at unit):* `tests/unit/core/test_public_github_anonymous_first.py::test_git_message_locale_is_normalized_over_inherited_locale` -- proves: Inherited locale vars are always overridden to C in the git subprocess env
  `assert env["LC_ALL"] == "C"; assert env["LANGUAGE"] == "C"`

A 2-key dict.update() on a module-level constant adds ~50ns per call -- unmeasurable noise on a path dominated by subprocess spawn (~50ms). No loops, no I/O, no algorithmic regression.

</details>

danielmeppiel:

> Normalize AuthResolver Git diagnostics so authenticated retry remains reliable across user locales.

cc @Naofel-eal @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

The panel converged that the fix preserves anonymous-first fallback, scoped credential use, token precedence, and diagnostic rendering while making auth-failure classification locale-independent. This retains the intended authorized retry instead of a locale-dependent failure path.

The two in-scope recommendations are folded in `93612c5d6`: the changelog now describes only the AuthResolver-managed authentication path, and the private loopback installed-CLI lifecycle seeds a non-English locale, records Git child locale values, asserts both are `C`, and retains the anonymous rejection, one path-scoped credential fill, authorized retry, and installed `SKILL.md` checks. Local exact-head unit coverage passed; the packaged lifecycle and CI remain to be observed on GitHub because this environment cannot build the packaged executable without unavailable dependencies.

**Aligned with:** Secure by default: locale-stable classification preserves anonymous-first, authorized-retry behavior without widening credential scope. Pragmatic as npm: the internal fix prevents locale-specific install failures without changing the CLI surface.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 0 | Locale normalization has one AuthResolver owner; the changelog scope is now precise. |
| CLI Logging Expert | 0 | 0 | 0 | Locale normalization preserves actionable auth retry behavior without changing diagnostic rendering. |
| DevX UX Expert | 0 | 0 | 0 | The fix preserves anonymous-first then authorized retry without a CLI surface change. |
| Supply Chain Security Expert | 0 | 1 | 0 | Auth child environments remain token-sanitized and credential precedence is unchanged. |
| OSS Growth Hacker | 0 | 0 | 0 | The changelog clearly states the user impact and no conversion surface regresses. |
| Auth Expert | 0 | 0 | 1 | Auth retry, token precedence, and credential fallback behavior are unchanged. |
| Documentation Writer | 0 | 1 | 0 | No doc-site update is needed; the release note scope is now accurate. |
| Test Coverage Expert | 0 | 2 | 0 | The locale-bearing private-loopback lifecycle coverage is now present. |

### Architecture

```mermaid
flowchart TD
    A["apm install"] --> B["GitHubPackageDownloader"]
    B --> C["AuthResolver.try_with_fallback"]
    C --> D["AuthResolver._build_git_env<br/>LC_ALL=C; LANGUAGE=C"]
    D --> E["anonymous Git clone"]
    E -->|auth failure| F["credential resolution"]
    F --> G["authenticated Git retry"]
    G --> H["install continues"]
```

### Recommendation

The implementation and in-scope review follow-ups are complete. Observe the newly queued GitHub check set before landing; no merge action was performed here.

---

<details>

#### Python Architect

- **[recommended]** Narrow the changelog claim to AuthResolver-managed Git operations at `CHANGELOG.md:35`.
  Folded in `93612c5d6`; the separate credential-fill subprocess is intentionally outside the classifier path.

#### CLI Logging Expert

No findings.

#### DevX UX Expert

No findings.

- **[recommended]** Scope the changelog claim to AuthResolver-managed Git operations at `CHANGELOG.md:35`.
  Folded in `93612c5d6`; token safety and credential precedence are unchanged.

#### OSS Growth Hacker

No findings.

#### Auth Expert

- **[nit]** Changelog locale scope at `CHANGELOG.md:35`.
  Folded in `93612c5d6`; Git credential fill remains intentionally outside the classifier path.

#### Documentation Writer

- **[recommended]** Narrow the changelog claim at `CHANGELOG.md:35`.
  Folded in `93612c5d6`; no documentation-site page drift was found.

#### Test Coverage Expert

- **[recommended]** Add locale-bearing private-loopback `apm install` coverage at `tests/integration/test_public_github_anonymous_lifecycle_e2e.py:249`.
  Folded in `93612c5d6`. The lifecycle now asserts the localized child-environment contract alongside the existing rejection, retry, credential-fill, and install assertions.

- **[recommended]** Add Scenario Evidence to the PR description.
  The lifecycle evidence is recorded in this advisory; the PR description remains unchanged.

Only AuthResolver child-environment mapping and tests changed; no performance surface was touched.
</details>

danielmeppiel:

> PR #2534 restores locale-independent auth classification, but its new installed-CLI lifecycle E2E deadlocks, leaving the user promise unproven.

cc @Naofel-eal @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

The load-bearing signal is the failed E2E: `tests/integration/test_public_github_anonymous_lifecycle_e2e.py::test_private_github_fallback_normalizes_locale_and_completes_lifecycle` timed out before reaching `assert result.returncode == 0`. The git shim buffers a long-lived `cat-file --batch-check` process, so the private-repository fallback lifecycle does not complete on this commit.

The remaining specialists agree that the production design is correctly centralized in `AuthResolver`, preserves token scoping and retry semantics, adds no CLI noise, and needs no architecture or documentation expansion. Once the shim streams bidirectional batch commands and the E2E passes, this is a focused reliability fix with no breaking surface.

**Aligned with:** Secure by default: locale normalization preserves classified, path-scoped credential fallback without widening token exposure, but the full lifecycle still needs passing E2E proof. OSS community driven: the changelog communicates the user benefit and credits the contributor. Pragmatic as npm: authentication should work consistently regardless of the machine's message locale.

**Growth signal.** Include this in the release reliability story rather than launching it separately.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 1 | Locale normalization is centralized in AuthResolver and covers anonymous and authenticated retry paths; no architecture changes requested. |
| CLI Logging Expert | 0 | 0 | 0 | Locale normalization preserves actionable auth guidance without adding CLI noise. |
| DevX UX Expert | 0 | 0 | 0 | Locale normalization restores private-repository credential retry without changing the CLI surface or recovery flow. |
| Supply Chain Security Expert | 0 | 0 | 0 | No supply-chain security concerns; locale pinning preserves credential classification without widening token exposure. |
| OSS Growth Hacker | 0 | 0 | 0 | The changelog clearly turns a locale-specific auth fix into user benefit and contributor credit; no growth concerns. |
| Auth Expert | 0 | 0 | 0 | Locale normalization is centralized in AuthResolver and preserves anonymous-to-credential retry semantics; no auth concerns found. |
| Doc Writer | 0 | 0 | 0 | The CHANGELOG records the user-visible fix; existing authentication and troubleshooting docs remain accurate with no documentation drift. |
| Test Coverage Expert | 1 | 1 | 0 | Auth unit tests pass, but the installed-CLI lifecycle test deadlocks in the modified git shim; Scenario Evidence is also missing. |

### Top 2 follow-ups

1. **[Test Coverage Expert] *(blocking-severity)*** Stop capturing output for long-lived bidirectional git batch commands and rerun the targeted lifecycle E2E. -- The current timeout is failed executable evidence for the core localized private-repository installation promise.
2. **[Test Coverage Expert]** Add the localized private-repository install scenario to the PR's Scenario Evidence table. -- The secure-by-default and DevX promise should map explicitly to its unit and E2E regression guards.

### Architecture

```mermaid
classDiagram
 direction LR
 class AuthResolver {
  <<ChainOfResponsibility>>
  +try_with_fallback()
  +_build_git_env() dict
  +is_public_github_auth_failure() bool
 }
 class GitHubTokenManager {
  <<CredentialProvider>>
  +resolve_credential_from_git()
 }
 class LocalePolicy {
  <<ImmutableMapping>>
  +LC_ALL C
  +LANGUAGE C
 }
 AuthResolver *-- GitHubTokenManager : resolves credentials
 AuthResolver ..> LocalePolicy : overlays
 class AuthResolver:::touched
 class LocalePolicy:::touched
```

```mermaid
flowchart TD
 A[apm install] --> B[anonymous Git clone]
 B --> C{auth failure classified?}
 C -- yes --> D[resolve path-scoped credential]
 D --> E[authenticated Git retry]
 E --> F[install completes]
 C -- no --> G[misleading network failure]
 H[AuthResolver._build_git_env] -->|LC_ALL=C and LANGUAGE=C| B
 H -->|LC_ALL=C and LANGUAGE=C| E
```

### Recommendation

Repair the git shim's streaming behavior and obtain a passing targeted installed-CLI lifecycle run before shipping; then add the concise Scenario Evidence row.

---

<details>

#### Python Architect

- **[nit]** No architectural follow-up is needed.
  AuthResolver remains the canonical chain, and the immutable locale mapping is the simplest correct design at this scope.

#### CLI Logging Expert

No findings.

#### DevX UX Expert

No findings.

No findings.

#### OSS Growth Hacker

No findings.

#### Auth Expert

No findings.

#### Doc Writer

No findings.

#### Test Coverage Expert

- **[blocking]** The auth lifecycle E2E deadlocks in the modified git shim at `tests/utils/git_credential_shim.py:180`
  The critical auth promise requires passing E2E evidence. `capture_output=True` buffers the long-lived `git cat-file --batch-check` subprocess output, preventing GitPython from receiving responses.
  *Suggested:* Capture output only for finite remote operations that need localized stderr synthesis; preserve inherited streaming for git batch commands. Then rerun the targeted E2E test.
  *Proof (failed):* `tests/integration/test_public_github_anonymous_lifecycle_e2e.py::test_private_github_fallback_normalizes_locale_and_completes_lifecycle` -- proves: A private GitHub dependency on a non-English machine retries with a token and completes installation. [secure-by-default,devx]
  `assert result.returncode == 0`
- **[recommended]** The PR body lacks the required Scenario Evidence table.
  Testing prose does not map the user scenario to an APM principle and executable test.
  *Suggested:* Add a Scenario Evidence row for localized private-repository installation with unit and E2E paths.
  *Proof (manual):* Reviewers can map the locale-specific retry to its promise and executable evidence. [secure-by-default,devx]

Only auth.py, CHANGELOG, auth tests, and the test shim changed; no performance trigger paths.

</details>

PR #2536, opened 2026-08-08
fix(plugins): normalize declared skills containers to one level (closes #2530)

edenfunf:
Closes #2530.

## Root cause

The issue reports this as specific to subdirectory plugin refs. It is not — it reproduces on any plugin whose manifest declares the conventional skills container, at repo root or in a subdirectory. The distinguishing factor is the **declaration style**, not the ref shape.

`_map_plugin_artifacts` copied each declared `skills` entry under its own name. That is correct when the entry *is* a skill:

```
"skills": ["./skills/engineering/tdd"]   ->  .apm/skills/tdd/SKILL.md      OK
```

But `dotnet/skills/plugins/dotnet-advanced` declares the container:

```
"skills": ["./skills/"]                  ->  .apm/skills/skills/<name>/SKILL.md
```

One level too deep. `.apm/skills/<name>/SKILL.md` is the exact depth that deployment, `--skill` enumeration, the `bin/` security scan and primitive counting all read, so the enumerator saw a single `skills` directory with no `SKILL.md` and reported `Available: (none)`.

The string form had the mirror defect: `"skills": "./skills/engineering/tdd"` merged the skill's *contents*, leaving a bare `SKILL.md` in the shared root under no name — same symptom, other manifest shape.

The bare install still worked because the deploy path never consulted `.apm/skills/` for these packages: it preferred a root `skills/` bundle, while `available_skill_names` carried a plugin-only branch reading `.apm/skills/`. Two different answers to "where do this package's skills come from".

For the record, the analysis in the issue comment (that `.apm/skills/` does not exist yet when the subset check runs) does not hold — the directory is present; its shape is wrong. A fallback in the enumerator would have masked the symptom while leaving the wrong on-disk shape in place.

## Changes

**`plugin_parser._holds_skill_dirs`** — classify each declared entry: one carrying its own `SKILL.md` keeps its name, one holding skills is merged. Applies to both the list and string forms. Undeclared discovery is unchanged; the default `skills/` is the convention container by definition.

**`skill_integrator.skill_source_dir`** — one helper now backs both deployment and `--skill` enumeration, replacing the plugin-only branch. A selectable name and a deployable name cannot drift apart again.

[...]

Issue #2530, opened 2026-08-07
[BUG] --skill matches nothing on subdirectory plugin refs: 'Available: (none)'


normandev92:
Subsetting skills works for repo-root plugin packages but fails for subdirectory plugin refs — the skill enumeration comes back empty, so any `--skill` value is rejected.

Repro:

```
apm install dotnet/skills/plugins/dotnet-advanced --skill csharp-scripts
```

Output:

```
[x] 1 package failed:
  +- dotnet/skills/plugins/dotnet-advanced -- --skill matched no declared skills in
'dotnet/skills/plugins/dotnet-advanced'. Requested: csharp-scripts. Available: (none).
[x] Installation failed with 1 error(s). No install transaction changes were committed.
```

The skill definitely exists: a bare `apm install dotnet/skills/plugins/dotnet-advanced` deploys `csharp-scripts` (plus `dotnet-pinvoke` and `nuget-trusted-publishing`) fine — it's the subset matching that can't see any skills. The package resolves as `package_type: marketplace_plugin` with a virtual subdirectory path.

For comparison, `apm install DuendeSoftware/duende-skills --skill ...` (a repo-root plugin) subsets correctly, so this looks specific to the subdirectory-ref intake path not feeding the enumerator that `--skill` validates against.

Expected: `--skill` filters subdirectory plugin refs the same way it does repo-root packages, persisting the `skills:` subset in apm.yml.

apm 0.27.0 on Windows 11.

*Disclosure: drafted with AI assistance; the behavior was reproduced and verified manually.*

normandev92:
Dug into this. The enumerator that `--skill` validates against is `SkillIntegrator.available_skill_names` (src/apm_cli/integration/skill_integrator.py:619), and it treats plugin packages differently from everything else:

```python
normalized = package_path / ".apm" / "skills"
root_bundle = package_path / "skills"
if package_info.package_type is PackageType.MARKETPLACE_PLUGIN:
    return SkillIntegrator._skill_names_in_directory(normalized)   # no fallback

root_names = SkillIntegrator._skill_names_in_directory(root_bundle)
return root_names or SkillIntegrator._skill_names_in_directory(normalized)  # fallback both ways
```

[...]

PR #2540, opened 2026-08-09
fix(plugins): let the resolved declaration decide which skills deploy (closes #2537)

### LINKED ISSUE #2537: [BUG] Plugin root skills/ overrides the plugin.json declaration: undeclared skills deploy, declared ones drop
A plugin's raw root `skills/` directory overrides its resolved `plugin.json` declaration, so the set APM deploys can differ from the set the manifest declares — in both directions.

Found while working on #2530; it is a distinct defect and is deliberately not fixed in #2536.

## Behavior

`_map_plugin_artifacts` resolves the manifest into `.apm/skills/`, where a declared `skills` array replaces default discovery rather than adding to it. Deployment does not read that result: it prefers a root `skills/` bundle whenever one holds at least one skill.

Three plugins, all with `skills/{a,b}` at the root:

| `plugin.json` `skills` | resolved into `.apm/skills` | actually deployed / selectable |
| --- | --- | --- |
| `["./skills/a"]` | `a` | `a`, `b` — undeclared `b` leaks in |
| `["./skills/", "./extra/solo"]` | `a`, `b`, `solo` | `a`, `b` — declared `solo` dropped |
| `["./skills/a", "./skills/b"]` | `a`, `b` | `a`, `b` — agrees |

The first row is the more likely one in practice: a plugin that ships several skills but declares only the ones it wants published still gets all of them deployed, with no diagnostic.

## Reproduction

```
mkdir -p plug/.claude-plugin plug/skills/a plug/skills/b
printf '# a\n' > plug/skills/a/SKILL.md
printf '# b\n' > plug/skills/b/SKILL.md
printf '{"name":"p","version":"1.0.0","skills":["./skills/a"]}' > plug/.claude-plugin/plugin.json

mkdir -p consumer/.claude
printf 'name: c\nversion: 1.0.0\ntargets: [claude]\ndependencies:\n  apm:\n    - path: ../plug\n' > consumer/apm.yml
cd consumer && apm install
ls .claude/skills          # a and b; only a was declared
```

`apm install --skill b` is likewise accepted, because `available_skill_names` enumerates the same raw directory.

## The open question

Does declaring `skills:` in `plugin.json` opt out of default `skills/` discovery?

Normalization already answers yes — `_resolve_sources` uses the manifest value *instead of* the default. Deployment answers no. That internal disagreement is the bug; which answer is correct is a product decision.

If "declared paths are requirements" (per the Plugin collection section of the package-types reference) means the declared set is exhaustive, the fix is to let the resolved `.apm/skills/` win for `MARKETPLACE_PLUGIN` in `SkillIntegrator.skill_source_dir`. That would also make `"skills": []` mean "no skills" as documented, and would fail closed when a declared path is rejected for escaping the plugin root.

It is worth noting this changes what existing consumers receive: any plugin declaring a partial list would start deploying fewer skills. That is why it is filed rather than bundled into #2536.

## Notes

- `#2536` makes enumeration and deployment agree with each other, so `--skill` no longer reports a set that differs from what deploys. It does not change which of the two sources wins.
- Neither `dotnet/skills/plugins/dotnet-advanced` (declares the whole container) nor `DuendeSoftware/duende-skills` (declares all 24 individually) triggers this.
- Environment: reproduced against `main` at 3aa036554 on Windows 11.

edenfunf:
Closes #2537. Stacked on #2536 — review that one first; this branch carries its two commits, and the diff below is the third one plus the review follow-ups.

## Problem

A plugin's raw root `skills/` directory overrode its resolved `plugin.json` declaration, so the deployed set could differ from the declared one in both directions — silently.

With `skills/{a,b}` at the plugin root:

| `plugin.json` `skills` | resolved into `.apm/skills` | deployed before | deployed now |
| --- | --- | --- | --- |
| `["./skills/a"]` | `a` | `a`, `b` | `a` |
| `["./skills/", "./extra/solo"]` | `a`, `b`, `solo` | `a`, `b` | `a`, `b`, `solo` |
| `["./skills/a", "./skills/b"]` | `a`, `b` | `a`, `b` | `a`, `b` |

Normalization already resolves the manifest into `.apm/skills/`, where a declared array replaces default discovery rather than adding to it. Deployment ignored that result and re-derived its own set from the pre-resolution tree.

## Change

`skill_source_paths` maps each deployable skill name to the directory it is promoted from, and backs both deployment and `--skill` enumeration. For a plugin the resolved declaration decides which skills exist.

Two consequences fall out of treating the resolution as authoritative, both intended:

- `"skills": []` now means no skills, as the package-types reference documents.
- A declared component rejected for escaping the plugin root fails closed instead of falling back to whatever the raw tree holds.

## Which skills is not where their content comes from

Each declared skill keeps being **sourced** from the root bundle whenever that copy exists. It is byte-identical to the normalized one but sits one level higher, and that matters: a relative link leaving the skill bundle is rewritten against the package root.

I first implemented this by simply routing plugins to `.apm/skills/`, and measured the result against the #2536 branch with the same plugin:

```
#2536  escaping: [guide](../../../apm_modules/_local/plug/docs/guide.md)   rewritten, resolves
naive  escaping: [guide](../../docs/guide.md)                             untouched, points into .claude/
```

`.apm/skills/alpha/` is a directory deeper than `skills/alpha/`, so the author's `../../docs/guide.md` no longer lands inside the package and the resolver leaves it alone. Preferring the root copy keeps that behaviour byte-identical. `test_plugin_skill_links_leaving_the_bundle_survive_declaration_filtering` installs a skill carrying such a link and asserts it is rewritten back at the package; it fails if the preference is removed.

Only skills declared from outside the root container — which never deployed at all before — are read from the normalized copy.

Non-plugin packages are untouched: nothing normalizes their `skills/`, so the raw directory remains the declaration, and `.apm/skills/` stays what it has always been for them — sub-skills promoted alongside a package that is not itself a skill.

## Verification

End-to-end, real packages, unchanged from #2536 including the result shape (`Skill integrated ->` line and counts):

- `dotnet/skills/plugins/dotnet-advanced` — declares the whole container; `--skill csharp-scripts` deploys exactly that skill.
- `DuendeSoftware/duende-skills` — declares its 24 skills individually; 24 skills + 2 agents.

Neither triggers the defect, which is why it went unnoticed; both are here as regression cover for the routing change.

Local plugins covering all three rows of the table above deploy exactly the declared set.

Full suite diffed against clean `main` on the same machine: identical failure sets, `100 subtests passed` on both, +9 from the new tests across both commits. Each new test was toggle-verified — reverting the corresponding change individually turns it red.

## Review follow-ups

The advisory panel shipped with follow-ups; all five landed in `fb96e932c`.

- **CHANGELOG.** `[Unreleased] → Fixed` entry that leads with the behaviour change for plugin authors — a manifest declaring a subset of what `skills/` holds stops deploying the rest — and how to keep the old set.
- **Docs.** `deploy-a-bundle.md` step 4 and the plugin-collection section of the package-types reference now both state that a declared key replaces the directory scan rather than adding to it.
- **Zero deployable skills is no longer silent.** Most plugins ship no skills and none of them should be nagged, so the note fires only at the shape where the answer changed: a declaration resolving to nothing while the package does carry a root `skills/` bundle. It names the skills left behind; the fix guidance sits behind `--verbose`. Emitted once per install regardless of target count.
- **`skill_source_dir` → `_skill_source_dir`.** It answers with a single directory, which is incomplete for a plugin, and reading it directly on a plugin path is how the raw tree came to override the declaration. `skill_source_paths` is the routing entry point; the private name keeps that structural rather than documentary. Introduced in #2530 and unmerged, so nothing outside this stack can be relying on it.
- **`_holds_skill_dirs` fail-closed branches.** Direct tests for the empty directory, skills buried a level deeper, a directory that vanished, and one that cannot be listed; precedence pinned with both shapes present at once.

Spec conformance took the citation route rather than a waiver. Section 8.1 settles which container holds a plugin collection's individually addressable skills — artifacts are mapped per the plugin manifest — and req-mf-022 requires the no-match diagnostic to name the available ones. Before the fix it named, and deployed, siblings the manifest never declared. `CONFORMANCE.{json,md}` regenerated.

Every follow-up test was toggle-verified in both directions: reverting each corresponding change individually turns exactly the matching test red.

### DIFF
--- a/.apm/instructions/architecture.instructions.md
+++ b/.apm/instructions/architecture.instructions.md
@@ -80,6 +80,7 @@ semicolon-delimited, and specific to the file(s) that own the fact.
 | APMPackage interpreted-manifest construction | models/apm_package.py (APMPackage.from_mapping) | `src/apm_cli/models/apm_package.py` |
 | Agent Plugin compatibility package projection | agent_plugins/projection.py (project_agent_plugin_package) | `src/apm_cli/agent_plugins/projection.py`; `src/apm_cli/models/validation.py` |
 | Network host literal parsing and loopback classification | utils/net.py (parse_host_address, is_loopback_host) | `src/apm_cli/utils/net.py` |
+| Legacy plugin declared-skill membership | deps/plugin_parser.py (_map_plugin_artifacts, normalized_plugin_skill_sources) | `src/apm_cli/deps/plugin_parser.py`; `src/apm_cli/integration/skill_integrator.py` |
 <!-- /canonical-owner-table -->
 
 Host + credential resolution includes public github.com anonymous-first ordering.
--- a/.github/instructions/architecture.instructions.md
+++ b/.github/instructions/architecture.instructions.md
@@ -80,6 +80,7 @@ semicolon-delimited, and specific to the file(s) that own the fact.
 | APMPackage interpreted-manifest construction | models/apm_package.py (APMPackage.from_mapping) | `src/apm_cli/models/apm_package.py` |
 | Agent Plugin compatibility package projection | agent_plugins/projection.py (project_agent_plugin_package) | `src/apm_cli/agent_plugins/projection.py`; `src/apm_cli/models/validation.py` |
 | Network host literal parsing and loopback classification | utils/net.py (parse_host_address, is_loopback_host) | `src/apm_cli/utils/net.py` |
+| Legacy plugin declared-skill membership | deps/plugin_parser.py (_map_plugin_artifacts, normalized_plugin_skill_sources) | `src/apm_cli/deps/plugin_parser.py`; `src/apm_cli/integration/skill_integrator.py` |
 <!-- /canonical-owner-table -->
 
 Host + credential resolution includes public github.com anonymous-first ordering.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (`"skills": "./skills/engineering/tdd"`) still lands under its own leaf name
   instead of spilling a bare `SKILL.md` into the shared skills root.
   (closes #2530)
+- A plugin's `plugin.json` is now authoritative for which skills deploy. The
+  raw root `skills/` directory overrode the resolved declaration, so the
+  deployed set could differ from the declared one in both directions --
+  silently: declaring `"skills": ["./skills/a"]` still deployed sibling `b`
+  alongside it, and a skill declared outside the conventional container was
+  dropped entirely. A declaration replaces default discovery rather than
+  adding to it, so `"skills": []` now means no skills, and a declared path
+  rejected for escaping the plugin root fails closed instead of falling back
+  to the raw tree. **Behaviour change for plugin authors:** if your
+  `plugin.json` declares a `skills` subset while shipping more skills under
+  `skills/`, the undeclared ones stop deploying -- declare them, or drop the
+  key to keep deploying everything the directory holds. When a declaration
+  resolves to no skills while the package does carry a root `skills/` bundle,
+  install now names the skills it left behind instead of saying nothing.
+  Where a declared skill also exists in the root bundle it keeps being read
+  from there, so relative links leaving a skill still resolve against the
+  package root. (closes #2537)
 - Multi-target `apm compile` now avoids repeating expensive project analysis
   for each target, making multi-target runs scale like single-target runs
   without changing generated output. (closes #2482)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Windows binary is now Authenticode-signed in the release workflow, eliminating
   the `Trojan:Script/Wacatac.H!ml` Windows Defender false positive on unsigned
   PyInstaller bundles. (#2435)
+- `apm install --skill <name>` now matches on plugins whose manifest declares
+  the conventional skills container (`"skills": ["./skills/"]`). The declared
+  container was normalized under its own name, burying every skill at
+  `.apm/skills/skills/<name>/` -- one level below the depth `--skill`
+  enumeration, deployment, the `bin/` security scan and primitive counting all
+  read, so selection reported `Available: (none)` even though a bare install
+  deployed those same skills. A declared entry that is itself a skill
+  (`"skills": "./skills/engineering/tdd"`) still lands under its own leaf name
+  instead of spilling a bare `SKILL.md` into the shared skills root.
+  (closes #2530)
 - Multi-target `apm compile` now avoids repeating expensive project analysis
   for each target, making multi-target runs scale like single-target runs
   without changing generated output. (closes #2482)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,23 +58,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (`"skills": "./skills/engineering/tdd"`) still lands under its own leaf name
   instead of spilling a bare `SKILL.md` into the shared skills root.
   (closes #2530)
-- A plugin's `plugin.json` is now authoritative for which skills deploy. The
-  raw root `skills/` directory overrode the resolved declaration, so the
-  deployed set could differ from the declared one in both directions --
-  silently: declaring `"skills": ["./skills/a"]` still deployed sibling `b`
-  alongside it, and a skill declared outside the conventional container was
-  dropped entirely. A declaration replaces default discovery rather than
-  adding to it, so `"skills": []` now means no skills, and a declared path
-  rejected for escaping the plugin root fails closed instead of falling back
-  to the raw tree. **Behaviour change for plugin authors:** if your
-  `plugin.json` declares a `skills` subset while shipping more skills under
-  `skills/`, the undeclared ones stop deploying -- declare them, or drop the
-  key to keep deploying everything the directory holds. When a declaration
-  resolves to no skills while the package does carry a root `skills/` bundle,
-  install now names the skills it left behind instead of saying nothing.
-  Where a declared skill also exists in the root bundle it keeps being read
-  from there, so relative links leaving a skill still resolve against the
-  package root. (closes #2537)
 - Multi-target `apm compile` now avoids repeating expensive project analysis
   for each target, making multi-target runs scale like single-target runs
   without changing generated output. (closes #2482)
@@ -95,6 +78,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** A plugin `skills` declaration now exclusively controls which
+  skills deploy. Declare every intended skill (or its immediate container), or
+  remove the key to retain conventional `skills/` discovery; undeclared
+  siblings no longer deploy. An explicit `"skills": []` deploys no skills and
+  reports one actionable migration diagnostic when it shadows root skills.
+  (closes #2537)
 - `apm install` now emits a trust-posture warning (via `[!]`) when a marketplace
   plugin deploys executables to Claude Code's PATH without an explicit `--trust-bin`
   flag. In non-interactive (non-TTY) contexts the default is `--no-trust-bin`.
--- a/CONFORMANCE.json
+++ b/CONFORMANCE.json
@@ -544,9 +544,10 @@
       "keyword": "MUST",
       "section": "4.3.2",
       "status": "active",
-      "test_count": 1,
+      "test_count": 2,
       "tests": [
-        "tests/spec_conformance/test_manifest_reqs.py::test_consumer_diagnoses_empty_skill_subset_match"
+        "tests/spec_conformance/test_manifest_reqs.py::test_consumer_diagnoses_empty_skill_subset_match",
+        "tests/spec_conformance/test_manifest_reqs.py::test_consumer_names_skills_a_plugin_collection_exposes"
       ]
     },
     {
--- a/CONFORMANCE.json
+++ b/CONFORMANCE.json
@@ -544,9 +544,10 @@
       "keyword": "MUST",
       "section": "4.3.2",
       "status": "active",
-      "test_count": 2,
+      "test_count": 3,
       "tests": [
         "tests/spec_conformance/test_manifest_reqs.py::test_consumer_diagnoses_empty_skill_subset_match",
+        "tests/spec_conformance/test_manifest_reqs.py::test_consumer_names_only_the_skills_a_plugin_manifest_declares",
         "tests/spec_conformance/test_manifest_reqs.py::test_consumer_names_skills_a_plugin_collection_exposes"
       ]
     },
--- a/CONFORMANCE.json
+++ b/CONFORMANCE.json
@@ -544,10 +544,9 @@
       "keyword": "MUST",
       "section": "4.3.2",
       "status": "active",
-      "test_count": 3,
+      "test_count": 2,
       "tests": [
         "tests/spec_conformance/test_manifest_reqs.py::test_consumer_diagnoses_empty_skill_subset_match",
-        "tests/spec_conformance/test_manifest_reqs.py::test_consumer_names_only_the_skills_a_plugin_manifest_declares",
         "tests/spec_conformance/test_manifest_reqs.py::test_consumer_names_skills_a_plugin_collection_exposes"
       ]
     },
@@ -822,6 +821,17 @@
       "test_count": 1,
       "tests": [
         "tests/spec_conformance/test_resolution_reqs.py::test_producer_should_carry_primitive_descriptions"
+      ]
+    },
+    {
+      "conformance_class": "consumer",
+      "id": "req-pr-006",
+      "keyword": "MUST",
+      "section": "8.1",
+      "status": "active",
+      "test_count": 1,
+      "tests": [
+        "tests/spec_conformance/test_manifest_reqs.py::test_plugin_skills_declaration_is_authoritative"
       ]
     },
     {
@@ -1315,7 +1325,7 @@
   "spec_version": "v0.1.1",
   "summary_by_class": {
     "consumer": {
-      "active": 84,
+      "active": 85,
       "skipped": 1,
       "unbound": 0,
       "xfail": 0
@@ -1339,5 +1349,5 @@
       "xfail": 0
     }
   },
-  "total_requirements": 115
+  "total_requirements": 116
 }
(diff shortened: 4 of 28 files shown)


sergio-sisternes-epam:

> Plugin skill deployment now honors plugin.json as the single authority, closing a silent over-deploy bug that undermined manifest auditability.

cc @edenfunf @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

All nine panelists converge on the same verdict: this is a net-positive correctness fix with no blocking defects. The core change -- making plugin.json's skills array authoritative over filesystem scanning -- aligns directly with APM's manifest-as-contract promise. Supply-chain-security confirms a smaller attack surface; python-architect validates the routing centralization; performance-expert finds no regressions. The only substantive cross-panel signal is a three-way agreement (devx-ux, oss-growth-hacker, doc-writer) that the behavioral change ships without a CHANGELOG entry or doc update, which is a communication gap, not a code gap.

The test-coverage-expert's single finding (`_holds_skill_dirs` edge-case coverage marked missing on a secure-by-default surface) is nit-severity and exercised transitively through integration fixtures. Per the weighting contract, missing unit coverage on a secure-by-default helper elevates slightly above pure nit, but the integration-tier transitive exercise and small blast radius keep it firmly in follow-up territory.

No dissent exists between panelists on severity classification. The cli-logging-expert's "empty skills array = silent no-op" finding and the python-architect's `skill_source_dir()` vestigial exposure finding are complementary -- both point at the same future misuse vector from different angles. Neither blocks ship.

**Aligned with:** portable-by-manifest (plugin.json declaration is now the sole deploy authority -- manifest portability strengthened), secure-by-default (shrinks deploy surface to exactly what the manifest declares; no undeclared code can sneak into the harness), pragmatic-as-npm (matches npm's package.json `files` semantics: declare what ships, nothing more).

**Growth signal.** plugin.json-as-authoritative is a trust primitive for enterprise plugin distribution. Package consumers can audit exactly what deploys by reading the manifest -- strong story beat for "why APM is safe for regulated environments". The CHANGELOG gap is the only adoption risk; once documented this converts a bug-fix PR into a positioning win.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 2 | Clean single-authority fix; `skill_source_paths()` correctly centralizes the skill-set decision; `skill_source_dir()` is the only structural debt worth noting. |
| CLI Logging Expert | 0 | 1 | 1 | Main gap is a silent empty-result path: when `plugin.json` declares `"skills": []`, the bundle path is skipped with zero user feedback. |
| DevX UX Expert | 0 | 1 | 2 | Fix is UX-positive: `plugin.json` becomes the single source of truth; behavioral breaking change ships without CHANGELOG entry or migration guidance. |
| Supply Chain Security Expert | 0 | 0 | 2 | Net security improvement; declaration-authoritative deployment shrinks attack surface; AST-based copytree guard is strictly stronger than old occurrence count. |
| OSS Growth Hacker | 0 | 1 | 1 | Strong ecosystem-trust win; missing CHANGELOG entry for behavioral change is a support-ticket generator. |
| Doc Writer | 0 | 2 | 0 | Two doc gaps confirmed: CHANGELOG [Unreleased] missing entry; deploy-a-bundle.md step 4 omits the `plugin.json` skills-key override. |
| Test Coverage Expert | 0 | 0 | 1 | Critical surfaces covered at integration-with-fixtures tier via CliRunner and real tmp_path fixtures; `_holds_skill_dirs` edge cases lack direct coverage. |
| Performance Expert | 0 | 0 | 3 | No algorithmic regressions; all new paths are O(N) in declared skill count where N is realistically 1-10. |

### Top 5 follow-ups

1. **[DevX UX Expert]** Add CHANGELOG `[Unreleased]` entry for the behavioral change -- Three panelists independently flagged this; silent behavior changes that remove previously-deployed skills are support-ticket generators without a migration line.
2. **[Doc Writer]** Update `deploy-a-bundle.md` step 4 to note that the `plugin.json` `skills` key overrides directory scan -- Plugin authors reading the guide will not discover the authoritative semantics without this addition.
3. **[CLI Logging Expert]** Emit a verbose diagnostic when a plugin declares zero deployable skills -- Distinguishes "no skills by design" from "declaration bug silently ate my skills" -- important for plugin author DX.
4. **[Python Architect]** Underscore-prefix or inline `skill_source_dir()` to prevent future misuse on plugin paths -- Vestigial public method creates a dual-read risk; structural prevention is cheap.
5. **[Test Coverage Expert]** Add direct unit tests for `_holds_skill_dirs` OSError and empty-dir branches -- Missing coverage on a secure-by-default helper; currently exercised only transitively.

### Architecture

```mermaid
classDiagram
    direction LR
    class SkillIntegrator {
        +skill_source_dir(package_path) Path
        +skill_source_paths(package_path, package_type) dict
        +available_skill_names(package_info) frozenset
        -_skill_names_in_directory(dir) list
        +integrate(package_info, project_root) Result
    }
    class BaseIntegrator {
        +integrate()
    }
    class PackageType {
        MARKETPLACE_PLUGIN
        SKILL
        AGENT
    }
    class plugin_parser {
        +_holds_skill_dirs(candidate) bool
        +_map_plugin_artifacts(manifest)
    }
    BaseIntegrator <|-- SkillIntegrator
    SkillIntegrator ..> PackageType : guards routing
    plugin_parser ..> SkillIntegrator : normalizes .apm/skills
    class SkillIntegrator:::touched
    class plugin_parser:::touched
```

```mermaid
flowchart TD
    A["plugin_parser._map_plugin_artifacts\ncopytree declared skills into .apm/skills/"] --> B["SkillIntegrator.skill_source_paths()\nEnumerate .apm/skills/ (authoritative set)"]
    B --> C{MARKETPLACE_PLUGIN?}
    C -->|Yes| D["Per-name: prefer root skills/name\nif SKILL.md exists there"]
    C -->|No| E["skill_source_dir fallback:\nroot skills/ or .apm/skills/"]
    D --> F["dict: name to source Path"]
    E --> F
    F --> G["integrate(): gate on _source_paths non-empty"]
    G --> H["_integrate_skill_bundle(source_paths=...)\ncopytree each skill to project"]
    F --> I["available_skill_names()\nfrozenset(source_paths.keys())\nUsed by --skill validation"]
```

```mermaid
sequenceDiagram
    participant PP as plugin_parser
    participant SI as SkillIntegrator
    participant FS as Filesystem
    CLI->>PP: _map_plugin_artifacts(manifest, pkg_dir)
    PP->>FS: copytree declared skill dirs into .apm/skills/
    PP-->>CLI: normalized package on disk
    CLI->>SI: integrate(package_info, project_root)
    SI->>SI: skill_source_paths(path, MARKETPLACE_PLUGIN)
    SI->>FS: list .apm/skills/ children (authoritative)
    SI->>FS: check root skills/name/SKILL.md (preferred source)
    SI-->>SI: dict of name to source_path
    SI->>FS: copytree each source to project .claude/skills/name
    SI-->>CLI: SkillIntegrationResult
```

### Recommendation

Zero blocking findings across all panelists. Security posture improves. Test coverage at integration tier is adequate for the changed surfaces. The CHANGELOG and doc gaps are real communication debt but do not affect correctness -- they should land in the same release cycle (ideally same day) but need not gate this merge. Ship now; file a follow-up issue for the five recommended items above.

---

<details>

#### Python Architect

- **[recommended]** `skill_source_dir()` is now vestigial for plugins and creates a dual-read risk at `src/apm_cli/integration/skill_integrator.py:625`
  For MARKETPLACE_PLUGIN, `skill_source_paths()` is authoritative -- yet `skill_source_dir()` still exists and is called in the integrate() gate condition. Any future caller using `skill_source_dir()` for a plugin would bypass the declaration filter. The docstring warns but nothing prevents misuse structurally.
  *Suggested:* Mark `skill_source_dir` with a `# INTERNAL` comment and prefix-underscore it, or inline its logic into `skill_source_paths` non-plugin branch.

- **[nit]** Redundant filesystem scan in integrate() gate at `src/apm_cli/integration/skill_integrator.py:1544`
  `skill_source_dir()` called again for non-plugin gate condition even though `_source_paths` was already computed from the same scan.

- **[nit]** `_holds_skill_dirs` duplicates the SKILL.md detection idiom already in `_skill_names_in_directory` at `src/apm_cli/deps/plugin_parser.py:147`
  Tolerable today but worth a shared comment if the marker ever changes.

#### CLI Logging Expert

- **[recommended]** Empty `skill_source_paths` result produces no diagnostic when a MARKETPLACE_PLUGIN has zero deployable skills at `src/apm_cli/integration/skill_integrator.py`
  When `_source_paths` is empty, the condition is falsy and falls through silently. User cannot distinguish "plugin has no skills by design" from "declaration bug silently ate my skills".
  *Suggested:* When `_is_plugin and not _source_paths`, emit verbose diagnostic: `Plugin 'X' declares no skills -- nothing to deploy`.

- **[nit]** `_surface_warning` in plugin_parser should route through `_logger` rather than `_rich_warning` directly at `src/apm_cli/deps/plugin_parser.py`
  Direct use of `_rich_warning` bypasses verbose gating.

#### DevX UX Expert

- **[recommended]** No CHANGELOG entry communicates the behavioral breaking change at `CHANGELOG.md`
  A plugin author whose `plugin.json` declares `["./skills/a"]` but ships `skills/a` and `skills/b` will lose `b` on upgrade with no prior notice.
  *Suggested:* Add Fixed entry: "Plugin `skills` declarations are now authoritative -- only paths listed in `plugin.json` deploy. Previously, undeclared skills in the raw `skills/` directory were silently included."

- **[nit]** Error message for missing declared component could name both fix actions
  "Fix the path in plugin.json or remove the declaration if the component is no longer needed" covers both recovery routes.

- **[nit]** `"skills": []` = nothing semantic deserves a one-liner in plugin authoring docs
  Plugin author reading the spec won't know empty array is meaningful and intentional rather than a no-op.

- **[nit]** `_holds_skill_dirs` swallows OSError broadly at `src/apm_cli/deps/plugin_parser.py`
  Catching all `OSError` on `candidate.iterdir()` means permission errors and broken mounts silently return False. Not exploitable given prior containment checks, but narrowing to `(PermissionError, FileNotFoundError)` would prevent masking unexpected conditions.

- **[nit]** `skill_integrator.py` copytree guard in `TestSkillIntegratorCopytreeSymlinkContainment` still uses substring count, not AST at `tests/unit/test_symlink_containment.py`
  Weaker than the upgraded AST guard now used for the other modules; inconsistent with the upgraded standard.

#### OSS Growth Hacker

- **[recommended]** Missing CHANGELOG entry for user-visible behavioral change
  Plugin authors whose packages accidentally deployed undeclared skills will see silent skill loss on upgrade. Without a CHANGELOG line, affected authors hit a mystery regression.
  *Suggested:* Add under [Unreleased] > Fixed: "Plugin skill deployment now respects `plugin.json` declaration as authoritative. (closes #2537, #2530)"

- **[nit]** Opportunity for a one-line story-shaped hook in docs
  "Your declaration is your deploy set, period" is a repostable claim worth surfacing in plugin authoring docs or a release note.

#### Auth Expert -- inactive

No auth-surface files changed; PR touches only plugin skill routing and deployment in plugin_parser.py and skill_integrator.py.

#### Doc Writer

- **[recommended]** CHANGELOG.md [Unreleased] has no entry for the plugin.json skills-authoritative behavior change at `CHANGELOG.md`
  [Unreleased] block contains only a single Fixed entry for `apm compile` performance. This PR changes observable deployment behavior and belongs in the changelog.

- **[recommended]** `deploy-a-bundle.md` step 4 omits that `plugin.json` `skills` key overrides directory scan at `docs/src/content/docs/consumer/deploy-a-bundle.md`
  Step 4 describes mapping the `skills/` directory but does not mention the skills key filter, leaving plugin authors without documentation of the authoritative semantics.

#### Test Coverage Expert

- **[nit]** `_holds_skill_dirs` OSError and empty-dir branches lack direct unit coverage at `src/apm_cli/deps/plugin_parser.py`
  Exercised transitively through `_map_plugin_artifacts` tests but OSError catch and no-skill-child branch are never directly triggered. Defensive edges; nit severity.
  *Proof (missing):* `tests/unit/test_plugin_parser.py::test_holds_skill_dirs_edge_cases` -- proves: Defensive edges of container detection do not regress silently [secure-by-default]

#### Performance Expert

- **[nit]** Non-plugin path scans root_bundle directory twice via `skill_source_dir()` at `src/apm_cli/integration/skill_integrator.py`
  `skill_source_dir()` called at line 1544 gate condition after `skill_source_paths()` already called it internally. ~20 extra stat() syscalls; <1ms for typical N=1-10. Trivially fixable.

- **[nit]** Deferred import of PackageType inside `skill_source_paths()` -- no action needed.
  Correct pattern for avoiding circular imports; single dict lookup in sys.modules after first load.

- **[nit]** `_holds_skill_dirs()` does iterdir() per declared entry -- acceptable O(K*M) at `src/apm_cli/deps/plugin_parser.py`
  Both K (declared skills) and M (children per dir) are small constants in practice.

</details>

danielmeppiel:

> Make plugin declarations authoritative without letting stale projections, unsafe links, or source-path drift undermine predictable skill deployment.

cc @edenfunf @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

The panel agrees that resolved membership must decide which plugin skills deploy, while byte sourcing remains separate. The current implementation needs in-scope repairs before release: generated `.apm/skills` membership can survive a declaration change, and staged or root symlinks can influence deployment outside validated declaration membership. Preserve the parser-owned normalized-name-to-original-source map so nested declarations retain package-relative links, and make the ownership boundary executable.

**Dissent.** No specialist disputes the direction. The difference is priority: declaration reconciliation and containment come first; diagnostics, docs, performance, and changelog polish follow in the same PR.

**Aligned with:** portable manifests, secure-by-default containment, predictable reinstall and audit behavior, and an actionable contributor migration.

### Panel summary

|---|---:|---:|---:|---|
| Python Architect | 2 | 0 | 0 | Preserve original source mapping and add the ownership guard. |
| CLI Logging | 0 | 1 | 0 | Make the exactly-once migration diagnostic actionable by default. |
| DevX UX | 1 | 0 | 0 | Reinstall must remove obsolete normalized skills. |
| Supply Chain Security | 1 | 0 | 0 | Validated declaration membership and containment must control bytes. |
| OSS Growth | 0 | 1 | 0 | Lead changelog migration text with the author action. |
| Doc Writer | 0 | 4 | 1 | Complete package guide, spec, producer docs, and breaking-change framing. |
| Test Coverage | 0 | 1 | 0 | Add the required installed-CLI lifecycle and mutation breaks. |
| Performance | 0 | 1 | 0 | Avoid the redundant non-plugin root scan. |

> B = blocking-severity findings, R = recommended, N = nits. Counts are signal strength, not gates.

### Top follow-ups

1. **[DevX UX] (blocking-severity)** Rebuild or reconcile generated `.apm/skills` on declaration changes, reinstall, and audit so stale members cannot remain deployable.
2. **[Supply Chain Security] (blocking-severity)** Derive deployment only from validated declared membership and reject root symlinks or external byte sources.
3. **[Python Architect] (blocking-severity)** Persist the normalized-name-to-original-source mapping in `plugin_parser._map_plugin_artifacts`, consume it from the integrator, and add an owner row and static boundary guard.
4. **[Test Coverage]** Add installed-candidate CLI lifecycle coverage for declaration removal, reinstall, audit, and two mutation breaks.
5. **[Doc Writer]** Complete concise package-authoring, Section 8.1, producer-doc, and breaking-change migration guidance.

### Recommendation

Address stale projection reconciliation, validated-member-only sourcing with symlink containment, and parser-owned source mapping before shipping. Then add lifecycle and mutation coverage and complete the concise author-facing migration message.

danielmeppiel:
## APM Spec Guardian

> Scope: **editorial-patch**. The four-person advisory review completed two rounds.

The first round required an anchored Section 8.1 consumer requirement. The second round identified root-base, symlink-containment, canonical-path, duplicate-leaf, and line-wrapping details; all were folded into `req-pr-006` after the returns. The panel cap is now exhausted, so this record preserves the round-two feedback without opening another round.

The requirement is projected through Section 8.7, Section 11.3.2, Appendix C, the requirements manifest, and generated conformance artifacts (114 statements: 109 MUST, 5 SHOULD). Focused parser, authority, outcome, and conformance checks passed.

<sub>This review is advisory; maintainers retain the shipping decision.</sub>

PR #2577, opened 2026-08-15
fix: stop writing generated_at to apm.lock.yaml (#2572)

### LINKED ISSUE #2572: [BUG] `generated_at` key in `apm-lock.yml` is constant source of merge conflicts.
**Describe the bug**
The `generated_at` key is consistently causing merge conflicts. Example: if two PRs separately modify different skills in the same repo, the `apm-lock.yml` file is regenerated creating two mergeable diffs _except_ for the `generated_at` key.

**To Reproduce**
1. Create a monorepo. Add two skills locally.
2. Make a change to skill A on a branch, create a PR.
3. Make a change to skill B on a branch, create a PR.
4. Merge PR for branch A.
5. PR B will now have a merge conflict ONLY on the line with `generated_at` key.

**Expected behavior**
Merging unrelated changes should not cause merge conflicts.

**Additional context**
As far as I can tell, there is no real reason for the inclusion of a timestamp in the lockfile and suggest it should be removed. At the very least, I should be able to optionally disable it entirely.

aryansk:
Closes #2572

## Summary

`generated_at` in `apm.lock.yaml` is a constant source of merge conflicts — two unrelated PRs that regenerate the lockfile collide only on this one line (reported in #2297 and again in #2572). This PR stops writing the key entirely, which is exactly what the triage suggested ("Open a PR that removes the `generated_at` field from `apm-lock.yml` generation").

## Changes

- `src/apm_cli/deps/lockfile.py`: remove `generated_at` from `LockFile.to_dict()` output (the `to_yaml` write path). The field is still parsed from legacy lockfiles so existing projects load unchanged and `apm lock export` keeps its timestamp fallback.
- `src/apm_cli/integration/mcp_integrator.py`: drop the now-dead `lockfile.generated_at = datetime.now(...)` stamp before save; remove the unused import.
- `tests/unit/install/test_mcp_lockfile_determinism.py`: the determinism tests now assert `generated_at` is absent from the written YAML (the actual regression for #2572) instead of asserting it round-trips.

## Validation

- `pytest tests/unit` → 19,802 passed, 2 skipped, 21 xfailed
- `ruff check` + `ruff format --check` clean on all touched files

## Compatibility

Legacy lockfiles that already contain `generated_at` keep parsing and the SBOM export fallback keeps working (it reads whatever the lockfile has). Newly written lockfiles simply omit the key.

### DIFF
--- a/src/apm_cli/deps/lockfile.py
+++ b/src/apm_cli/deps/lockfile.py
@@ -799,9 +799,13 @@ class LockFile:
         # since the flat fields remain the source of truth in YAML.
         _self_dep = self.dependencies.pop(_SELF_KEY, None)
         try:
+            # ``generated_at`` is deliberately NOT written (issue #2572): a
+            # per-run timestamp in the lockfile is a constant source of merge
+            # conflicts across unrelated branches. The field is still parsed
+            # from legacy lockfiles so ``apm lock export`` keeps its timestamp
+            # fallback for existing projects.
             data: dict[str, Any] = {
                 "lockfile_version": emit_version,
-                "generated_at": self.generated_at,
             }
             if self.apm_version:
                 data["apm_version"] = self.apm_version
@@ -1012,8 +1016,8 @@ class LockFile:
     def is_semantically_equivalent(self, other: LockFile) -> bool:
         """Return True if *other* has the same deps, MCP/LSP servers, and configs.
 
-        Ignores ``generated_at`` and ``apm_version`` so that a no-change
-        install does not dirty the lockfile.
+        Ignores ``apm_version`` so that a no-change install does not dirty
+        the lockfile.
         """
         if self.lockfile_version != other.lockfile_version:
             return False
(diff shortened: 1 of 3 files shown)


danielmeppiel:
Thanks @aryansk for addressing the timestamp conflicts and implementing the removal that the earlier triage explicitly requested. I want to acknowledge that guidance: we have since chosen a narrower migration policy, rather than asking you to keep refreshing this implementation.

I'm closing this in favor of the contract landed in #2616 on main:

- New lockfiles omit generated_at.
- Existing timestamp-bearing lockfiles preserve the value on semantic no-ops.
- Substantive changes to those legacy files still refresh the timestamp.

Your PR would go further by omitting the field from all subsequent writes. That would remove a real remaining source of timestamp-only conflicts in legacy repositories, so this is not an exact duplicate and the difference is not merely cosmetic.

Our decision is to keep repository-controlled migration rather than automatically remove the field everywhere. With the new writer, a repository can delete its generated_at line once and later writes will not restore it. Universal removal is not needed for dependency resolution, and we are declining that additional automatic migration for now. This is a policy choice, not a claim that every older reader requires the field.

The #2616 behavior is on main; this comment does not imply it is already in the published v0.29.0 release. Thank you for the contribution.

PR #2627, opened 2026-08-19
fix(deps): invalidate the manifest parse cache when apm.yml is rewritten, so re-downloads re-stamp

McNultyyy:
> Originally stacked on #2620; now **rebased onto `main`** (post-merge) as a single commit.

## Summary

Fixes a pre-existing bug surfaced while verifying #2620 end-to-end: `APMPackage.from_apm_yml` memoizes per path for the process lifetime, but marketplace-plugin flows rewrite `apm.yml` twice within a run (synthesis at `0.0.0`, then the short-SHA stamp). On a **same-process re-download** (e.g. integrate's content-hash mismatch pre-check rmtrees and re-acquires), re-synthesis writes a fresh `0.0.0` manifest but `from_apm_yml` returns the stale stamped instance — `stamp_plugin_version`'s `version == "0.0.0"` guard skips, and the tree is left **unstamped** on disk, hashing to a value no lockfile ever recorded.

Observable on current main (#2620 included): `apm install --update` against a stale lockfile records the unstamped-tree hash, and every subsequent fresh install hard-fails the supply-chain check with a hash matching neither domain. (Empirically reproduced with `anthropics/skills`.)

## Fix

- `invalidate_apm_yml_cache_entry(path)`, called from every `apm.yml` rewrite site inside package trees (synthesis, `stamp_plugin_version`, virtual-file manifests).
- Cache mutations hold a module lock — the invalidation iterates the dict while parallel pre-download/resolver worker threads insert into it (`RuntimeError: dictionary changed size during iteration`, reproduced; stress test included).
- A generation counter makes invalidation a barrier for in-flight parses: a parse that read pre-rewrite bytes skips its cache insert (the caching lives in `from_apm_yml`; `from_mapping` stays cache-free).
- `download_package`'s persistent-cache fast path already routes through `stamp_plugin_version` since #2654 — with the invalidation now inside `stamp_plugin_version`, that shared implementation also becomes safe on same-process re-downloads.

With this, `apm install --update` becomes a safe migration path for pre-#2620 Windows lockfiles. Full auto-migration is stacked PR #2628.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2628, opened 2026-08-19
feat(install): auto-migrate legacy CRLF-domain lockfiles and warm trees (apm#2619)

McNultyyy:
> **Stacked on the manifest-cache stamping fix #2627** — review the last commit only (`e6bc809e`). #2620 is merged; this branch is rebased on `main` past it.

## Summary

Removes the manual migration step #2620 leaves behind. Lockfiles and warm `apm_modules` trees recorded by a pre-fix APM on Windows carry the CRLF-domain `content_hash` of the APM-authored synthetic files; without this, upgrading surfaces a misleading "supply-chain attack" hard failure until the user deletes `apm_modules/` + `apm.lock.yaml` and regenerates.

## How it migrates

- **Fresh downloads**: if the locked hash equals the fresh tree re-hashed with *only* the APM-authored files (synthetic `apm.yml`, inline `hooks.json`) CRLF-expanded (`compute_package_hash_with_overrides` — no tree mutation), the difference is provably the benign legacy line-ending domain: accept, warn once, re-record the platform-independent hash. Genuine tampering still hard-fails (negative-control test included).
- **Cached installs**: legacy CRLF bytes in APM-authored files are rewritten to LF *before* the hash is re-recorded, so stale CRLF-domain hashes stop propagating into regenerated lockfiles.

Safety properties: upstream (git-materialized) content is never touched; `hooks.json` counts as APM-authored only when the plugin manifest declares hooks inline (a file-path declaration is a byte-copy of upstream content); authored paths behind symlinked directories are skipped, mirroring `compute_package_hash`'s enumeration semantics.

Verified end-to-end against the real `anthropics/skills` repro: a pre-fix Windows lockfile installs `--frozen` with a one-time warning and converges to `sha256:606a7fcb…`; a hand-CRLF-ified warm tree self-heals on the next install; reruns are silent.

Depends on the stamping-cache fix: the mismatch-triggered re-download must re-stamp for the fresh hash to be canonical.

## Panel follow-up (#2620): frozen replay coverage

Adds the requested hermetic marketplace-plugin lockfile write/read + `install --frozen` replay tests: the positive replay re-downloads through the real synthesize+stamp chain in the same process **without** clearing the manifest parse cache (a direct regression guard for #2627 — removing its invalidation makes the test fail), and a tampered-lockfile negative control keeps the frozen supply-chain gate closed.

## Spec conformance

The relaxed check is APM's internal raw-bytes package-tree `content_hash` (`docs/reference/lockfile-spec.md`); OpenAPM v0.1 reserves canonical-tree `content_hash` semantics for v0.2 (§5.6 editorial note) and anchors normative integrity on `tree_sha256` / `resolved_hash` / `deployed_file_hashes`, none of which change here. Waiver below.

apm-spec-waiver: one-time migration shim for pre-fix Windows lockfiles -- accepts the provably-benign legacy CRLF-domain content_hash of APM-authored synthetic files once and re-records the canonical hash; content_hash canonical-tree semantics are reserved for OpenAPM v0.2 (v0.1 spec section 5.6 editorial note), so no v0.1 normative contract changes

### DIFF
--- a/.github/workflows/crlf-invariance.yml
+++ b/.github/workflows/crlf-invariance.yml
@@ -23,6 +23,7 @@ on:
       - 'src/apm_cli/deps/package_validator.py'
       - 'src/apm_cli/deps/github_downloader.py'
       - 'src/apm_cli/models/validation.py'
+      - 'src/apm_cli/install/legacy_crlf.py'
       - 'scripts/crlf_invariance_probe.py'
       - '.github/workflows/crlf-invariance.yml'
   push:
@@ -35,6 +36,7 @@ on:
       - 'src/apm_cli/deps/package_validator.py'
       - 'src/apm_cli/deps/github_downloader.py'
       - 'src/apm_cli/models/validation.py'
+      - 'src/apm_cli/install/legacy_crlf.py'
       - 'scripts/crlf_invariance_probe.py'
       - '.github/workflows/crlf-invariance.yml'
   # Merge queues do not support path filtering; run unconditionally there so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Lockfiles and warm `apm_modules` trees recorded by a pre-#2620 APM on
+  Windows are now migrated automatically: when a locked hash matches the
+  fresh tree re-hashed with only APM-authored files (synthetic `apm.yml`,
+  inline-hooks `hooks.json`) expanded to the legacy CRLF domain, the install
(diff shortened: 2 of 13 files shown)


danielmeppiel:
Thanks for addressing the legacy CRLF migration problem. The maintainer decision is that frozen must remain frozen: replay of the recorded state, not an automatic migration or re-recording path. We are not accepting the proposed automatic/frozen migration scope, so I'm closing this PR.

This is a contract decision, not a claim that your bounded legacy detection or tampering negative controls are invalid. #2627's independent manifest-cache repair and #2971's approved checkout-byte/hash portability fix remain separate. An explicit, auditable non-frozen migration could be discussed if affected-user demand and an owner establish the need, but it is not selected now.

PR #2635, opened 2026-08-20
fix(plugins): warn when declared skills entries collide on a skill name (closes #2629)

### LINKED ISSUE #2629: [BUG] Two declared plugin skills entries with the same skill name silently overwrite during normalization
## Summary

When a plugin manifest declares two `skills` entries that each hold a skill of the same name, normalization merges both into `.apm/skills/<name>/` and the last entry silently wins. Nothing warns, and nothing records that a skill was dropped.

## Repro

`.claude-plugin/plugin.json`:

```json
{"name": "collide", "version": "1.0.0", "skills": ["./a/", "./b/"]}
```

```
a/tdd/SKILL.md   ->  "# from a"
b/tdd/SKILL.md   ->  "# from b"
```

After `_map_plugin_artifacts` (`src/apm_cli/deps/plugin_parser.py`):

```
.apm/skills/tdd/SKILL.md   ==  "# from b"
```

One `tdd` remains. The copy from `./a/` is gone, with no diagnostic on any verbosity level.

## Why it surfaces now

Before #2536 a declared list entry was copied under its own name (`.apm/skills/a/tdd/`), so two containers could not land on the same path. That was #2530: nothing at that depth was selectable with `--skill` or deployable. Now that a declared container merges to the depth the rest of the pipeline reads, two containers contributing the same skill name resolve to one path, and `shutil.copytree(..., dirs_exist_ok=True)` settles it by overwriting.

So the overwrite is not a regression in outcome -- previously both copies were unreachable -- but it is a new silent failure mode on a path that now works.

## Suggested behavior

Detect the collision while merging and surface it: name the skill and both declared entries that provide it. `SkillIntegrator._promote_sub_skills` already tracks ownership at deploy time via `owned_by`; normalization has no equivalent. Failing closed is also defensible for a single package, whose own manifest should be internally unambiguous -- worth a decision rather than an assumption.

## Scope

- Within one package. Collisions between skills from *different* packages at deploy time are #1120.
- Raised as a follow-up in the PR #2536 review panel (supply-chain lens); out of scope for that PR, which only restores the correct container depth.

edenfunf:
Closes #2629.

**Stacked on #2536** — the first two commits are that PR (the collision this fixes only exists once declared containers merge to one level); only `b89b4da` is new here. Rebases to a one-commit diff once #2536 lands.

## Problem

When a plugin manifest declares two `skills` entries that each hold a skill of the same name:

```json
{"name": "collide", "version": "1.0.0", "skills": ["./a/", "./b/"]}
```

```
a/tdd/SKILL.md   ->  "# from a"
b/tdd/SKILL.md   ->  "# from b"
```

both entries merge onto `.apm/skills/tdd/`, and `shutil.copytree(..., dirs_exist_ok=True)` settles the collision by overwriting: the last declared entry wins and the copy from `./a/` is gone, with no diagnostic on any verbosity level.

## Fix

Normalization now tracks, per run, which declared entry provided each skill name (the same depth deployment and `--skill` read). When a second entry lands an already-provided name, it surfaces one warning naming the skill and both entries:

```
[!] Plugin skills entries 'a' and 'b' both provide skill 'tdd'; only the copy
    from the last declared entry survives. Rename one skill or remove the
    duplicate entry.
```

Decisions, both following the deploy-time precedent in `SkillIntegrator._promote_sub_skills`:

- **Warn, don't block.** The overwrite outcome is unchanged (last declared entry still wins); deploy-time collisions warn rather than fail, and failing install here is a larger behavioral call than a diagnostic follow-up should make unilaterally. Happy to switch to fail-closed if that's preferred for a single package's own manifest.
- **Same-source overlap stays quiet.** `"skills": ["./skills/", "./skills/tdd"]` resolves the name to the same source directory — both copies are byte-identical, so nothing is lost (mirrors the content-identical skip at deploy time).

## Tests

Four new unit tests in `tests/unit/test_plugin_parser.py`: the issue's repro (warning fired, last-wins outcome asserted), a container child colliding with a direct skill entry (the two manifest shapes land on the same path), distinct names across entries staying quiet, and same-source overlap staying quiet. The two warning tests fail without the fix.


danielmeppiel:
Thanks @edenfunf for investigating this and proposing the diagnostics. The parent #2536 was superseded by merged #2540, and the [follow-up on #2629](https://github.com/microsoft/apm/issues/2629#issuecomment-5388477701) verified the fail-closed behavior in both declaration orders.

We will keep excluding ambiguous skill names from selection/deployment rather than introduce warn-and-last-wins behavior, so I'm closing this PR as superseded. That is the relevant guarantee, not a claim that all intermediate staging bytes are identical. Your investigation remains useful context; there is no need to refresh this branch.

PR #2214, opened 2026-07-15
fix(audit): exempt is_virtual packages from config-consistency manifest check

hugoguitton-lucca:
# fix(audit): accept valid manifestless virtual skills

## TL;DR

`apm audit --ci` no longer reports a missing-manifest inconsistency for a valid
virtual `claude_skill` subpath that contains `SKILL.md` and no `apm.yml`.
The exemption now requires both canonical virtual-subdirectory identity and the
canonical on-disk Claude-skill shape, so malformed packages and real MCP drift
still fail.

> [!NOTE]
> Fixes #2213. Closes #2207. The regression test is RED on pre-fix `main`
> and when the exemption is removed.

## Problem (WHY)

- [x] A successful install could leave CI permanently red:
  ["A successful `apm install` cannot repair the failure because no package
  manifest exists by design."](https://github.com/microsoft/apm/issues/2207)
- [x] The original one-bit exemption trusted `is_virtual` without proving the
  installed package still had a supported manifestless shape.
- [x] The previous unit-only proof did not cross install, frozen replay, and
  the public audit command. This matters because
  ["Grounding outputs in deterministic tool execution transforms probabilistic
  generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/)
- [!] The baseline-check reference said every missing locked-package manifest
  fails, omitting both supported manifestless package shapes.

Why these matter: audit is a governance boundary. It must accept supported
package shapes without creating a lockfile bit that can hide genuine package or
MCP configuration drift.

## Approach (WHAT)

| # | Fix | Principle | Source |
|---:|---|---|---|
| 1 | Centralize the missing-manifest decision in `_allows_missing_manifest`. | “Favor small, chainable primitives over monolithic frameworks.” | [PROSE](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) |
| 2 | Require virtual-subdirectory identity **and** detected `CLAUDE_SKILL` shape. | “Add what the agent lacks, omit what it knows.” | [Agent Skills](https://agentskills.io/skill-creation/best-practices) |
| 3 | Exercise real local Git install, module deletion, frozen replay, and public audit. | “Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action.” | [PROSE](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/) |
[...]

PR #2643, opened 2026-08-20
fix(audit): allow local Claude-skill deps to omit apm.yml

### LINKED ISSUE #2611: [BUG] `apm audit --ci` rejects local `SKILL.md` dependencies in 0.28.0
## Summary

APM 0.28.0 installs, locks, deploys, and ordinarily audits a local dependency
whose root contains `SKILL.md` without `apm.yml`. However, `apm audit --ci`
exits 1 on the unchanged project. Its `config-consistency` check requests an
`apm.yml` that the valid skill bundle never contained and that another
`apm install` does not create.

APM's default documented CI workflow runs `apm install` before
`apm audit --ci`, so the lockfile and deployed files are present before the
gate reads them. This reproducer follows that sequence: installation succeeds,
ordinary audit passes, and the unchanged installed state is then rejected only
by the CI gate.

This prevents projects that intentionally model a raw local skill directory as
an independent APM dependency from using the CI audit gate. Project-owned
skills included directly from the root package's `.apm/skills/` directory are
an effective workaround. The error is reported as an MCP configuration
inconsistency even though the reproducer declares no MCP dependencies.

## Reproduction

Validated on Darwin 25.5.0 arm64 with uv 0.12.1 and the
PyPI `apm-cli==0.28.0` distribution

Prerequisites:

- a POSIX `sh` environment;
- [`uv`](https://docs.astral.sh/uv/) and network access to PyPI; and
- no credentials or pre-existing APM project.

The script creates and retains a fresh temporary project. It pins APM 0.28.0,
asserts that installation and ordinary audit succeed, and exits zero only when
the CI audit fails solely with the expected `config-consistency` finding.

```sh
#!/bin/sh
set -eu

WORKDIR=$(mktemp -d "${TMPDIR:-/tmp}/apm-local-skill-audit-XXXXXX")
PROJECT="$WORKDIR/project"
mkdir -p "$PROJECT/skills/example"
cd "$PROJECT"

uv venv --python 3.13 "$WORKDIR/venv"
UV_NO_PROGRESS=1 uv pip install --quiet \
    --python "$WORKDIR/venv/bin/python" \
    apm-cli==0.28.0
APM="$WORKDIR/venv/bin/apm"
PYTHON="$WORKDIR/venv/bin/python"

cat >apm.yml <<'YAML'
name: apm-local-skill-audit-repro
version: 0.1.0
description: Reproduce CI audit behavior for a local skill
targets:
  - agent-skills
dependencies:
  apm:
    - path: ./skills/example
  mcp: []
includes: auto
scripts: {}
YAML

cat >skills/example/SKILL.md <<'MARKDOWN'
---
name: example
description: A minimal local skill used to reproduce an APM audit issue.
---

# Example

Return the word `example`.
MARKDOWN

"$APM" --version
"$APM" install --no-policy >install.log 2>&1
"$APM" audit --no-policy >ordinary-audit.log 2>&1

if "$APM" audit --ci --no-policy --format json \
    >ci-audit.json 2>ci-audit.stderr; then
    echo "apm audit --ci unexpectedly succeeded" >&2
    exit 1
fi

"$PYTHON" - "$PROJECT/ci-audit.json" <<'PYTHON'
import json
import pathlib
import sys

result = json.loads(pathlib.Path(sys.argv[1]).read_text())
failed = [check for check in result["checks"] if not check["passed"]]
if len(failed) != 1 or failed[0]["name"] != "config-consistency":
    raise SystemExit(f"unexpected failed checks: {failed!r}")
details = failed[0].get("details", [])
if not any(
    "package manifest not found" in detail
    and "skills/example/apm.yml" in detail
    for detail in details
):
    raise SystemExit(f"expected missing-manifest detail not found: {details!r}")
print(json.dumps(failed[0], indent=2))
PYTHON

echo "Reproduced; workspace retained at: $WORKDIR"
```

## Actual behavior

`apm install --no-policy` exits 0, writes a lock entry with
`package_type: claude_skill`, and deploys the skill. Ordinary
`apm audit --no-policy` also exits 0. The CI audit then exits 1 with this sole
failed check:

```json
{
  "name": "config-consistency",
  "passed": false,
  "message": "1 MCP config inconsistenc(ies) -- run 'apm install' to reconcile",
  "details": [
    "./skills/example: package manifest not found at .../skills/example/apm.yml; re-run 'apm install' to restore it"
  ]
}
```

Re-running `apm install` does not add the requested manifest.

## Expected behavior

`apm audit --ci` should exit 0 after a successful install when the local
dependency is an unchanged, valid root `SKILL.md` package. It should not
require the optional `apm.yml` or suggest that installation can restore a file
that was never part of the dependency.

## References

- [APM 0.28.0 release](https://github.com/microsoft/apm/releases/tag/v0.28.0)
- [Enforce in CI: the default action installs before running the audit gate](https://microsoft.github.io/apm/enterprise/enforce-in-ci/#recipe-minimal-github-actions-gate)
- [Package-types documentation: a root `SKILL.md` is a supported skill bundle and `apm.yml` is optional](https://microsoft.github.io/apm/reference/package-types/#skill-bundle-skillmd-at-root)
- [`config-consistency` missing-manifest handling in v0.28.0](https://github.com/microsoft/apm/blob/v0.28.0/src/apm_cli/integration/mcp_config_view.py#L209-L300)
- [Issue #2207 and its fix in PR #2214: the corresponding false positive for virtual subdirectory skills](https://github.com/microsoft/apm/issues/2207)
- [Issue #2443: the related open report for repo-root Git skills on the frozen-install path](https://github.com/microsoft/apm/issues/2443)

lukiod:
Fixes #2611.

## Description

`_allows_missing_manifest()` in `mcp_config_view.py` already waives a missing
`apm.yml` for virtual subdirectory Claude-skill packages (the fix in #2214),
but a local `path:` dependency whose `package_type` is `claude_skill` fell
straight to the unconditional `return False`, so `apm audit --ci` rejected a
valid root-`SKILL.md` local skill on every run, even right after a clean
`apm install`. The error message ("re-run `apm install` to restore it") is
actively misleading, since `apm install` never produces the manifest it asks
for — the dependency legitimately never had one.

Fixed by letting local dependencies fall through to the same on-disk shape
check virtual subdirectory packages already use. A local directory is part
of the project's own source tree, not a download target, so it always
exists and always gets the strict `detect_package_type()` verification
(same defense against a mislabeled or malformed package the virtual path
already has), rather than the cold-cache lockfile-trust fallback that exists
specifically for virtual packages whose directory may not be materialized
yet in a setup-only CI job.

### Checklists

Documentation
- [ ] Updated for new or modified user-facing features or behaviors
- [x] No user-facing change (bug fix restores documented behavior — the
      package-types reference already says `apm.yml` is optional for a root
      `SKILL.md` package)

Testing
- [x] Added or modified tests to cover new code paths
      (`test_manifestless_local_claude_skill_waived` — the exact reported
      scenario, reproduced with the issue's own repro before the fix and
      confirmed passing after; `test_manifestless_local_package_without_skill_shape_records_problem`
      — a companion regression guard confirming a local dependency that
      claims `claude_skill` but lacks a real `SKILL.md` still gets flagged,
      matching the existing virtual-package guard)
- [ ] Not required

Performance
- [ ] Tests ran and results are added in the PR description
- [ ] Issue filed with a link in the PR description
- [x] Not required


danielmeppiel:

> Allowing shape-validated local Claude skills removes a first-run audit false positive while preserving strict manifest checks for ordinary or absent paths.

cc @lukiod @danielmeppiel @sergio-sisternes-epam -- a fresh advisory pass is ready for your review.

The panel agrees that this is a bounded, user-trust-positive correction: a local dependency explicitly marked as a Claude skill and containing a root `SKILL.md` should not be rejected solely because it lacks `apm.yml`. Delegating classification through `detect_package_type` keeps the waiver owned by the canonical detector rather than creating a second interpretation of package shape.

Fold three small safeguards in this PR: test that a missing local Claude-skill directory remains strict, add a lifecycle/CLI regression for `apm audit --ci`, and document the narrow local root-`SKILL.md` exception alongside the absent-path boundary. This preserves secure failure behavior and makes the promise discoverable without expanding the waiver to normal local dependencies.

**Aligned with:** portable by manifest (the exception is deliberately narrow: only an explicitly classified local Claude skill with a root SKILL.md may omit apm.yml); secure by default (ordinary local dependencies and nonexistent local paths remain manifest-strict); pragmatic as npm (valid local skills should pass audit without an unnecessary manifest).

**Growth signal.** Removing this first-run audit false positive makes local Claude-skill adoption feel reliable without changing APM's public conversion surface.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 1 | 0 | The canonical MCP manifest-waiver owner is extended correctly; add an absent local Claude-skill regression case. |
| CLI Logging Expert | 0 | 0 | 0 | The valid local Claude-skill waiver removes a false audit problem while malformed paths retain actionable diagnostics. |
| DevX UX Expert | 0 | 1 | 0 | Add a CLI-level regression test for the restored valid-local-skill audit path. |
| Supply Chain Security Expert | 0 | 1 | 0 | Local Claude skills are shape-validated and missing local paths fail closed; add a regression test for that boundary. |
| OSS Growth Hacker | 0 | 0 | 0 | Fix removes a first-run audit false positive for local Claude skills without changing conversion surfaces. |
| Doc Writer | 0 | 1 | 0 | The config-consistency reference omits the newly valid local root-SKILL.md exception. |
| Test Coverage Expert | 0 | 1 | 0 | Local manifestless skill audit fix has unit coverage but lacks a real CLI regression trap. |

### Top 3 follow-ups

1. **[Python Architect, Supply Chain Security Expert]** Add a regression test for an explicitly local Claude-skill dependency whose directory is absent and assert manifest-not-found behavior. -- This protects the fail-closed boundary so local lock metadata cannot grant a cold-cache waiver.
2. **[DevX UX Expert, Test Coverage Expert]** Add a lifecycle regression covering a local root-SKILL.md dependency without apm.yml through normal and CI audit paths. -- This proves the user-facing `apm audit --ci` promise rather than only the internal config-view decision.
3. **[Doc Writer]** Document the narrow manifestless local Claude-skill exception and clarify that absent local paths remain strict. -- The config-consistency reference must accurately describe the newly valid shape and its security boundary.

### Architecture

```mermaid
classDiagram
    class CurrentMcpConfigView {
      <<ValueObject>>
      +derive() CurrentMcpConfigView
      +problems tuple
    }
    class McpConfigViewModule {
      <<CanonicalAuthority>>
      +_allows_missing_manifest()
    }
    class LockedDependency {
      <<ValueObject>>
      +source str
      +package_type str
      +local_path str
    }
    CurrentMcpConfigView ..> McpConfigViewModule : delegates
    McpConfigViewModule ..> LockedDependency : reads
```

```mermaid
flowchart TD
    A["apm audit --ci"] --> B["CurrentMcpConfigView.derive()"]
    B --> C{"manifest exists?"}
    C -->|no| D["_allows_missing_manifest()"]
    D -->|local present| E["detect_package_type()"]
    D -->|local missing| F["manifest problem"]
    E -->|root SKILL.md plus claude_skill| G["audit succeeds"]
    E -->|other shape| F
```

### Recommendation

Fold the missing-directory test, CLI lifecycle regression, and bounded documentation update in this PR, then ship the focused audit correction. The highest-signal safeguard is the missing-local-directory test because it protects the fail-closed local-path boundary.

---

<details>

#### Python Architect
- **[recommended]** Cover the missing local Claude-skill directory branch explicitly at `tests/unit/integration/test_mcp_config_view.py:437`.

#### CLI Logging Expert
No findings.

#### DevX UX Expert
- **[recommended]** Cover valid local Claude skill through `apm audit --ci` at `tests/unit/test_audit_ci_command.py`.

- **[recommended]** Add coverage for a missing local Claude-skill directory at `tests/unit/integration/test_mcp_config_view.py:437`.

#### OSS Growth Hacker
No findings.

#### Auth Expert -- inactive
No auth-relevant surface touched.

#### Doc Writer
- **[recommended]** Document manifestless local Claude skills in config-consistency at `docs/src/content/docs/reference/baseline-checks.md:128`.

#### Test Coverage Expert
- **[recommended]** Add an end-to-end local root-SKILL audit regression test at `tests/integration/test_virtual_skill_audit_e2e.py`.

No performance-relevant surface touched.
</details>

lukiod:
Addressed the review panel's three follow-ups and the two inline Copilot comments.

Since your own commits already landed the missing-directory regression test (`test_manifestless_missing_local_claude_skill_records_problem`), the doc update, and a thorough real-binary e2e test (`test_manifestless_local_skill_installs_and_audits_without_state_drift`) covering both normal and `--ci` audit paths, I rebased on top of those rather than duplicating them, and only added what wasn't there yet:

- Fixed the two inline comments: the docstring/comment wording implied a local `path:` dependency is always in-tree and always present ("checked in, always present" / "part of the project's own source tree"). That's inaccurate — `resolve_local_dep_dir()` allows a local dependency to point anywhere, including outside the repo (e.g. `../sibling`), with no existence guarantee. Reworded both to say the path is resolved directly from the filesystem, not materialised into `apm_modules/`.

Verified against the current tree, not assumed: `uv run pytest tests/unit/ -q` → `20168 passed, 3 skipped, 21 xfailed` (matches your baseline, no new failures). `uv run ruff check` on both touched files → `All checks passed!`.

PR #2661, opened 2026-08-22
fix(pack): preserve marketplace metadata outcomes

### LINKED ISSUE #2524: [BUG] apm pack silently drops description/version for remote packages when metadata fetch fails (exit 0, no warning); --check-clean then passes
**Describe the bug**

When the remote metadata fetch fails for any reason other than an HTTP 404 on an internal repository, `apm pack` drops `description` and `version` from every remote `git-subdir` entry, prints `[*] Built marketplace.json`, writes nothing to stderr, and exits 0. A network outage, a proxy that refuses connections, an expired token, or a rate-limited runner all produce a truncated `marketplace.json` that looks like a successful build. Consumers then reject the affected plugins, because `description` is required once emitted.

This is the general case behind #1847, which fixed one specific trigger (raw.githubusercontent.com returning 404 for INTERNAL repos) by adding a REST fallback. The reporting gap it exposed is still open: every other failure mode stays silent.

Two code sites produce the silence:

- `marketplace/builder.py::_prefetch_metadata` collects the thread-pool results under `except Exception: pass`, so a per-package fetch failure leaves no trace at default verbosity. The same method returns early under `--offline` without recording that remote packages were skipped.
- `marketplace/output_mappers.py::_apply_field_with_precedence` omits the field when neither the entry value nor the fetched value is present, and emits no diagnostic. The surrounding code does emit `BuildDiagnostic` warnings for smaller problems, such as a marketplace name that is not kebab-case.

The diagnosis exists but is unreachable in practice: `Could not fetch remote metadata for <name>` is logged at `DEBUG`. Running `apm pack -v` shows nothing; only `APM_LOG_LEVEL=DEBUG` surfaces it.

The second consequence is worse than the first. `--check-clean` regenerates the index and diffs it against the file on disk, so when the network is down on both sides, a truncated `marketplace.json` passes the release gate:

| `marketplace.json` on disk | Network | `apm pack --check-clean --dry-run` |
| --- | --- | --- |
| truncated | blocked | **exit 0 — gate passes** |
| truncated | available | exit 4 — drift detected |

A CI job that runs the gate on a runner with a degraded network therefore certifies a broken index as clean.

**To Reproduce**

1. Create a project with an `apm.yml` whose marketplace declares one remote `git-subdir` package. The reference below is a public repository pinned to a full SHA, so no token and no ref resolution are needed:

   ```yaml
   name: repro
   version: 1.0.0
   description: Minimal marketplace with one remote git-subdir package

   targets:
     - claude

   marketplace:
     owner:
       name: repro-owner
     outputs:
       claude: {}
     versioning:
       strategy: per_package
     packages:
       - name: adapt-nifi-flows-to-2-x
         source: Netcracker/qubership-nifi
         subdir: agent-packages/adapt-nifi-flows-to-2-x
         ref: 09c4bf708d787eba45046904ab5b40b9ac597c6b
   ```

2. Run `apm pack` with the network available. The entry gets `"description": "Apply NiFi 1.x to 2.x upgrade advisor recommendations to exported flow JSON files."` and `"version": "1.2.0"`, both inherited from the upstream package's own `apm.yml`.

3. In a fresh directory with the same `apm.yml`, run `apm pack` with outbound HTTPS pointed at a closed port. Note that `--offline` is *not* passed:

   ```bash
   env https_proxy=http://127.0.0.1:9 HTTPS_PROXY=http://127.0.0.1:9 ALL_PROXY=http://127.0.0.1:9 apm pack
   ```

4. Observe the exit code and the output:

   ```text
   [*] Built marketplace.json [claude] (1 package(s)) -> .../.claude-plugin/marketplace.json
   exit=0
   stderr: 0 bytes
   ```

   The written entry has lost both fields:

   ```json
   {
     "name": "adapt-nifi-flows-to-2-x",
     "source": {
       "source": "git-subdir",
       "url": "Netcracker/qubership-nifi",
       "path": "agent-packages/adapt-nifi-flows-to-2-x",
       "ref": "09c4bf708d787eba45046904ab5b40b9ac597c6b",
       "sha": "09c4bf708d787eba45046904ab5b40b9ac597c6b",
       "tag_pattern": "v{version}"
     }
   }
   ```

5. Confirm the release gate accepts it. Against the truncated file from step 4:

   ```bash
   env https_proxy=http://127.0.0.1:9 HTTPS_PROXY=http://127.0.0.1:9 ALL_PROXY=http://127.0.0.1:9 \
     apm pack --check-clean --dry-run; echo "exit=$?"   # exit=0
   apm pack --check-clean --dry-run; echo "exit=$?"      # exit=4
   ```

6. Confirm the information is DEBUG-only. `apm pack -v` mentions metadata zero times; `APM_LOG_LEVEL=DEBUG apm pack` prints `DEBUG apm_cli.marketplace.builder Could not fetch remote metadata for adapt-nifi-flows-to-2-x`.

**Expected behavior**

A failed metadata fetch should be visible at default verbosity, and should be able to fail the build.

1. Report it. Every swallowed fetch failure, and every remote package skipped because of `--offline`, should surface as a `warning`-level `BuildDiagnostic` naming the package and the cause. The build summary should not read `[*] Built marketplace.json (1 package(s))` when that package lost half its metadata.
2. Let CI reject it. A flag along the lines of `--strict-metadata` should promote those warnings to a hard failure. Exit codes 0 through 4 are taken by success, build error, schema error, `--check-versions`, and `--check-clean`, so this needs a new one.
3. Keep `--check-clean` honest. The gate currently compares two artifacts that can be wrong in the same way. Failing the run when the regeneration side degraded would close that hole regardless of what the file on disk contains.

Point 2 is what we need most: a marketplace publisher wants one command that either produces a complete index or fails.

- OS: macOS 15 (Darwin 25.5.0, arm64)
- Python Version: 3.12
- APM Version: 0.28.0 (installed from PyPI via `uvx --from apm-cli==0.28.0`); also reproduced on the 0.26.0 native binary and observed in production on 0.24.1
- VSCode Version (if relevant): N/A

**Logs**

Default verbosity, network blocked, no `--offline`:

```text
[!] No 'license:' field in apm.yml; the SBOM will record NOASSERTION for this package. ...
[*] Built marketplace.json [claude] (1 package(s)) -> .../.claude-plugin/marketplace.json
[i] Marketplace artifacts ready:
```

Nothing on stderr. The same run under `APM_LOG_LEVEL=DEBUG`:

```text
DEBUG apm_cli.marketplace.builder Resolved GitHub token for metadata fetch (source=gh-auth-token)
DEBUG apm_cli.marketplace.builder Could not fetch remote metadata for adapt-nifi-flows-to-2-x
  File ".../apm_cli/marketplace/builder.py", line 1026, in _fetch_remote_metadata
```

**Additional context**

This has hit us twice in production on [Netcracker/qubership-ai-packages](https://github.com/Netcracker/qubership-ai-packages), a marketplace with 27 packages of which 10 are remote `git-subdir` entries. Both times a contributor ran `apm pack` without working network access, committed the result, and all 10 remote packages lost `description` and `version` at once. CI ran `apm pack --check-versions --check-clean --dry-run` on both commits and passed them. The breakage surfaced only when consumers started rejecting the packages, and was repaired by restoring the fields by hand.

Related: #1847 (the 404-on-INTERNAL-repos trigger, fixed) and #1725 (local virtual-subdirectory packages not enriched).

danielmeppiel:
Closes #2524.

Preserves per-package metadata enrichment outcomes through pack output, JSON, strict mode, and the check-clean gate. Failed or offline remote metadata now warns by default; `--strict-metadata` exits 5 before writing; `--check-clean` exits 4 without comparing uncertifiable regeneration.

Validation: `PYTHONPATH=src python3 -m pytest tests/unit/marketplace/test_metadata_enrichment.py tests/integration/test_remote_marketplace_metadata_outage_lifecycle.py -q -x` (3 passed, 1 skipped); lifecycle test collected. Ruff could not run because the local Python environment has no ruff module and uv cannot fetch build dependencies.


danielmeppiel:

> Preserving marketplace metadata outcomes is valuable, but strict multi-output failure semantics must be atomic and phase-aligned before release.

The panel converges on a trust issue: strict marketplace metadata validation currently occurs during the per-output write loop, so a later failure can leave an earlier artifact or bundle written before exit 5. Packaging must not report a failed strict outcome after mutating only part of the requested output set. Preflight validation before every write, backed by an ordered multi-output no-mutation and idempotence proof, is the right correction.

DevX identified a related contract mismatch: JSON can record a successful metadata enrichment before an independently fetched `--check-clean` gate later fails with exit 4. The emitted result must describe the completed command phase, not an earlier intermediate state. Documentation should distinguish metadata unavailability from ordinary cleanliness drift and make dry-run ordering explicit. Architecture is otherwise appropriate; the logging polish can follow separately.

**Aligned with:** portable manifest packaging through consistent output state; secure defaults through fail-before-write strict behavior; pragmatic automation through phase-aligned JSON.

**Growth signal.** Reliable failure semantics let CI and marketplace users trust that APM's machine-readable result describes the command that actually completed.

### Panel summary

| Persona | High | R | N | Takeaway |
|---|---|---|---|---|
| Python Architect | 0 | 0 | 1 | Outcome ownership is appropriately scoped. |
| CLI Logging | 0 | 1 | 0 | Builder warnings currently duplicate the CLI status prefix. |
| DevX UX | 1 | 1 | 0 | Final JSON outcome can disagree with clean-check failure. |
| Supply Chain | 1 | 0 | 0 | Strict mode can leave partial artifacts. |
| OSS Growth | 0 | 1 | 0 | Outage and drift recovery need distinct guidance. |
| Auth | 1 | 0 | 0 | Inspect fallback behavior before expanding scope. |
| Docs | 0 | 1 | 0 | Check-clean docs need dry-run and outage clarity. |
| Test Coverage | 1 | 1 | 0 | Add atomic ordered multi-output lifecycle proof. |
| Performance | 0 | 1 | 0 | Verify proposed metadata lookup optimization first. |

### Top 3 follow-ups

1. **[Supply Chain Security Expert and Test Coverage Expert]** Preflight strict metadata before every output write and prove an ordered multi-output failure leaves every artifact and bundle unchanged across a retry. This prevents partial publication state.
2. **[DevX UX Expert]** Align `metadata_enrichment` in JSON with the final command phase so an exit-4 clean-check failure cannot claim certification. Machine consumers need one coherent outcome.
3. **[Doc Writer]** Separate metadata outage remediation from ordinary drift and document dry-run validation before artifact generation. This gives CI users an actionable recovery path.

### Recommendation

Correct the atomic strict-metadata and phase-aligned JSON failure semantics, with the multi-output no-mutation and idempotence regression proof, before shipping. Then fold the focused documentation clarification in the same change.

---

<details>

- **[recommended] CLI Logging:** remove presentation prefixes from builder warning data.
- **[high priority] DevX UX:** surface the clean-check metadata result in the JSON envelope.
- **[high priority] Supply Chain and Test Coverage:** strict metadata must be validated before all writes, including bundle output.
- **[recommended] Docs and Growth:** metadata outage is not drift; validation should use `--dry-run`.
- **[review before scope expansion] Auth and Performance:** verify credential fallback and lookup complexity against the exact implementation.
</details>

<sub>This panel is advisory. It does not decide merge readiness. Re-apply the `panel-review` label after addressing feedback to re-run.</sub>

PR #2665, opened 2026-08-22
fix(audit): preserve user hooks during drift checks

### LINKED ISSUE #2641: apm audit always reports drift on .claude/settings.json if the project has its own hooks
**Version:** apm 0.26.0 (`64a7fb5`)

If my project already has hooks in `.claude/settings.json` and I install a plugin
that also has hooks, `apm audit --ci` fails from then on. It reports
`.claude/settings.json` as modified every single time.

`apm install` isn't the problem -- it merges fine and doesn't touch anything of
mine. It's the drift check: it compares the whole file against a replay of just
the packages, and that replay can't contain my hooks, so the file always looks
modified.

The error message says to run `apm install` to re-sync, but that doesn't help --
it re-merges correctly and the drift comes straight back.

## Repro

```bash
mkdir repro && cd repro
cat > apm.yml <<'EOF'
name: repro
version: 0.1.0
description: drift repro
author: x
targets:
  - claude
dependencies:
  apm:
    - git: DietrichGebert/ponytail
      ref: ^4.9.0
      targets: [claude]
  mcp: []
includes: auto
scripts: {}
EOF

# my own hook, nothing to do with APM
mkdir -p .claude
echo '{"hooks":{"PreToolUse":[{"matcher":"","hooks":[{"type":"command","command":".claude/hooks/mine.sh"}]}]}}' \
  > .claude/settings.json

apm install --target claude
apm audit --ci --no-policy
```

What I get:

```
│          │ drift  │ drift detected: 1 file(s): .claude/settings.json │
  drift details:
    - modified: .claude/settings.json
[x] 1 of 10 check(s) failed
```

Delete `.claude/settings.json`, run the exact same thing again, and it's
`All 10 check(s) passed`. That's the only difference. Any plugin with hooks does
this -- ponytail is just one that's public and easy to point at.

## A couple of things I noticed

`.claude/settings.json` doesn't appear anywhere in `apm.lock.yaml` -- not in
`deployed_files`, not under `deployments`, no hash. So a file with no ledger entry
is still getting reported as drifted.

Which sort of makes sense, because it isn't really a deployed file. APM edits it
in place and shares it with me. Diffing a shared file against a package-only
replay is always going to flag it.

I checked that install really is fine: I gave it a settings file with
`permissions`, `statusLine`, a `PreToolUse` entry and a `SessionStart` entry, and
all four came through untouched with the plugin's hooks added alongside.

## Why it matters

If your project has any hooks of its own -- a formatter, a secret scanner, a
pre-tool-use guard -- you can't install a plugin with hooks and keep
`apm audit --ci` passing. Since that's the documented CI gate, you end up having
to pick one or the other.

There's no way to take just the skills either. `--skill <name>` cuts down which
skills get deployed but the hooks still install.

## Possible fix

APM already writes `.claude/apm-hooks.json`, and every entry in it is tagged with
`_apm_source: "<package>"`. That looks like enough to diff only APM's own entries
and leave the rest of the file alone.

Other options if that doesn't fit:

- Store a hash of just the injected part in the lockfile and compare against that.
- Or mark merge targets as their own thing in the ledger and skip whole-file drift
  on them, so the check stops failing in a way nobody can act on.

Happy to test a patch against the repro if that helps.

danielmeppiel:
## TL;DR

`apm audit --ci` now compares only APM-owned entries in merged hook configurations such as `.claude/settings.json`. User-added hooks no longer create false drift, while an edited, removed, or malformed APM hook still produces a `modified` finding. The target registry remains the source of native config paths, and the APM sidecar remains byte-for-byte checked.

> [!IMPORTANT]
> Closes #2641.

## Problem

Drift replay compared shared merged hook config files as complete byte streams, so a user hook beside a package hook was reported as package drift. APM must still fail closed when its own hook projection is missing, edited, or malformed, without re-deriving the native-hook target vocabulary.

## Approach

| Area | Change |
|---|---|
| Ownership | `hook_ownership.py` is the sole owner of the APM-owned structural projection. |
| Comparison | Native merged configs are projected to sidecar-marked APM entries before drift comparison; `apm-hooks.json` remains byte-compared. |
| Target metadata | Projection specs come from the existing HookIntegrator registry. |
| Guardrails | Static-boundary and mutation tests prevent a projection bypass. |
| Recovery | Malformed native event shapes produce a `modified` drift finding instead of an audit crash. |

## Trade-offs

- User-owned settings and hook entries are deliberately excluded because they cannot be recovered from the APM sidecar.
- The sidecar remains byte-for-byte compared to preserve strict ownership metadata integrity.
- Invalid native event shapes are treated as `modified` drift, not ignored and not allowed to abort audit.

`apm-spec-waiver: This audit implementation correction changes no OpenAPM manifest, registry, or lockfile contract.`

## Validation

```
PYTHONPATH=src python3 -m pytest -q tests/unit/install/test_drift.py tests/integration/test_drift_check.py tests/integration/test_architecture_authorities.py::test_drift_hook_membership_exemptions_use_canonical_registries tests/integration/test_architecture_authorities.py::test_merged_hook_ownership_markers_have_one_owner tests/integration/test_architecture_authorities.py::test_shared_hook_drift_projection_guard_rejects_bypass
79 passed
```

```
PYTHONPATH=src python3 -c 'from apm_cli.cli import main; main()' audit --ci -f json
10/10 checks passed
```

```
uvx --offline ruff check src/ tests/
uvx --offline ruff format --check src/ tests/
uvx --offline pylint --disable=all --enable=R0801 --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/
```

## How to test

1. Install a package hook into a Claude-targeted project, add a user hook under `.claude/settings.json`, and run `apm audit --ci --format json`; expect exit code `0` and no drift.
2. Edit the package-owned hook without changing the sidecar; rerun audit and expect exit code `1` with `.claude/settings.json` reported as `modified`, while the user hook remains.
3. Replace the native `PreToolUse` list with `null`; rerun audit and expect exit code `1`, the same `modified` path, and byte-for-byte preservation of the malformed file.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


danielmeppiel:
Closing as superseded by merged #2682, which preserves the original contributor commit attribution and implements the same shared-hook audit fix on current main. Thank you for the original contribution.

PR #2681, opened 2026-08-24
refactor: centralize git auth config reindexing

icecold009:
## Summary

- Centralize Git authentication environment/config reindexing in a shared utility.
- Simplify the authentication and GitHub-host resolution paths around that utility.
- Add architecture-boundary guidance plus unit and integration coverage for the shared behavior.

## Validation

- `git diff --check` — passed.
- Targeted pytest was not run locally in this submission; CI should exercise the new unit and integration coverage.


danielmeppiel:
Thanks @icecold009 for working on a single implementation of Git auth-config filtering and reindexing.

Current main now has the shared [`utils/git_env.py` owner](https://github.com/microsoft/apm/blob/7f36da6e61fd5cebe29c16ced77fb4351823ec06/src/apm_cli/utils/git_env.py), with the cleanup and managed-header paths routed through it as part of #2759/#2810. That fulfills the centralization goal of #2398 without introducing another helper.

I'm closing this PR as superseded rather than asking you to refresh overlapping work. Thank you for the contribution. The resolution is on main, not in the currently published v0.29.0 release.

PR #2700, opened 2026-08-26
fix(release): restore cross-platform promotion gates

danielmeppiel:
## TL;DR

The v0.29.0 promotion dry runs exposed cross-platform test drift, a Git credential-shim deadlock, mutable network fixtures, incorrect public/private GitHub credential routing, unbounded wall-clock assumptions, and Windows release-validation drift. This PR fixes every blocker without weakening descriptor verification, authentication boundaries, checksum validation, or Agent Plugin fail-closed behavior.

> [!IMPORTANT]
> Final non-publication promotion run [33026705516](https://github.com/microsoft/apm/actions/runs/33026705516) passed on head `359c7fabb`: all five platform builds, all three integration jobs, both macOS validations, and all three standalone release-validation jobs are green. Publication jobs were intentionally skipped by `test_build=true`.

## Problem (WHY)

- Windows cannot rename an open verified descriptor, while the test assumed POSIX semantics.
- The credential shim captured persistent `git cat-file --batch-check`, causing a producer/consumer deadlock.
- Mutable upstream plugin data adopted Agent Plugins 1.0 and correctly triggered APM's fail-closed native-harness gate.
- Copilot MCP, Kiro hook, ADO identity, strict decoding, uninstall, and nested-worktree tests asserted stale contracts.
- Codex public release metadata consumed the private-module PAT instead of the workflow token.
- Concurrent sparse materialization used a plugin and a nested skill from the same repository subtree.
- Full Unix integration corpora exceeded stale 30-minute bounds on hosted runners.
- A filesystem-backed Hypothesis property used a wall-clock deadline and flaked under Windows runner load.
- Windows release validation still invoked live Copilot inference, contrary to the release pipeline's decoupled-inference contract.
- An architecture mutation sandbox omitted files required by its static guard.

These fixes keep the suite grounded in executable behavior: ["Grounding outputs in deterministic tool execution transforms probabilistic generation into verifiable action."](https://danielmeppiel.github.io/awesome-ai-native/docs/prose/). They also replace mutable assumptions with concrete fixtures because ["agents pattern-match well against concrete structures"](https://agentskills.io/skill-creation/best-practices).

## Approach (WHAT)

- Model Windows descriptor sharing explicitly while preserving the verified open handle.
- Stream non-remote Git plumbing through inherited stdio; retain captured output for finite rewritten remote commands.
- Pin plugin coverage to immutable commits and use independent sparse roots for concurrent dependencies.
- Align integration assertions with current production contracts and complete architecture mutation inputs.
- Route public GitHub release metadata through `GITHUB_API_TOKEN: ${{ github.token }}` while keeping Models and private-module credentials separate.
- Give the full hosted Unix integration corpus a measured 60-minute ceiling while retaining xdist grouping.
- Disable Hypothesis deadlines only for filesystem-backed properties; example counts and deterministic seeds remain unchanged.
- Bring Windows release validation into parity with Unix: runtime installation remains release-gated, while live inference runs only when `APM_RUN_INFERENCE_TESTS=1`.

## Implementation (HOW)

| Surface | Change |
|---|---|
| Verified asset tests | Assert WinError 32 while the descriptor is open on Windows, stable bytes, then successful rename after closure. |
| Git credential shim | Preserve streaming behavior for persistent non-remote plumbing and add a live `cat-file` regression. |
| Plugin E2E | Use immutable legacy/native fixtures and an independent skill sparse root; retain explicit native Agent Plugin rejection. |
| Integration contracts | Update Copilot MCP paths, Kiro source selection, ADO identity, strict decoding, uninstall output, and stale MCP cleanup setup. |
| Architecture guards | Require current nested-worktree fragments and copy every static-guard input into the mutation sandbox. |
| Codex setup | Prefer `GITHUB_API_TOKEN` for public Releases API metadata while preserving digest verification and fail-closed behavior. |
| Release workflow | Wire the public API token into all nine runtime-setup steps and bound full Unix corpora to 60 minutes. |
| Windows validation | Skip live `apm run` scenarios unless inference is explicitly enabled; use `GITHUB_APM_PAT` for dependency tests. |
| Property tests | Remove load-sensitive deadlines only from filesystem-backed policy cache properties. |

```mermaid
flowchart LR
    A[Promotion dry runs] --> B[Platform and fixture regressions]
    A --> C[Credential and inference routing]
    A --> D[Hosted timing bounds]
    B --> E[Deterministic regression tests]
    C --> F[Separated public private and live credentials]
    D --> G[Measured bounded gates]
    E --> H[Green promotion]
    F --> H
    G --> H
```

## Trade-offs

- Production descriptor handles remain open; only the platform-specific test expectation changes.
- Only commands without rewritten remotes inherit stdio; remote commands still capture authentication failures.
- Stable success coverage moves to APM-owned immutable fixtures; native schema drift remains a pinned fail-closed case.
- Public and private GitHub credentials are deliberately separate; Codex digest verification is unchanged.
- Release validation proves runtime installation and deterministic user flows but leaves paid/live inference to `ci-runtime.yml`.
- Hosted full-corpus steps remain bounded; their ceiling now reflects observed 24-34 minute successful runs plus runner-load headroom.

## Validation

- Final promotion: [33026705516](https://github.com/microsoft/apm/actions/runs/33026705516) - **success**.
- Builds: Linux x86_64, Linux ARM64, Windows x86_64, macOS Intel, macOS ARM - **all passed**.
- Integration: Linux x86_64, Linux ARM64, Windows x86_64, plus macOS native validation - **all passed**.
- Release validation: Linux x86_64, Linux ARM64, Windows x86_64, macOS Intel, macOS ARM - **all passed**.
- PR checks: lint, CodeQL, architecture ratchets, Windows compatibility, both Linux shards, lifecycle smoke, binary smoke, coverage, CLA, and merge gate - **all passed**.
- Full unit baseline during investigation: `20482 passed, 3 skipped, 21 xfailed, 100 subtests passed`.
- Plugin network E2E: `12 passed`; focused credential/workflow regressions: `54 passed`.

## How to test

1. Run the unit suite and CI lint mirror.
2. Run `TestPluginNetworkE2E` with a GitHub token.
3. Dispatch `build-release.yml` with `test_build=true`; expect every non-publication build, integration, and release-validation gate to pass.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -99,6 +99,7 @@ jobs:
       - name: Run smoke tests
         if: github.ref_type == 'tag' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
         env:
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_TOKEN: ${{ secrets.GH_MODELS_PAT }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}
         run: uv run pytest tests/integration/test_runtime_smoke.py -v
@@ -228,6 +229,7 @@ jobs:
         if: github.ref_type == 'tag' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
         env:
           APM_E2E_TESTS: "1"
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}
           # Intel validates the promoted lifecycle subset against its native
           # binary. Live credential acceptance belongs to auth-acceptance.yml.
@@ -259,6 +261,7 @@ jobs:
         if: github.ref_type == 'tag' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
         env:
           APM_E2E_TESTS: "1"
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}
         run: |
           cd /tmp/apm-isolated-test
@@ -342,6 +345,7 @@ jobs:
       - name: Run integration tests
         env:
           APM_E2E_TESTS: "1"
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}
           # Scarce ARM runners force the full suite onto one runner, so
           # parallelise the non-live corpus in-process with xdist.
@@ -373,6 +377,7 @@ jobs:
       - name: Run release validation tests
         env:
           APM_E2E_TESTS: "1"
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}
         run: |
           cd /tmp/apm-isolated-test
@@ -440,6 +445,7 @@ jobs:
         if: matrix.platform != 'windows'
         env:
           APM_E2E_TESTS: "1"
+          GITHUB_API_TOKEN: ${{ github.token }}
           GITHUB_APM_PAT: ${{ secrets.GH_CLI_PAT }}    # Primary: APM module access
           # The tag gate runs the non-live corpus on one runner. Four workers stay
           # below the merge gate's aggregate width while loadgroup preserves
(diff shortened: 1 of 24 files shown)


danielmeppiel:
## Docs sync advisory

Verdict: no_change  *  Pages affected: 0  *  LLM calls: 1/15  *  Took: 1s

No docs impact detected. Reason: the final PR changes release workflows, validation scripts, and test infrastructure only; it does not change a user-facing CLI, schema, primitive format, or API contract.

PR #2583, opened 2026-08-15
fix: make plugin.json dry-run match non-force pack behavior

aryansk:
## Summary

Make `apm pack --dry-run` honor the same existing-file overwrite policy as a real non-force pack.

When `plugin.json` already exists and `--force` is absent, dry-run now reports the skip instead of claiming the manifest would be written.

## Changes

- evaluate the existing-file/`--force` policy before the dry-run write preview
- preserve `--force --dry-run` behavior, which still previews an overwrite
- report the writer's canonical skip or preview outcome in `--json` output
- add regression coverage proving an existing manifest is preserved and no misleading `Would write` message is emitted

## Scenario Evidence

| Scenario | Principle(s) | Test(s) | Type | Regression trap |
| --- | --- | --- | --- | --- |
| Existing plugin manifest, `apm pack --claude-plugin --dry-run --json` | DevX | `tests/integration/test_pack_install_flow.py::TestPackCmd::test_pack_json_reports_plugin_manifest_outcomes` | integration-with-fixtures | #2553: reports `skipped`, keeps `dry_run` empty, and preserves the file. |
| Existing plugin manifest, then non-force `apm pack --claude-plugin --json` | DevX | `tests/integration/test_pack_install_flow.py::TestPackCmd::test_pack_json_reports_plugin_manifest_outcomes` | integration-with-fixtures | #2553: makes the same skip decision and leaves the explicit post-state unchanged. |
| Existing plugin manifest with `--force --dry-run` | DevX | `tests/unit/core/test_plugin_manifest.py::TestWritePluginManifest::test_dry_run_existing_file_with_force_previews_overwrite` | unit | Future-tense `Would overwrite` preview remains non-destructive. |

## Scope

This addresses the dry-run parity facet of #2553. The separate proposal for a version-drift release gate remains intentionally out of scope for this focused change.

Addresses #2553

PR #2746, opened 2026-09-01
fix(pack): fail --check-clean on a plugin.json that contradicts apm.yml (closes #2553)

### LINKED ISSUE #2553: apm pack: --dry-run reports plugin.json will be written, but non-force run silently skips it
## Summary

`apm pack --dry-run` reports it will regenerate `plugin.json`, but a real run without `--force` silently skips it — so the dry-run preview doesn't match actual behavior. Separately, neither `--check-versions` nor `--check-clean` detects this drift.

## Version

APM CLI 0.27.0

## Repro

```bash
# apm.yml has version: 1.0.8; existing .claude-plugin/plugin.json also has "version": "1.0.8"
sed -i '' 's/version: 1.0.8/version: 1.0.9/' apm.yml

apm pack --dry-run
# [i] Would write plugin manifest to .../.claude-plugin/plugin.json
# implies plugin.json will be updated to 1.0.9

apm pack
# [!] .../.claude-plugin/plugin.json already exists; skipping plugin.json generation. Re-run with --force to overwrite it.
# exits 0

grep version .claude-plugin/plugin.json
# "version": "1.0.8"   <-- unchanged, now silently out of sync with apm.yml
```

## Expected

Either:
- `--dry-run` output should say the manifest write will be **skipped** (matching real, non-`--force` behavior), not "would write"; or
- `apm pack` (no `--force`) should warn or fail when the existing `plugin.json`'s version disagrees with `apm.yml`, instead of only logging a `[!]` and exiting 0.

## Also checked: existing release gates don't catch this

```bash
apm pack --check-versions --dry-run
# [*] Version alignment OK [strategy=lockstep, expected=1.0.9]   <-- only checks apm.yml against the versioning strategy, not plugin.json

apm pack --check-clean --dry-run
# [*] Marketplace working tree clean [outputs=claude, codex]     <-- only diffs marketplace.json, not plugin.json
```

Neither existing gate would catch a `plugin.json` stuck on a stale version. We ended up writing a repo-local script (`check_plugin_metadata.py`) to assert `plugin.json` matches `apm.yml` in CI, because `apm pack`/`--check-versions`/`--check-clean` don't cover it.

## Impact

Consumers installing via `apm install owner/repo@vX.Y.Z` read `plugin.json` for the version — a stale manifest ships a wrong version string with no error anywhere in the producer's pipeline.

edenfunf:
# fix(pack): fail `--check-clean` on a stale `plugin.json`

`apm pack` never rewrites an existing `plugin.json` without `--force`, and no gate ever looked at the file. So a manifest left on an old version keeps shipping that version, and nothing goes red.

This is the second facet of #2553. The dry-run half was fixed by #2583 and is untouched here.

## The gap

`apm.yml` at `1.0.9`, committed `.claude-plugin/plugin.json` at `1.0.8`, on current `main`:

```text
$ apm pack                             -> exit 0, file stays 1.0.8
$ apm pack --check-versions --dry-run  -> exit 0
$ apm pack --check-clean               -> exit 0
```

`--check-versions` only compares `apm.yml` against the marketplace versioning strategy; `--check-clean` only diffed `marketplace.json`. The preserve path logged the same `already exists; skipping` line whether the file was identical or a release behind, so there was nothing to notice either.

## After

```text
$ apm pack --check-clean
[!] .../plugin.json disagrees with apm.yml on: version. It was left untouched; re-run
    with --force to regenerate it, or run 'apm pack --check-clean' in CI to fail the
    release on this.
[x] Plugin manifest drift [manifests=1]
[exit=4]
```

The preserve decision now reports `unchanged` (quiet `[i]`) or `stale` (warning naming the fields), and `--check-clean` reads that decision rather than re-comparing on disk, so the gate and the warning can't disagree.

## Design notes

- Comparison is one-directional: every key APM derives from `apm.yml` must match, hand-added keys are ignored. The never-overwrite policy exists so hand-maintained manifests survive, and a gate that failed on an extra `author` block would defeat it.
- An absent `plugin.json` is not drift. Pack would create it, so nothing stale ships, and failing there would break pipelines that don't commit the generated file.
- `plugin_manifests.stale` is added alongside the existing `skipped` list rather than splitting it, so current `--json` consumers are unaffected.
- Only `--check-clean` gets this, not `--check-versions`, to avoid two owners for one signal.

## Validation

Every on-disk state through the real CLI:

```text
identical / reformatted / hand-added key / absent   exit=0
stale version                                       exit=4  [drift: version]
missing generated fields                            exit=4  [drift: description, license]
not a JSON object                                   exit=4  [drift: <not a JSON object>]
```

With the production change reverted, 5 of the new CLI tests fail. With it in place: 98 passed across the pack and plugin-manifest suites, 2841 passed across every importer of the changed modules. Lint, format, pylint R0801, architecture and auth-signal gates clean. The existing `plugin_manifests` JSON contract (`written` / `skipped` / `dry_run`) is unchanged.

Refs #2553


danielmeppiel:
Thanks for the contribution. PR #2583 (d193578909e8b3a573da0cab529c1d279f80d870), now on main, already fixed this behavior. The production guard checks for an existing plugin.json before dry-run handling, so dry-run and real non-force writes both skip and preserve the file. Targeted verification of test_existing_file_preserved_without_force and test_dry_run_existing_file_reports_non_force_skip passed: 2 passed. Closing this PR as redundant.

edenfunf:
@danielmeppiel Sorry, I probably didn't describe this one well.

#2583 fixed the dry-run behaviour and this branch doesn't change that. I'm after the other half of the issue, where the release gates don't catch a stale `plugin.json`. Still happens on current `main`, with `apm.yml` at `1.0.9` and a committed `plugin.json` at `1.0.8`:

```text
$ apm pack                             -> exit 0, file stays 1.0.8
$ apm pack --check-versions --dry-run  -> exit 0
$ apm pack --check-clean               -> exit 0
```

Wrong version ships with everything green. This branch exits 4 and names the field. The two tests you ran pass here too, so they don't separate the two states.

I don't seem to be able to reopen it myself. If you think it's worth another look I'd really appreciate it, and no worries at all if not.

PR #2798, opened 2026-09-04
fix: admit schema-bearing Claude plugin manifests (closes #2780)

### LINKED ISSUE #2780: [BUG] Conforming Claude Code plugin rejected: schema-bearing .claude-plugin/plugin.json never reaches the legacy path
## Summary

`apm install` refuses any Claude Code plugin whose `.claude-plugin/plugin.json` carries the official Claude Code `$schema`. The manifest is valid for Claude Code, and the identical package installs cleanly once the `$schema` line is deleted.

The rejection comes from a root-manifest router being reused as a predicate on the legacy manifest, where it raises instead of returning.

## Reproduction

Three files, no network:

```bash
mkdir -p plug/.claude-plugin consumer
printf -- '---\nname: hello\ndescription: Minimal solo skill.\n---\n\n# hello\n' > plug/SKILL.md
cat > plug/.claude-plugin/plugin.json <<'JSON'
{
  "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
  "name": "hello",
  "version": "1.0.0"
}
JSON
cd consumer && apm install --target claude ../plug
```

Actual:

```
[x] Failed to install APM dependencies: Local package is invalid: Failed to
process Claude plugin: Unsupported schema-bearing plugin manifest:
https://json.schemastore.org/claude-code-plugin-manifest.json. APM accepts
schema-bearing plugin.json only with
https://agent-plugins.org/schemas/1.0.0/plugin.schema.json; remove $schema only
for a schema-less legacy plugin.
```

Expected: it installs as a legacy Claude plugin, which is what happens when `$schema` is absent.

Also reproduces over the network against a real package: `apm install blader/humanizer`. `--force` does not bypass it.

## What varies

| `plugin.json` location | `$schema` | Result |
|---|---|---|
| `.claude-plugin/` | Claude Code (schemastore) | `Unsupported schema-bearing plugin manifest` |
| `.claude-plugin/` | agent-plugins.org | `must be admitted from the package root` |
| `.claude-plugin/` | none | installs |
| root `plugin.json` | Claude Code (schemastore) | `Unsupported schema-bearing plugin manifest`, raised earlier |

Row 2 is the error the call site is actually reaching for. Row 1 is the same call raising before it can answer. Row 3 is what I expected row 1 to do.

## Root cause

```mermaid
flowchart TD
    A["apm install<br/>package with .claude-plugin/plugin.json"] --> B["detect_package_type<br/>MARKETPLACE_PLUGIN"]
    B --> C["_validate_marketplace_plugin<br/>validation.py:776"]
    C --> D["normalize_plugin_directory<br/>plugin_parser.py:456<br/>plugin.json = optional metadata"]
    D --> E{"classify_plugin_manifest_schema<br/>local_bundle.py:79"}
    E -->|"no $schema"| F["return LEGACY<br/>line 82"]
    E -->|"agent-plugins.org"| G["return AGENT_PLUGIN"]
    E -->|"any other $schema"| H["raise AgentPluginManifestError<br/>line 94"]
    F --> I["plugin installs"]
    G --> J["raise AgentPluginLegacyBoundaryError<br/>must be admitted from the package root"]
    H --> K["caught at validation.py:806<br/>Failed to process Claude plugin"]
    K --> L["install fails"]
    style H fill:#ffe0e0
    style L fill:#ffe0e0
```

`classify_plugin_manifest_schema` is scoped to root manifests by its own definition. `local_bundle.py:72-73` reads: "Admission route selected exclusively by root plugin.json." It returns `LEGACY` when `$schema` is absent (line 82) and raises for any `$schema` it does not recognise (line 94).

`normalize_plugin_directory` handles the legacy `.claude-plugin/` layout, where plugin.json is "treated as optional metadata" (`plugin_parser.py:460`). At line 489 it calls the router for one purpose only, to reject an Agent Plugins manifest placed outside the package root:

```python
if classify_plugin_manifest_schema(manifest) is PluginSchemaRoute.AGENT_PLUGIN:
    raise AgentPluginLegacyBoundaryError(
        "Schema-bearing plugin.json must be admitted from the package root, "
        "not Claude plugin normalization"
    )
```

With a Claude Code `$schema` the call raises, so the comparison never evaluates and the legacy path is unreachable.

## Suggested fix

Treat an unrecognised `$schema` as `LEGACY` at this call site, either by catching `AgentPluginManifestError` around `plugin_parser.py:489` or by giving the router a non-raising variant for legacy admission. Strictness for a root `plugin.json` is unaffected.

Happy to send a PR if you want the fix in that shape. Given the persona-review contract in CONTRIBUTING, I would rather confirm the approach with you first than guess at it.

## Why this shape appears in the wild

`$schema` is a generic JSON Schema pointer, and Claude Code plugins normatively carry the schemastore URL. This project's own CHANGELOG entry for #1061 cites that schema when describing what `plugin.json` array entries must contain. Any marketplace shipping schema-conformant Claude plugins is affected, and the remedy in the error text asks authors to stop conforming to their own ecosystem's schema.

## Related

* #2556: same package, different gate, fixed in 0.29.0. I verified that fix works: with `$schema` removed and `"skills": ["./"]` kept, `blader/humanizer` installs cleanly on 0.29.0. This report is the remaining barrier for that package.
* #2654 introduced the schema admission.
* #2735 and #2776 also concern package-type precedence, but they apply when a root `apm.yml` is present. This package has none.

## Environment

apm 0.29.0 (`5ac6733`), macOS arm64. Line references are against v0.29.0.

veewee:
## Problem

`apm install` refuses any Claude Code plugin whose `.claude-plugin/plugin.json` carries the official Claude Code schemastore `$schema`. Deleting that one line makes the identical package install cleanly, so the error text asks authors to stop conforming to their own ecosystem's schema.

```
[x] Failed to install APM dependencies: Local package is invalid: Failed to
process Claude plugin: Unsupported schema-bearing plugin manifest:
https://json.schemastore.org/claude-code-plugin-manifest.json.
```

`classify_plugin_manifest_schema` (`src/apm_cli/bundle/local_bundle.py:79`) is a router for root manifests, as its own docstring states: "Admission route selected exclusively by root plugin.json". It returns `LEGACY` only when `$schema` is absent, and raises for every `$schema` it does not recognise.

`normalize_plugin_directory` (`src/apm_cli/deps/plugin_parser.py:485`) reused that router as a boolean predicate on the nested legacy manifest, for one purpose: rejecting an Agent Plugin manifest placed outside the package root. A Claude Code `$schema` makes the call raise before the comparison can evaluate, so the legacy path is unreachable.

## Solution

Use the existing `is_agent_plugin_schema_id` helper (`src/apm_cli/agent_plugins/validation.py:118`) at that call site. It answers Agent Plugins namespace ownership without raising, which is all the legacy boundary check needs. Seven lines of source change, no new public API.

The router itself is untouched, so admission for a root `plugin.json` keeps its current strictness, and an Agent Plugin manifest outside the package root is still rejected. A schema-bearing `.claude-plugin/plugin.json` now deploys the same file set as a schema-less one.

| `plugin.json` location | `$schema` | Before | After |
|---|---|---|---|
| `.claude-plugin/` | Claude Code, or any foreign URL | `Unsupported schema-bearing plugin manifest` | installs as a legacy Claude plugin |
| `.claude-plugin/` | agent-plugins.org | `must be admitted from the package root` | unchanged, message reworded |
| `.claude-plugin/` | none | installs | unchanged |
| root `plugin.json` | any | unchanged | unchanged |

Both failure directions of the predicate are safe. A near-miss hostname such as `https://agent-plugins.org.example.com/schemas/1.0.0/...` does not match `AGENT_PLUGINS_SCHEMA_PREFIX`, so it routes to legacy and gains no privilege. A URL that does match can only produce a rejection here, never an admission. `$schema` is never propagated into the synthesized `apm.yml`.

## One existing test row removed

The `https://example.com/plugin.schema.json` case in `test_nested_marketplace_manifest_cannot_bypass_root_schema_admission` encoded the bug rather than the gate: a foreign `$schema` is not an Agent Plugin manifest, so it has no root admission to bypass. The two Agent Plugins rows, which are what the test guards, remain.

Fixes #2780

## Type of change

- [x] Bug fix

The commit carries an `apm-spec-waiver:` line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `apm install` now accepts a conforming Claude Code plugin whose
+  `.claude-plugin/plugin.json` carries the official Claude Code `$schema`.
+  The root-manifest schema router was reused as the predicate on the nested
+  legacy manifest, and that router raises on an unrecognised `$schema` rather
+  than answering, so the legacy path could never be reached. Strict admission
+  for a root `plugin.json` is unchanged, and an Agent Plugin manifest placed
+  outside the package root is still rejected. (closes #2780) (#2798)
 - Windows users no longer get repeated line-ending churn when APM rewrites
   `apm.yml`: install, uninstall, dependency resolution, and revision-pin
   updates now produce deterministic LF output. An `apm.yml` that already uses
--- a/src/apm_cli/deps/plugin_parser.py
+++ b/src/apm_cli/deps/plugin_parser.py
@@ -480,11 +480,17 @@ def normalize_plugin_directory(plugin_path: Path, plugin_json_path: Path | None 
     ):
         manifest = parse_plugin_manifest(plugin_json_path)
         from ..agent_plugins.errors import AgentPluginLegacyBoundaryError
-        from ..bundle.local_bundle import PluginSchemaRoute, classify_plugin_manifest_schema
+        from ..agent_plugins.validation import is_agent_plugin_schema_id
 
-        if classify_plugin_manifest_schema(manifest) is PluginSchemaRoute.AGENT_PLUGIN:
+        # Only an Agent Plugin manifest is barred here.  A foreign $schema,
+        # including the Claude Code schemastore URL that conforming Claude
+        # plugins carry, is legacy metadata.  classify_plugin_manifest_schema()
+        # is scoped to root manifests and raises rather than answers for an
+        # unrecognised $schema, so it cannot serve as the predicate (#2780).
+        schema_id = manifest.get("$schema")
+        if isinstance(schema_id, str) and is_agent_plugin_schema_id(schema_id):
             raise AgentPluginLegacyBoundaryError(
-                "Schema-bearing plugin.json must be admitted from the package root, "
+                "An Agent Plugin manifest must be admitted from the package root, "
                 "not Claude plugin normalization"
             )
 
--- a/tests/unit/agent_plugins/test_loader.py
+++ b/tests/unit/agent_plugins/test_loader.py
@@ -926,13 +926,18 @@ def test_schema_bearing_manifest_never_falls_back_to_legacy_normalization(
     [
         PLUGIN_SCHEMA_ID,
         "https://agent-plugins.org/schemas/2.0.0/plugin.schema.json",
-        "https://example.com/plugin.schema.json",
     ],
 )
 def test_nested_marketplace_manifest_cannot_bypass_root_schema_admission(
     tmp_path: Path,
     schema_id: str,
 ) -> None:
+    """Only an Agent Plugin manifest is barred from the nested legacy path.
+
+    A foreign ``$schema``, such as the Claude Code schemastore URL, is legacy
+    metadata and normalizes.  ``TestNormalizePluginDirectory`` in
+    ``tests/unit/test_plugin_parser.py`` covers that case (#2780).
+    """
     manifest = tmp_path / ".claude-plugin" / "plugin.json"
     manifest.parent.mkdir()
     manifest.write_text(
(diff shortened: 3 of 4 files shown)


veewee:

veewee:
Fixed in https://github.com/microsoft/apm/pull/2797

Issue #2566, opened 2026-08-12
[BUG] GitLab (self-hosted) org-policy auto-discovery is impossible: all candidate repo names are invalid GitLab project paths


rrazvd:
**Describe the bug**
On self-hosted GitLab, APM org-policy auto-discovery can never resolve, because every candidate repository name in the discovery cascade is an invalid GitLab project path.

Per the [Policy Files docs](https://microsoft.github.io/apm/enterprise/apm-policy/#where-it-lives), discovery probes, in order: `.github-private` → `.github` → `.apm` → `_apm`. GitLab rejects any project path that starts with `.`, `_`, or `-`:

> Path can only include non-accented letters, digits, '_', '-' and '.'. It must not start with '-', '_', or '.', nor end with '-', '_', '.', '.git', or '.atom'.

So none of the four candidates can be created:

| Candidate | Valid on GitLab? | Reason |
| --- | --- | --- |
| `.github-private` | ❌ | leading `.` |
| `.github` | ❌ | leading `.` |
| `.apm` | ❌ | leading `.` |
| `_apm` | ❌ | leading `_` |

This is stricter than Azure DevOps (#2429): ADO forbids a leading `_` only on the *project* name, so PR #2450 could keep repository `_apm` inside project `apm`. On GitLab the `_apm` fallback is also invalid, so there is no working coordinate at all. Additionally, probing these paths on our host surfaces an HTTP 410, which prints a policy-fetch warning on every `apm install` / `apm audit` (inconsistent with the docs stating GitLab "falls through with no policy applied").

**To Reproduce**
1. On a project whose `origin` is a self-hosted GitLab remote (`git@gitlab.example.com:<group>/<repo>.git`), run command `apm audit --ci` (without `--no-policy`), or `apm install`.
2. With parameters: default org-policy auto-discovery (no explicit `--policy`).
3. See error: a policy-fetch warning (HTTP 410 on our host) and no org policy applied. Then try to create the org policy repo under the group — GitLab rejects `.github-private`, `.github`, `.apm`, and `_apm` with the path error quoted above.

**Expected behavior**
A supported way to host and auto-discover org policy on self-hosted GitLab using a valid, configurable GitLab project path (e.g. `apm-policy`), plus a clean "no policy" outcome (not a 410 warning) when none is present — analogous to the ADO project-coordinate fix in PR #2450, but for GitLab where even `_apm` is an invalid path.

**Environment:**
 - OS: macOS
 - APM Version: 0.25.0
[...]

PR #2933, opened 2026-09-09
feat(policy): walk GitLab subgroups for org policy discovery (#2753)

### LINKED ISSUE #2753: [FEATURE] GitLab: support subgroup-scoped policy discovery (team-level apm-policy), not only the top-level group

On GitLab, org-policy auto-discovery is anchored exclusively to the **top-level group**: APM takes the first path segment of the project remote and looks for `<top-level-group>/apm-policy` (nested subgroup scopes are explicitly not searched, per the [Policy Files docs](https://microsoft.github.io/apm/enterprise/apm-policy/#where-it-lives) and the fix in #2566 / PR #2662, shipped in v0.29.0).

This is a problem for large organizations where the top-level group is shared across many independent teams and departments. Consider a project whose remote is:

```
gitlab.com/acme/dept-a/area-1/team-x/my-project
```

The first segment is `acme` — the whole company. To have any policy enforced at `apm install` time, it would have to be published at `acme/apm-policy`, which would govern **every team and department** under `acme`, not just the one that owns the project. A team that only wants to govern its own projects (under `.../team-x`) has no supported way to scope a policy to its subgroup.

The practical effect today: `apm install` prints `No org policy found for this project`, `apm audit --ci` reports `No org policy found ...; enforcement skipped`, and CI stays green while **none** of the team's dependency/MCP/compilation rules are ever enforced — a false sense of security.

Subgroup-aware policy discovery on GitLab. When resolving a project's org policy, APM should walk up the group hierarchy from the project's immediate parent to the top-level group, looking for an `apm-policy` project at each level, and apply the closest one found (composing with ancestors via the existing `extends:` chain). For example, for a remote like `gitlab.com/acme/dept-a/area-1/team-x/my-project`, APM would probe, in order:

```
acme/dept-a/area-1/team-x/apm-policy
acme/dept-a/area-1/apm-policy
acme/dept-a/apm-policy
acme/apm-policy
```

This mirrors how GitLab itself scopes settings (CI/CD variables, runners, etc.) hierarchically by group/subgroup, and lets a team enforce policy over just its own projects without owning company-wide governance.

- **Top-level `<company>/apm-policy` only:** rejected — it would impose one team's rules on the entire organization.
- **An explicit override env var** (e.g. `APM_GITLAB_POLICY_GROUP=<full/sub/group>`) pointing discovery at a specific subgroup: acceptable as a minimal first step, but note that a similar `APM_GITLAB_POLICY_ROOT_GROUP` override was added and then removed in PR #2662 (*"remove the unsupported root-group override"*), so the design tradeoffs here need revisiting.
- **`apm audit --ci --policy ./apm-policy.yml` pointing at a repo-local file in CI:** a common workaround, but `--policy` is audit-only and is **not** wired through `apm install`, so install-time enforcement is still skipped and the policy isn't shared/discovered across the team's other repos.
- **`--no-policy`:** disables enforcement entirely — not acceptable for a governance use case.

**Additional context**

- Related: #2566 (GitLab top-level-group discovery, closed via PR #2662, v0.29.0) — this request extends that work from the root group down to subgroups.
- Related umbrella: #1925 (policy enforcement on non-GitHub remotes).
- Environment: APM 0.29.0, macOS, `gitlab.com` SaaS, deeply nested subgroups.
- Observed with `apm install --verbose` → `[i] No org policy found for this project`, and `apm audit --ci` → `[!] No org policy found at unknown; enforcement skipped`.

rrazvd:
## TL;DR

GitLab org-policy auto-discovery only ever probed the **top-level group** (`<top-level-group>/apm-policy`), so a project nested under subgroups could not enforce its own governance without changing the whole organization. This PR walks the GitLab subgroup tree from the project's own group up to the top-level group and applies the **closest** `apm-policy` (closest wins), and lets `extends:` compose across nested namespaces. Flat `<group>/<project>` remotes are unchanged. Closes #2753.

## Problem (WHY)

For a project at `gitlab.com/acme/dept-a/team-x/my-project`, discovery resolved only `acme/apm-policy` and discarded every intermediate subgroup:

- A single team could **not** scope its own policy under its subgroup — the only place APM looked was the top-level group, imposing one team's rules org-wide (or none).
- The prior docs stated the limitation outright: *"GitLab discovery uses only the top-level group ... It does not search nested subgroup scopes."*
- This is the accepted follow-up to #2566 (top-level GitLab discovery, shipped in v0.29.0). Issue #2753 is `status/accepted`.

Root cause: `_parse_remote_url` returned only `path_parts[0]` as the org, and `_auto_discover` passed that single org to the GitLab fetch — the full namespace was never derived.

## Approach (WHAT)

| Decision | Choice |
|---|---|
| Leaf selection | Walk namespaces deepest → top-level; **first non-`absent` wins** (closest) |
| No policy at a level | `absent` (404) ascends to the parent group |
| Closer policy the token can't read | GitLab 404s a private project the same as a missing one, so a denied closer `apm-policy` is skipped in favour of the next ancestor (`404 == no policy`, like GitHub/ADO) — a platform limitation, documented (grant the token read access). |
| Error / malformed / 403 | **Fail closed immediately** — never coerced into "no policy" |
| Composition | Opt-in via existing `extends:`; now accepts nested-namespace refs |
| Flat / personal namespace | Probes only `<group>/apm-policy` — identical to prior behavior |
| Remote parsing | Extracted to a dedicated `policy/_remote.py` owner (no split authority; architecture-guard enforced) |

## Implementation (HOW)

| File | Change |
|---|---|
| `src/apm_cli/policy/_remote.py` (new) | The git-remote identity family, extracted from `discovery.py` (which grew past the 2100-line file-length ratchet): `_git_remote_origin_url` (single `git remote get-url origin` owner), `_remote_url_parts` (single host+segments splitter), `_parse_remote_url` (refactored to consume the splitter — behavior-preserving for GitHub/GHE/ADO/visualstudio), plus `_extract_org_host_port_from_git_remote` / `_extract_org_from_git_remote`. `discovery.py` re-exports them, so call sites/tests are unchanged. |
| `src/apm_cli/policy/discovery.py` | New `_gitlab_namespace_descending` (deepest→top namespaces from the origin remote) and `_gitlab_walk_candidate` (the walk: first non-`absent` wins, `absent` ascends). Reads `origin` once and reuses it for identity + the walk. |
| `src/apm_cli/policy/discovery.py` (`_extract_extends_host`) | Leaf-aware pre-fetch host-pin guard: a bare single-label first segment is a namespace **only on GitLab leaves** (`acme/dept-a/apm-policy`); on GitHub/ADO leaves it is still parsed as a (cross-)host, so `extends: "evil/org/repo"` remains rejected (Security Finding F1 preserved). |
| `src/apm_cli/policy/_gitlab.py` | `_fetch_gitlab_chain_parent` accepts nested-namespace `extends:` refs (`namespace = parts[:-1]`, `repo = parts[-1]`) after an optional leaf-host-prefix strip. |
| `scripts/architecture_linter/checks/install_policy_gitlab_and_bundle.py`, `install_policy_intent.py`, `.apm/architecture/owners/install-deployment.json` | Add a static delegation guard (`install-deployment-policy-remote-origin-owner`), bound to a dedicated `_remote.py` owner-registry record: only `_remote.py` may read `git remote get-url origin` or define the remote-URL splitter/parsers. Also re-point the existing `gitlab-facade-orchestration` guard at `_gitlab_walk_candidate`. |
| tests | New `test_discovery.py` coverage for the walk, nested `extends:`, and F1; an owner-guard mutation-break case in `test_architecture_owner_rule_mutations.py` and the frozen-inventory entry in `test_architecture_runner.py` for the new guard. |
| docs + `CHANGELOG.md` | Updated `apm-policy.md`, `governance-guide.md`, `policy-reference.md`, `apm-usage/governance.md`; added the `Unreleased` changelog entry. |

> [!NOTE]
> The GitLab REST transport and `AuthResolver` auth already URL-encode a full multi-segment `project_path`, so no transport/auth changes were needed — `org="acme/dept-a"` was already fetchable.

## Diagram

The GitLab branch of `_auto_discover`: each namespace level is probed closest-first; only a clean `absent` ascends, every other outcome returns immediately.

```mermaid
flowchart TD
    A[origin remote] --> B[_gitlab_namespace_descending]
    B --> C["deepest: acme/dept-a/team-x"]
    C -->|found / error| R[return result]
    C -->|absent| D["acme/dept-a"]
    D -->|found / error| R
    D -->|absent| E["acme (top-level)"]
    E -->|found / error| R
    E -->|absent| Z[outcome = absent, clean no-op]
```

## Trade-offs

- **Up to N REST probes when no policy exists** (N = subgroup depth). The common no-governance case costs a few extra REST calls; the walk short-circuits on the first hit, and logging stays at `debug` so `--verbose` is not spammed. Absent REST results are served from the existing per-namespace policy cache on warm runs; caching the *absent* REST outcome itself is a possible follow-up, out of scope. (The concealment Git probe, a separate path, **is** cached — see below.)
- **GitLab conceals private projects with 404.** A closer `apm-policy` the token cannot read is indistinguishable from a missing one, so it is skipped in favour of the next ancestor — the same `404 == no policy` behaviour GitHub/ADO discovery already have. An earlier revision tried to detect this via an authenticated `git ls-remote` probe; it was removed after review as scope creep (it caught only a narrow case, could not be soundly cached, and false-positived on an `apm-policy` project containing only a README). Mitigation: grant the CI token read access to every `apm-policy` it should honour.
- **Closest-wins is not tighten-only at the discovery layer.** Anyone who can create an `apm-policy` project in an intermediate subgroup can publish a *weaker* policy that shadows a stricter ancestor for projects under it (only `extends:` chains enforce tighten-only). This is intrinsic to team-scoped governance; the mitigation (restrict `apm-policy` project creation via GitLab RBAC) is documented in the governance guide's sharp-edges list.
- **`extends: "org"` resolves to the top-level group**, not the nearest ancestor — kept consistent with GitHub/ADO semantics and documented explicitly, rather than silently redefining the shorthand per host.
- **Nested-namespace vs cross-host ambiguity** in `extends:` refs is resolved by an FQDN heuristic (a host has a `.` or `:port`); a GitLab group whose name literally contains a dot would fail closed (rejected). This is safe and rare, and the docs prescribe the host-qualified spelling (`gitlab.com/acme.tools/...`, with `host:port` when the origin uses a port) as the workaround.

## Benefits

1. A team under any subgroup depth can own its `apm-policy` without touching org-wide governance.
2. A top-level `acme/apm-policy` still applies to every project with no closer policy — zero migration for existing setups.
3. Cross-level composition works via `extends:` (including nested-namespace parents).
4. Remote-URL splitting and the git-remote subprocess each now have a single canonical owner (architecture-rule aligned).

## Validation

All CI-mirror gates pass locally on the merge base.

<details><summary>Lint, duplication, tests, architecture, conformance</summary>

```
$ uv run --extra dev ruff check src/ tests/            -> All checks passed!
$ uv run --extra dev ruff format --check src/ tests/   -> 1768 files already formatted
    --min-similarity-lines=10 --fail-on=R0801 src/apm_cli/  -> 10.00/10
$ uv run python -m pytest tests/unit/policy tests/spec_conformance/test_policy_reqs.py
    -> 1246 passed, 1 xfailed
$ uv run python -m pytest tests/unit/install -k "policy or preflight or discover"
    -> 323 passed
$ uv run python -m pytest tests/integration/test_architecture_semantic_rule_mutations.py
    -> 148 passed
$ uv run python -m pytest tests/unit/scripts/test_architecture_runner.py
    -> 72 passed
$ bash scripts/lint-architecture-boundaries.sh          -> exit 0
```
</details>

### Scenario evidence

| User-promise scenario | Proven by | APM principle |
|---|---|---|
| Closest subgroup policy wins | `test_gitlab_subgroup_closest_wins` | Governance is scoped, not all-or-nothing |
| Absent level ascends to parent | `test_gitlab_subgroup_absent_ascends_to_parent` | Deterministic discovery order |
| No policy anywhere → clean `absent` | `test_gitlab_subgroup_all_absent_is_clean_absent` | Missing policy is not an error |
| Error fails closed (first and inner level) | `test_gitlab_subgroup_error_fails_closed`, `..._error_at_inner_level_fails_closed` | Fail closed on ambiguity |
| Flat / personal namespace unchanged | `test_gitlab_flat_project_falls_back_to_top_level_org` | No regression |
| Nested `extends:` allowed, cross-host FQDN rejected | `TestValidateExtendsHostNestedNamespace` | Host-pin guard (F1) intact |
| `_parse_remote_url` refactor is behavior-preserving | pre-existing `TestParseRemoteUrl` ADO/GHE/visualstudio cases | Single-owner refactor, no behavior drift |

> [!IMPORTANT]
> The nested-`extends:` host-pin fix was surfaced by the local `apm-review-panel` dogfood pass (supply-chain + auth personas, independently): the original tests exercised `_fetch_chain_parent` directly and bypassed `_validate_extends_host`, masking the reject. Regression tests now drive the real validation path.

## How to test

- [ ] Point a repo's `origin` at a nested GitLab remote (`gitlab.com/acme/dept-a/team-x/my-project`) and publish `apm-policy.yml` in `acme/dept-a/apm-policy`.
- [ ] Run `apm audit --ci` (or `apm install`) and confirm the `acme/dept-a` policy is applied (closest wins), not the top-level one.
- [ ] Remove that policy; confirm discovery falls through to `acme/apm-policy`.
- [ ] Remove all policies; confirm a clean `absent` (no warning, no fetch-failure).
- [ ] `uv run python -m pytest tests/unit/policy/test_discovery.py`.


rrazvd:
Thanks for the review — addressed all four findings in `e9cf62ef4`.

**🔴 Critical — host-pin bypass for single-label prefixes (`discovery.py`)**
Correct catch: my `_extract_extends_host` relaxation was too broad. It's now **leaf-aware** — a bare single-label first segment is treated as a namespace **only when the leaf is GitLab** (where nested subgroups make `acme/dept-a/apm-policy` valid). On GitHub/ADO leaves a 3-segment ref is `host/owner/repo`, so `extends: "evil/org/.github"` is again parsed as a cross-host ref and rejected (F1 preserved). Regression tests added: `test_single_label_first_segment_stays_a_host_on_github_leaf`, `test_validate_rejects_single_label_cross_host_on_github_leaf`.

**🟠 Moderate — host-like prefix / malformed port (`_gitlab.py`)**
A host-like first segment (FQDN or `host:port`) must now match the leaf host **and** port exactly; otherwise the ref is rejected as invalid instead of being folded into the namespace. Port parsing is inside the `try`, so a malformed port (`gitlab.example.test:not-a-port/...`) fails closed instead of raising `ValueError`. Tests: `test_gitlab_parent_rejects_host_prefix_with_mismatched_port`, `test_gitlab_parent_rejects_malformed_port_without_crashing`.

**🟠 Moderate — origin read twice (`discovery.py`)**
`_auto_discover` now reads `git remote get-url origin` once and reuses it for both identity and the namespace walk — `_gitlab_namespace_descending` takes the already-read URL, and `_extract_org_host_port_from_git_remote` accepts it via a sentinel so it never re-reads. Removes the double subprocess and the TOCTOU on a changing remote. The single-subprocess contract test (`tests/unit/policy/test_discovery.py`, ADO) still passes.

**⚪ Nit — publishing checklist (`governance.md`)**
Updated to say GitLab policy may live under any ancestor namespace (closest wins), not the top-level group only.

Local gates green: `ruff check`/`format`, `pylint R0801` 10/10, `tests/unit/policy` + `tests/spec_conformance/test_policy_reqs.py` (1247 passed), `tests/unit/install -k "policy or preflight or discover"` (323 passed), `scripts/lint-architecture-boundaries.sh` (exit 0).

rrazvd:
Addressed the second review in `7ae2ce674`.

**Canonical-owner guard (incomplete) — `discovery.py`**
Added a registered static delegation guard, `install-deployment-policy-remote-origin-owner` (in `install_policy_gitlab_and_bundle.py`, wired via `install_policy_intent.EXTRA_RULES`). It forbids any policy-tree module other than `discovery.py` from reading `git remote get-url origin` or defining the remote-URL splitter/parsers (`_remote_url_parts`, `_parse_remote_url`, `_git_remote_origin_url`), so a future caller cannot reintroduce the double-read/divergent-parse. Verified it has teeth (a planted violation in another policy file is flagged; removing it returns the linter to green). This mirrors the existing `install-deployment-gitlab-policy-adapter` guard, which likewise ships without a standalone owner-inventory record: `discovery.py` already has exactly one owner record (`cached-policy-shape` in `contracts-tooling.json`) and the registry rejects both a second record selecting the same file and a cross-group guard reference, so the guard is the enforcement mechanism here.

**Dotted GitLab namespace guidance (missing) — `apm-policy.md`**
Added a note: a bare nested `extends:` ref whose first segment contains a dot (group named e.g. `acme.tools`) is read as a host and rejected; users must spell it host-qualified (`extends: "gitlab.com/acme.tools/team/apm-policy"`).

**Conformance coverage misclassified — `test_policy_reqs.py`**
Agreed — removed the `req-pl-011` test that only re-asserted the GitLab provider's namespace order. Per req-pl-011 §6.1.1 that order is *implementation-defined*, so it isn't a normative conformance property, and the unit suite already covers it. Regenerated `CONFORMANCE.*` (req-pl-011 back to 2). The policy-tree change is covered under Mode B by an `apm-spec-waiver:` trailer (the walk refines the existing req-pl-011 provider; no new normative requirement) — the detector reports `WAIVED`.

Local gates: `ruff check`/`format` (src/tests) clean, `scripts/lint-architecture-boundaries.sh` exit 0, Mode B detector exit 0 (WAIVED), `tests/unit/policy` + `tests/spec_conformance` green (the one unrelated `copilot_plugins` failure is pre-existing on `main`).

rrazvd:
Addressed the third review in `0eebdf871`.

**Vacuous facade guard + failing mutation (`discovery.py`)**
Correct — hoisting `is_gitlab` renamed the branch to `elif is_gitlab:` and moved the adapter call into `_gitlab_walk_candidate`, so `check_gitlab_facade_orchestration` scanned a stale `elif is_gitlab_hostname(host):` region and passed vacuously (its mutation case `install-deployment-gitlab-facade-orchestration` no longer produced a violation). The check now scopes its orchestration scan to `_gitlab_walk_candidate`'s body. Verified: `test_semantic_rule_catches_its_mutation[install-deployment-gitlab-facade-orchestration]` passes again.

**Guard-less rule without a mutation case (`install_policy_intent.py`)**
Correct — the frozen matrix asserts exact set-equality with every guard-less rule. Gave `install-deployment-policy-remote-origin-owner` teeth the matrix can exercise (it now also requires `discovery.py` to define all three canonical read/parse helpers) and added its mutation case (`def _git_remote_origin_url(` -> `_impl`). `test_matrix_covers_every_guardless_rule_exactly_once` and the new case's `catches_its_mutation` both pass. I kept it as a guard-less semantic rule rather than an owner-registry record because `discovery.py` already has exactly one owner record (`cached-policy-shape`) and the registry rejects a second record selecting the same file — this mirrors the sibling `install-deployment-gitlab-policy-adapter`/`-facade` guards.

**policy-reference.md internal inconsistency**
Updated the `extends` value table (added the GitLab `namespace/.../repo` ancestor form and the top-level-group meaning of `org`) and the Discovery-vs-`extends` note (now states GitLab discovers the closest `apm-policy` up the subgroup tree), so the page has one consistent contract.

**Dotted-group workaround incomplete for explicit ports (apm-policy.md)**
Documented that when the origin uses a port, the host-qualified `extends:` ref must carry the exact `host:port` authority (e.g. `gitlab.example.com:8443/acme/team/apm-policy`), since `GITLAB_HOST` is only the hostname and `_fetch_gitlab_chain_parent` requires host+port to match.

Local gates: `ruff` (src/tests) clean, `scripts/lint-architecture-boundaries.sh` exit 0, mutation matrix structural + the two affected cases green.

rrazvd:
Pre-emptive hardening in `7bcbc1854` (before requesting another review): ran the full architecture suite locally and caught one more gap the earlier per-case runs missed.

`tests/unit/scripts/test_architecture_runner.py::test_registered_rule_inventory_matches_frozen_semantic_contract` pins a frozen set of every registered rule id. The new `install-deployment-policy-remote-origin-owner` guard was missing from it, so this test would have failed in CI. Added it to `_EXPECTED_RULE_ID_TEXT`.

Full local runs now green:
- `tests/integration/test_architecture_semantic_rule_mutations.py` — 148 passed
- `tests/unit/scripts/test_architecture_runner.py` — 72 passed
- registry / registry-conflicts / intent-guards / owner-rule-mutations / linter-entrypoint — all passed
- `ruff` (src/tests), `scripts/lint-architecture-boundaries.sh` (exit 0)

rrazvd:
Addressed in `fcacaa26e` — bound the remote-origin guard to the owner registry as requested.

**What changed**
- `install-deployment-policy-remote-origin-owner` is now a real **owner-guard**: registered with `guard_ids=(<id>,)` (a `Rule(...)`, not `_semantic_rule`), so it is a *known* guard the registry validates bidirectionally and requires to execute exactly once.
- It is referenced from `discovery.py`'s existing owner record (`cached-policy-shape` in `contracts-tooling.json`). Note on the earlier back-and-forth: `discovery.py` can hold only one owner-registry record (selector uniqueness is global), and a guard's id-prefix is independent of the shard it's referenced from, so the remote-origin decision joins that record's guard list rather than getting a second record.
- Moved its mutation-break case from the guard-less matrix to the owner-guard matrix (`test_architecture_owner_rule_mutations.py`).

**Verification** — full architecture suite green:
- `test_architecture_owner_rule_mutations.py` + `test_architecture_semantic_rule_mutations.py` + `test_architecture_runner.py` + `test_architecture_registry.py` + `test_architecture_registry_conflicts.py` + `test_architecture_intent_guards.py` -> 603 passed
- `scripts/lint-architecture-boundaries.sh` -> exit 0 (guard executes once, bound to its owner record)
- The guard's mutation case fires (`test_owner_rule_catches_its_guard_mutation[install-deployment-policy-remote-origin-owner]`), and HEAD is clean.

rrazvd:
Addressed both findings in `4de20be38`.

**Concealed GitLab 404 -> weaker ancestor (fail-open)**
Real concern, thanks. GitLab returns 404 both for a missing project and for a private one the token cannot read, so the walk could silently downgrade a denied closer team policy to a weaker ancestor. Fix: before applying an ancestor over skipped closer levels, `_gitlab.first_concealed_closer_policy` verifies via **authenticated Git** whether any skipped closer `apm-policy` project actually exists; if one is confirmed, discovery **fails closed** (`cache_miss_fetch_fail`) naming the concealed project, instead of applying the weaker ancestor. The common inheritance case (a team with no closer project) is unaffected -- Git can't confirm a non-existent project, so the walk ascends normally, and the all-absent path never probes. The residual case (a token blind to the project over *both* REST and Git) is indistinguishable from genuinely absent and still ascends; that's a GitLab-platform limitation, now documented with a `caution` note advising read access for every honoured `apm-policy` project. Kept `_gitlab_project_state_via_git` inside the adapter so the facade-orchestration guard still holds. Tests: `test_gitlab_subgroup_concealed_closer_policy_fails_closed`, `TestFirstConcealedCloserPolicy`, and the ascend test now asserts the skipped levels are verified.

**Origin-read guard didn't enforce single delegation**
Correct -- the guard only required the three helper names to exist, so a second `subprocess.run([... "remote", "get-url", "origin"])` elsewhere in `discovery.py` would have passed. Added an exact-count clause (the origin-read argv must appear exactly once in the owner) and switched the owner-guard mutation to **introduce a duplicate read**, so the single-reader invariant is now enforced and proven by the mutation matrix.

Verification (full local runs): owner-guard + guard-less mutation matrices, runner, `tests/unit/policy`, `tests/spec_conformance/test_policy_reqs.py`, install-side policy -> **2139 passed**; `ruff`/`format` clean; `scripts/lint-architecture-boundaries.sh` exit 0.

rrazvd:
Ran the repo's `apm-review-panel` dogfood (7 personas) over the full branch diff before requesting another review; it surfaced real issues, fixed in `af821d5f5`.

**Fail-open by default (devx, blocking)** — the concealed-closer result returned `cache_miss_fetch_fail`, which honours `policy.fetch_failure_default` (default `warn`), so a concealed private closer policy *silently proceeded with no enforcement* — contradicting the doc's "fails closed." Fixed: it now returns `incomplete_chain`, which **always fails closed** regardless of the knob, and the CLI surfaces the specific "closer policy concealed — grant read access" message instead of the generic "check connectivity" copy. Added a gate-level test asserting it blocks under `fetch_failure_default=warn`.

**Uncached git probe on the hot path (performance, blocking)** — the concealment `git ls-remote` probe ran uncached on **every** install in the common "team inherits the org policy" path (N-1 network round-trips). Fixed: each probe verdict is cached per `(host, namespace, repo)` under the policy cache (same TTL); `--no-cache` bypasses it. Tests cover cache-hit (probes once) and no-cache-bypass.

**Closest-first coverage gap (test-coverage, blocking)** — `first_concealed_closer_policy` was only tested with a match at index 0. Added middle-level and among-many-confirmed tests.

**Docs (devx + doc-writer)** — reworded the concealed-policy caution to state the *unconditional* block; added a governance-guide "sharp edges" bullet covering both the closest-wins shadowing risk (a closer subgroup policy can weaken a stricter ancestor — restrict `apm-policy` project creation via GitLab RBAC) and the concealed-blocks-install behavior; added the skill note.

Panel also confirmed no blocking issues in: the leaf-aware F1 host-pin, the `_remote.py` extraction (faithful, no auth/behavior drift), fail-closed semantics across the walk, credential routing in the git probe, and internal doc consistency.

Local gates green: `ruff`/`format`, `pylint R0801`, `tests/unit/policy` (1235), install-side policy (323), owner+guardless mutation matrices, arch linter, Mode B (WAIVED), file-length guard, YAML/relative_to/auth-signals. The full unit suite's only failures were pre-existing/environmental (xdist-parallelism flakes in unrelated `test_uninstall_*` / windows-diagnostics files; they pass in isolation).

rrazvd:
Addressed all four findings in `7e7b1f4d6`.

**Caching indeterminate as absent (blocking, `_gitlab.py`)** — correct and important. `_gitlab_project_state_via_git` returns `None` for a missing project AND for auth/network/timeout failures alike, so caching an `absent` verdict let a transient failure be remembered as "confirmed absent" for the TTL and apply a weaker ancestor. Fixed: only the **definitive `present`** verdict is cached now; `None` is never cached. Since absence is never definitive, the common inherit-from-org path re-probes each install — that cost is accepted (documented in the trade-offs) rather than ship an unsound negative cache.

**Network access in cache-only mode (blocking, `discovery.py`)** — `cache_only` is the offline local-bundle contract; the probe could still `git ls-remote`. Fixed: `cache_only` is threaded into `first_concealed_closer_policy`, which never probes in that mode — a cached `present` fails closed, and an unverifiable level **fails closed deterministically** instead of reaching out. Test: `test_cache_only_never_probes_and_fails_closed_when_unverifiable`.

**`cached_stale` ancestor bypassed the concealment check (suppressed)** — the gate was `outcome in {found, empty}`, but a `cached_stale` ancestor also carries a usable policy and would be applied without checking for a concealed closer. Fixed: gate on `result.policy is not None`.

**Rule description named the wrong owner** — updated the `install-deployment-policy-remote-origin-owner` description to say `policy/_remote.py` (matching the registry record, guard, and definitions).

Verification: `ruff`/`format`, arch linter (exit 0), `tests/unit/policy` (1236) green; new tests cover present-cached / indeterminate-not-cached / cache-only-fails-closed / closest-among-many.

rrazvd:
Rather than fix the fifth round of findings on the concealed-policy detection, I removed the mechanism entirely (`9f4231450`) — it was scope creep beyond #2753 and, as the reviews established, not worth its complexity.

**Why it's gone**
- It only caught a narrow case (project git-readable but the policy *file* 404s).
- It could not be soundly negative-cached: `git ls-remote` can't distinguish a missing project from an auth/network failure, so a transient error would have been cached as "absent."
- This round's finding was decisive: the probe checks *project* existence, but a project can exist with only a README — so it **false-positived and blocked inheritance from a valid ancestor**, worse than the ambiguity it targeted.

**What replaces it**
A GitLab `absent` (404) simply ascends to the next ancestor — the same `404 == no policy at this level` behaviour GitHub and ADO discovery already have, and strictly better than the pre-PR single-level behaviour (an ancestor policy now applies where nothing did before). The GitLab 404-conceals-private-projects ambiguity is documented as a platform limitation (grant the CI token read access), matching how the docs already treat 404s elsewhere.

This resolves all four findings from the last review by removal, and drops ~290 lines (net diff 1135 → 844 insertions). The core of the PR is unchanged: subgroup walk, nested `extends:`, leaf-aware host-pin (F1), and the `_remote.py` extraction + its owner-guard.

Local gates green: `ruff`/`format`, arch linter (exit 0), `tests/unit/policy` (1227) + spec (1246 combined), install-side policy (323), owner + facade mutation cases. `discovery.py` is 2063 lines (well under the ratchet).

rrazvd:
Addressed the latest Copilot review.

**Owner-guard vs owner-registry mismatch** (`install_policy_gitlab_and_bundle.py`): the `_remote.py` owner record names `_extract_org_host_port_from_git_remote` as part of the canonical git-remote decision, but the static guard `install-deployment-policy-remote-origin-owner` only counted/banned three defs -- a second identity extractor added elsewhere would have slipped through. Fixed in `4851ba7ac`: added the extractor to `_REMOTE_PARSER_DEFS` and bumped the expected count to 4. Verified the guard has teeth (renaming the extractor now fails with `found 3`), the arch linter passes clean at HEAD, and the owner-guard mutation case still fires.

The three other inline comments in the previous round referenced the concealed-policy detection code, which was removed entirely in the simplification pass -- they're moot.

rrazvd:
## 🏛️ apm-review-panel (local dogfood, working tree)

Ran the multi-persona panel on the branch diff. **Consolidated recommendation: APPROVE-WITH-NITS** — unanimous, no REQUEST-CHANGES.

| Reviewer | Verdict |
|---|---|
| supply-chain-security | APPROVE-WITH-NITS — host-pin holds vs cross-host credential routing; `hash_mismatch`/`incomplete_chain` remain unconditionally fail-closed; fetch surface bounded (namespace-depth × 1) |
| python-architect | APPROVE-WITH-NITS — `_remote.py` extraction is the right seam; single-owner boundary is real and dual-guarded |
| auth-expert | APPROVE-WITH-NITS — credential never routed by attacker-influenceable host/ref; walk pins host+port to the project's own remote; extends chain validates host before fetch |
| devx-ux-expert | APPROVE-WITH-NITS + **1 REQUIRED** (visibility) |
| test-coverage | APPROVE-WITH-NITS — user-visible promise defended by behavioral tests; owner-guard mutant provably killed |

### Addressed in `97767eb9c`

- **REQUIRED (devx-ux): the closest-wins walk was invisible at default verbosity.** A project silently picking up a subgroup policy instead of the top-level one produced no output in `warn` mode. Added a `subgroup_scoped` flag to `PolicyFetchResult`, set in the walk when the winning namespace is below the top-level group, threaded through `outcome_routing`, so the `Policy: <source>` line now shows at info even in non-verbose warn mode. Top-level resolutions keep the prior silence.
- **Coverage (test-coverage):** added the positive host-pin **port** branch (matching non-null port strips + fetches), an `empty`-outcome short-circuit at a deep subgroup (README-only apm-policy must shadow a real ancestor, not fall through), and an end-to-end wiring test running the real `_gitlab_namespace_descending` from a nested origin. Extended the walk tests to assert `subgroup_scoped`.

### Considered, not applied

- **python-architect advisory** to narrow the two `except Exception` clauses in `_remote.py`: `test_https_url_parse_exception_returns_none` pins the broad catch as intentional cross-version defensiveness, so the premise (urlparse only raises ValueError) doesn't hold. Clarified the comment instead.

### Advisory / deferred (documented, non-blocking)
- closest-wins is not tighten-only at discovery level, and GitLab 404 conceals private projects → both are real trust-boundary shifts but honestly documented with the RBAC mitigation. No silent drift from the security model.
- Latency of deep-nesting walks and the CHANGELOG lead-line / `extends: "org"` doc parenthetical are left as doc nits.

Verification: `ruff check` + `format` clean; targeted suites **318 passed**; architecture linter exit 0; owner-guard mutation case still kills its mutant.

sergio-sisternes-epam:

> GitLab org-policy discovery now walks nested subgroups closest-first so a team can publish apm-policy under a subgroup; first non-absent wins, 404 ascends, non-404 fails closed.

cc @rrazvd @danielmeppiel -- a fresh advisory pass is ready for your review.

Specialists converge: this is enterprise GitLab completeness, not a consumer-surface change. Python-architect sees a clean split (`_remote.py` owns parse; GitLab closest-wins is a facade Chain of Responsibility) with no split-authority defect. Supply-chain and auth keep F1 host-pin GitLab-gated and AuthResolver on the leaf; nested walk is GitLab-only and extra REST does not bypass credentials. CLI logging and DevX agree the user-facing promise is quiet-by-default for flat remotes and names the resolved subgroup at default verbosity when the walk hits. Performance is O(d) serial REST with a closest-wins data dependency -- expected, not compounding. No persona returned blocking-severity findings.

Test evidence is load-bearing and mostly green: walk scenario tests (44 passed) hold the closest-wins / 404-skip / fail-closed unit traps. The remaining gap is glue, not the walk itself: no test asserts `outcome_routing` forwards `subgroup_scoped` to the install logger (`tests/unit/policy/test_pr_832_findings.py` already proves default-verbosity install names the subgroup, so the user promise holds on a neighboring path). Doc-writer recommended folding the two contract sentences the CHANGELOG already implies -- closest-wins is not tighten-only, and non-404 errors fail closed -- onto the mental-model page. Those are communication, not behavior, and they are the highest-signal docs fold.

Strategically this closes #2753 without a README/quickstart beat. Concealment-probe removal was the right scope cut. Cost to existing GitHub/ADO users is zero (host-pin unchanged). Defensibility is the GitLab-shaped org tree incumbents usually flatten; keep `extends` opt-in tighten-only so closest-wins cannot be misread as a policy merge.

**Aligned with:** Secure by default: F1 host-pin stays strict on GitHub/ADO leaves; GitLab extra REST uses AuthResolver on the leaf host; non-404 walk errors fail closed. Governed by policy: closest-wins is first non-absent document, not a tighten-only merge; nested walk is GitLab-only; extends remains opt-in at the top-level org. Multi-harness / multi-host: GitLab subgroup namespaces walk closest-first; GitHub and ADO keep flat org discovery and the same host-pin. OSS / community-driven: external author iterated through Copilot review (including dropping concealment probe as scope creep); follow-ups should stay fold-sized so the contributor is not taxed. Pragmatic as npm: no new verbs or flags; default verbosity names the winning subgroup and keeps flat remotes silent.

**Growth signal.** Enterprise completeness, not a top-of-funnel launch beat. Reuse the CHANGELOG so-what in release notes: a team can publish apm-policy under a subgroup; closest wins; extends stays opt-in tighten-only. Do not lead the next release post with this over consumer-facing Unreleased items. 404-skip and closest-wins-is-not-tighten-only copy belongs in the enterprise FAQ, not social.

### Panel summary

|---|---|---|---|---|
| Python Architect | 0 | 0 | 1 | Remote parsing is a single owner in `_remote.py`; GitLab closest-wins is a facade Chain of Responsibility. No blocking split-authority defect. |
| CLI Logging Expert | 0 | 0 | 0 | Subgroup-scoped GitLab policy hits promote to info at default verbosity; walk probes stay at debug; top-level warn/off remains silent. Ship. |
| DevX UX Expert | 0 | 0 | 0 | No CLI verbs, flags, or help-text change; closest-wins GitLab discovery names the resolved subgroup at default verbosity and keeps flat remotes quiet. |
| Supply Chain Security Expert | 0 | 0 | 0 | F1 host-pin stays GitLab-gated with passing tests; closest-wins weakening and 404 skip are documented same-host edges. Extra REST uses AuthResolver on the leaf host. |
| OSS Growth Hacker | 0 | 0 | 0 | Enterprise GitLab completeness with a user-facing CHANGELOG hook and a concrete probe-order example; not a top-of-funnel launch beat. |
| Auth Expert | 0 | 0 | 0 | GitLab nested orgs stay on AuthResolver; F1 host-pin remains strict on GitHub/ADO leaves. No credential bypass. |
| Doc Writer | 0 | 3 | 1 | GitLab walk docs match closest-wins, 404-skip, and extends:org=top-level; fold the not-tighten-only and fail-closed contracts onto the mental-model page. |
| Test Coverage Expert | 0 | 1 | 0 | GitLab walk promises have unit traps; outcome_routing does not assert subgroup_scoped forwarding to the logger. |
| Performance Expert | 0 | 0 | 0 | GitLab walk is O(d) serial REST with closest-wins data dependency; origin is one subprocess. command_logger is O(1). No compounding fault. |

### Top 3 follow-ups

1. **[Test Coverage Expert]** Assert that outcome_routing forwards subgroup_scoped to the install logger. -- Governed-by-policy glue is missing at unit while 44 walk tests and test_pr_832_findings.py already prove the user-visible subgroup name. A one-line logger-forwarding trap survives the next routing refactor.
2. **[Doc Writer]** On the mental-model page, state closest-wins is not tighten-only and that non-404 walk errors fail closed. -- Those two sentences are the contract this PR actually ships; leaving them only in CHANGELOG/code lets operators misread closest-wins as a merge and 404-skip as fail-open.
3. **[Doc Writer]** Add not-tighten-only and extends:org=top-level to the apm-usage governance skill. -- Skill authors will copy the GitLab walk without the extend/override rule unless the skill names it.

### Architecture

```mermaid
classDiagram
    direction LR
    class RemoteOrigin {
      <<CanonicalOwner>>
      +_git_remote_origin_url(project_root) str
      +_remote_url_parts(url) tuple
      +_parse_remote_url(url) tuple
      +_extract_org_host_port_from_git_remote(project_root) tuple
    }
    class PolicyDiscovery {
      <<Facade>>
      +_auto_discover(project_root) PolicyFetchResult
      +_gitlab_namespace_descending(remote_url) list
      +_gitlab_walk_candidate() PolicyFetchResult
      +_extract_extends_host(ref, leaf_host) str
    }
    class GitLabAdapter {
      <<Adapter>>
      +_fetch_from_gitlab_repo() PolicyFetchResult
      +_fetch_gitlab_chain_parent() PolicyFetchResult
    }
    class PolicyFetchResult {
      <<ValueObject>>
      +outcome str
      +subgroup_scoped bool
      +source str
    }
    class CommandLogger {
      <<Base>>
    }
    class InstallLogger {
      <<Subclass>>
      +policy_resolved(source, subgroup_scoped)
    }
    class AuthResolver {
      <<Strategy>>
      +try_with_fallback(host, fn)
    }
    class OutcomeRouting {
      <<CollectThenRender>>
      +route_discovery_outcome()
    }
    CommandLogger <|-- InstallLogger
    PolicyDiscovery *-- RemoteOrigin : one origin read
    PolicyDiscovery *-- GitLabAdapter : walk delegates fetch
    PolicyDiscovery ..> PolicyFetchResult : returns
    GitLabAdapter ..> PolicyFetchResult : returns
    GitLabAdapter ..> AuthResolver : REST headers
    OutcomeRouting ..> PolicyFetchResult : reads subgroup_scoped
    OutcomeRouting ..> InstallLogger : policy_resolved
    note for PolicyDiscovery "Chain of Responsibility: closest namespace -> parent -> top-level"
    note for RemoteOrigin "Single owner of git remote get-url origin"
    class RemoteOrigin:::touched
    class PolicyDiscovery:::touched
    class GitLabAdapter:::touched
    class PolicyFetchResult:::touched
    class InstallLogger:::touched
    class OutcomeRouting:::touched
```

```mermaid
flowchart TD
    start["apm install / apm audit --ci"] --> discover["discover_policy_with_chain src/apm_cli/policy/discovery.py"]
    discover --> auto["_auto_discover discovery.py"]
    auto --> origin["[EXEC] _git_remote_origin_url src/apm_cli/policy/_remote.py"]
    origin --> ident["_extract_org_host_port_from_git_remote _remote.py"]
    ident --> host{is_gitlab_hostname host}
    host -->|no| other["_fetch_from_repo / _fetch_ado_org_policy discovery.py"]
    host -->|yes| ns["_gitlab_namespace_descending discovery.py"]
    ns --> walk["_gitlab_walk_candidate discovery.py"]
    walk --> fetch["[NET] _gitlab._fetch_from_gitlab_repo src/apm_cli/policy/_gitlab.py"]
    fetch --> outcome{result.outcome}
    outcome -->|absent 404| more{more namespaces}
    more -->|yes| walk
    more -->|no| absent["return PolicyFetchResult outcome=absent"]
    outcome -->|found or empty| flag["subgroup_scoped = slash in namespace"]
    outcome -->|error malformed 403| closed["fail-closed return non-absent"]
    flag --> chain["_resolve_and_persist_chain _validate_extends_host discovery.py"]
    chain --> parent["[NET] _gitlab._fetch_gitlab_chain_parent _gitlab.py nested extends GitLab-only"]
    parent --> route["route_discovery_outcome src/apm_cli/policy/outcome_routing.py"]
    closed --> route
    absent --> route
    other --> route
    route --> log["InstallLogger.policy_resolved src/apm_cli/core/command_logger.py"]
```

```mermaid
sequenceDiagram
    participant Auto as discovery._auto_discover
    participant Remote as policy._remote
    participant Walk as discovery._gitlab_walk_candidate
    participant GL as policy._gitlab
    participant Log as InstallLogger.policy_resolved
    CLI->>Auto: discover_policy_with_chain
    Auto->>Remote: [EXEC] _git_remote_origin_url
    Remote-->>Auto: origin URL
    Auto->>Remote: _extract_org_host_port_from_git_remote remote_url=
    Auto->>Walk: namespaces closest-first
    Walk->>GL: [NET] _fetch_from_gitlab_repo org=acme/dept/team
    GL-->>Walk: outcome=absent 404
    Walk->>GL: [NET] _fetch_from_gitlab_repo org=acme/dept
    GL-->>Walk: outcome=found
    Walk-->>Auto: subgroup_scoped=True
    Auto->>Log: route_discovery_outcome
```

### Recommendation

Ship the GitLab closest-wins walk; architecture, auth, supply-chain, logging, and DevX agree the behavior is sound and host-pin is intact. Highest-signal fold is the outcome_routing logger-forwarding trap, then the two mental-model sentences (not-tighten-only + fail-closed). Nits (stale `_remote.py` docstring, host:port in the extends table) can wait.

---

<details>

#### Python Architect

- **[nit]** `_remote.py` module docstring still says discovery.py owns `_git_remote_origin_url`. at `src/apm_cli/policy/_remote.py:6`
  The owner registry, linter rule `install-deployment-policy-remote-origin-owner`, and the functions themselves all live in `policy/_remote.py`, but the new module docstring still says `discovery.py` reads origin via its own `_git_remote_origin_url`. That sentence is the next maintainer's map and will reintroduce the double-read this PR just closed.
  *Suggested:* Rewrite the module docstring so `discovery.py` re-exports the helpers and `_auto_discover` reuses one origin URL via `_git_remote_origin_url` in `_remote.py`.

#### CLI Logging Expert

No findings.

#### DevX UX Expert

No findings.

No findings.

#### OSS Growth Hacker

No findings.

#### Auth Expert

No findings.

#### Doc Writer

- **[recommended]** Mental-model page never states closest-wins is not tighten-only. at `docs/src/content/docs/enterprise/apm-policy.md:90`
  `discovery.py` `_gitlab_walk_candidate` returns the first non-absent policy with no merge; tighten-only lives only in `extends:` chains. `apm-policy.md` teaches closest-wins then immediately "Children can only tighten, never relax." Readers of the SoT mental-model page can conclude a closer subgroup policy cannot weaken an ancestor. The footgun is documented only in `governance-guide.md` Known gaps, so the primary page drifts from the contract.
  *Suggested:* After the closest-wins walk, one sentence plus xref: closest-wins replaces the ancestor outright and may be weaker; tighten-only applies only inside `extends:`. Point at `governance-guide.md` GitLab closest-wins is not tighten-only.
- **[recommended]** GitLab walk fail-closed on non-404 errors is undocumented. at `docs/src/content/docs/enterprise/apm-policy.md:96`
  `_gitlab_walk_candidate` continues only on `outcome==absent` (HTTP 404). 401/403, timeout, connection error, and malformed return immediately (fail-closed), matching `_auto_discover`. The new GitLab copy says a miss at any level is a clean no-policy fallthrough and the 404 caution describes skip-and-continue only. Operators can assume a timeout or auth error also walks to the ancestor.
  *Suggested:* Next to the 404 caution: only 404 continues the walk; any other fetch error fails closed at that level and does not probe shallower groups.
- **[recommended]** apm-usage governance skill omits the two other GitLab contracts it should carry. at `packages/apm-guide/.apm/skills/apm-usage/governance.md:426`
  The skill correctly states closest-wins and 404==absent/grant-read, but not that closest-wins is not tighten-only, and not that `extends: org` is the top-level group rather than the nearest ancestor. Agent readers of this condensed SoT will miss the same footguns the website now splits across `apm-policy.md` and `governance-guide.md`.
  *Suggested:* One clause: closest policy may be weaker than an ancestor (RBAC the create permission); `extends: org` is `<top-level-group>/apm-policy`, not the parent subgroup. Leave dotted-name/host:port on the website page.
- **[nit]** policy-reference extends table omits host:port qualification. at `docs/src/content/docs/enterprise/policy-reference.md:105`
  `_fetch_gitlab_chain_parent` requires an explicit host-like first segment to match leaf hostname AND port. The table says only host-qualify a dotted first segment. Canonical host:port note is already on `apm-policy.md`; the reference row is incomplete for self-managed origin-with-port.
  *Suggested:* Extend the `namespace/.../repo` cell: host-qualify dotted first segments, including exact `host:port` when origin has a port; see Policy Files.

#### Test Coverage Expert

- **[recommended]** No test asserts outcome_routing forwards subgroup_scoped to the install logger. at `src/apm_cli/policy/outcome_routing.py:219`
  Scenario Evidence rows for closest-wins / absent-ascends / all-absent / fail-closed / flat / nested-extends / parse_remote_url map to real tests in `tests/unit/policy/test_discovery.py`; grepped those names and ran them (44 passed in 1.89s). The follow-up user promise -- a non-verbose warn install names which GitLab subgroup policy applied -- is split across three modules: discovery sets `PolicyFetchResult.subgroup_scoped`, `command_logger.policy_resolved` emits on that flag, and outcome_routing is the only glue. Grepped tests/ for `subgroup_scoped`: hits only in `test_discovery.py` (flag on the fetch result) and `test_install_logger_policy.py` (logger called with the kwarg directly). `route_discovery_outcome` callers never pass `subgroup_scoped=True` and only `assert_called_once()` on `policy_resolved`. Dropping the two `subgroup_scoped=` kwargs in `outcome_routing.py` would leave every existing test green while `apm install` / `apm audit` silently omit the subgroup source.
  *Suggested:* Extend `tests/unit/policy/test_pr_832_findings.py::TestOutcomeRoutingTable` with `test_found_forwards_subgroup_scoped_to_logger` (and a cached_stale sibling): build `PolicyFetchResult(outcome='found', policy=ApmPolicy(enforcement='warn'), source='org:gitlab.com/acme/dept-a/apm-policy', subgroup_scoped=True)`, call `route_discovery_outcome` with a MagicMock logger, and assert `logger.policy_resolved(..., subgroup_scoped=True)`.
  *Proof (missing at):* `tests/unit/policy/test_pr_832_findings.py::TestOutcomeRoutingTable.test_found_forwards_subgroup_scoped_to_logger` -- proves: When closest-wins resolves a GitLab subgroup policy, a default-verbosity warn install still names the subgroup that applied. [governed-by-policy,devx]
  `logger.policy_resolved.assert_called_once_with(source='org:gitlab.com/acme/dept-a/apm-policy', cached=False, enforcement='warn', age_seconds=None, subgroup_scoped=True)`

#### Performance Expert

No findings.

</details>

sergio-sisternes-epam:
## PR triage recommendation

needs-issue

**Recommendation only, not merge or scope approval.** A responsible
human maintainer must approve. Labels, automated advice, and
silence are not approval.

## Linked issue

#2753 -- open, not labelled `status/accepted`. Automated triage once
proposed acceptance; maintainer danielmeppiel withdrew `status/accepted`
on 2026-09-12 as part of a roadmap reset. Current issue labels:
`theme/governance`, `area/audit-policy`, `area/enterprise`,
`type/feature`, `triage/recommended`.

## Proposed classification

`theme/governance`, `area/enterprise`, `area/audit-policy`, `type/feature`

Existing human choices on #2753 stay unchanged.

## Suggested next action

Wait for a responsible human maintainer to re-record bounded scope
and acceptance on #2753 (or a successor issue) per CONTRIBUTING.md.
Do not treat this comment, Copilot reviews, or the withdrawn
acceptance as authorization to merge.

CODEOWNERS review requests already present: danielmeppiel,
sergio-sisternes-epam. This worker does not add or remove review
requests.

## Suggested PR comment

Thank you for this pull request and for the careful follow-up on
review comments. APM starts with an issue, not an implementation
(https://github.com/microsoft/apm/blob/main/CONTRIBUTING.md).

This PR closes #2753, but that issue is no longer labelled
`status/accepted`: a maintainer withdrew acceptance on 2026-09-12.
Please keep the user problem on #2753 so a maintainer can review
and accept the scope again before implementation is invited.
This PR is labelled `status/deferred` until that happens.

This is advisory classification only. It is not merge approval
and does not assign or request reviewers.

